PR #3, opened 2015-04-02
Stories/dt/keyword density

terw-dan:
created keyword density function and subheading function. Updated function queuer and added comments to function.


js/analyzer.js
```diff
@@ -1,65 +1,232 @@
...
+};
 
-/*text preprocessor*/
+/**
+ * Analyzer initialisation. Loads defaults and overloads custom settings.
+ */
+Analyzer.prototype.init = function() {
 
+    //init preprocessor if not exists
+    if (typeof yst_pp != 'object' || yst_pp.inputText != this.config.textString) {
```
CarolineGeven: Please do not abbreviate `pp`. I assume this means preprocessor?

js/analyzer.js
```diff
@@ -1,65 +1,232 @@
...
+    return yst_pp.stripSpaces(textString);
+};
+
+/**
+ * PreProcessor object definition. Creates _store object and calls init.
+ */
 PreProcessor = function (text){
     //create _store object to store data
     this._store = {};
     this._store.origText = text;
```
CarolineGeven: Please don't abbreviate `origText`

CarolineGeven: Please don't abbreviate `origText`

js/analyzer.js
```diff
@@ -1,65 +1,232 @@
...
+    //remove first function from queue and execute it.
+    if(this.queue.length > 0){
+        this._output.push(this[this.queue.shift()]());
+        this.runQueue();
+    }
+};
+/**
+ * clears current queue of functions, effectively stopping execution of the analyzer.
+ */
+Analyzer.prototype.abortQueue = function(){
```
CarolineGeven: In which cases is this function used?

js/analyzer.js
```diff
@@ -1,65 +1,232 @@
...
+ * @returns resultObject
+ */
+Analyzer.prototype.keywordDensity = function(){
+    if(yst_pp._store.wordcount >= 100) {
+        var keywordMatches = yst_pp._store.cleanText.match(new RegExp(this.config.keyword, 'g'));
+        if ( keywordMatches != null ) {
+            var keywordCount = keywordMatches.length;
+            var keywordDensity = (keywordCount / yst_pp._store.wordcount - (keywordCount - 1 * keywordCount)) * 100;
+        } else {
+            var keywordDensity = 0;
```
CarolineGeven: For readability I would advise to move keywordDensity = 0 above the if statement. This way you don't need the else statement and is it more clear the switch will not be called if keywordDensity is 0

js/analyzer.js
```diff
@@ -1,65 +1,232 @@
...
+    this.queue = [];
+};
+
+
+/**
+ * checks the keyword density of given keyword against the cleantext stored in _store.
+ * @returns resultObject
+ */
+Analyzer.prototype.keywordDensity = function(){
+    if(yst_pp._store.wordcount >= 100) {
```
CarolineGeven: In the original PHP code the statement checks if > 100. Not >= 100

js/analyzer.js
```diff
@@ -1,65 +1,232 @@
...
+            if (formattedHeaders.match(new RegExp(this.config.keyword, 'g')) || headers[i].match(new RegExp(this.config.keyword, 'g'))) {
+                foundInHeader++;
+            }
+        }
+        if (foundInHeader > 0) {
+            var result = {keywordFound: foundInHeader, numberofHeaders: headers.length};
+            var rating = 9;
+        }else{
+            var result = {keywordFound: 0, numberofHeaders: headers.length}
+            var rating = 7;
```
CarolineGeven: According to the PHP code the rating should be 3 (https://github.com/Yoast/wordpress-seo/blob/trunk/admin/class-metabox.php#L1952)

js/analyzer.js
```diff
@@ -1,65 +1,232 @@
...
+Analyzer.prototype.keywordDensity = function(){
+    if(yst_pp._store.wordcount >= 100) {
+        var keywordMatches = yst_pp._store.cleanText.match(new RegExp(this.config.keyword, 'g'));
+        if ( keywordMatches != null ) {
+            var keywordCount = keywordMatches.length;
+            var keywordDensity = (keywordCount / yst_pp._store.wordcount - (keywordCount - 1 * keywordCount)) * 100;
+        } else {
+            var keywordDensity = 0;
+        }
+        switch (true) {
```
CarolineGeven: Please move the rating to a different function, so the functions have a single responsibility. In the case of `keywordDensity` I would expect a function that gets the keyword density. The keyword density rating would be a different function.

js/analyzer.js
```diff
@@ -1,65 +1,232 @@
...
+        result = {name: 'subHeadings', result: null, rating: 7};
+    }else {
+        var foundInHeader = 0;
+        for (var i = 0; i < headers.length; i++) {
+            var formattedHeaders = this.stripKeywords(headers[i]);
+            if (formattedHeaders.match(new RegExp(this.config.keyword, 'g')) || headers[i].match(new RegExp(this.config.keyword, 'g'))) {
+                foundInHeader++;
+            }
+        }
+        if (foundInHeader > 0) {
```
CarolineGeven: Please move the rating to a different function, so the functions have a single responsibility. In the case of subhheaderChecker I would expect a function that checks the subheader. The subheader rating would be a different function.

js/analyzer.js
```diff
@@ -1,65 +1,232 @@
...
+        result = {name: 'subHeadings', result: result, rating: rating};
+    }
+    return result;
+};
+
+/**helper functions*/
+
+/**
+ * removes certain words from string
+ * @params textString and
```
CarolineGeven: I'm missing the `@returns` in the docs.

js/analyzer.js
```diff
@@ -1,65 +1,232 @@
...
+        wordString += '('+this.config.wordsToRemove[i]+')\\b';
+    }
+    var wordsRegex = new RegExp(wordString, 'g');
+    textString =  textString.replace(wordsRegex, '');
+    //remove double space
+    return yst_pp.stripSpaces(textString);
+};
+
+/**
+ * PreProcessor object definition. Creates _store object and calls init.
```
CarolineGeven: 'text' variable is missing in the docs

js/analyzer.js
andizer: Can we do strict comparisons (probably CodeClimate will yell about this in the future) ?

js/analyzer.js
```diff
@@ -1,65 +1,232 @@
...
+ * Analyzer initialisation. Loads defaults and overloads custom settings.
+ */
+Analyzer.prototype.init = function() {
 
+    //init preprocessor if not exists
+    if (typeof yst_pp != 'object' || yst_pp.inputText != this.config.textString) {
+        yst_pp = new PreProcessor(this.config.textString);
+    }
+    //if no available function queues, make new queue
+    if (typeof this.queue != 'Array') {
```
andizer: I would make a function of line 20-29 to set the queue, because it's a single responsibility

andizer:
CR done :+1:

js/analyzer.js
terw-dan: Isn't used yet. For future testing and to be able to interrupt the process.

PR #4, opened 2015-04-03
Stories/dt/stop words

terw-dan:


js/analyzer.js
```diff
@@ -24,11 +24,14 @@ Analyzer.prototype.init = function() {
...
     }
     //if no available keywords, load default array
-    if(typeof this.config.wordsToRemove === 'undefined'){
-        this.config.wordsToRemove =  [' a', ' in', ' an', ' on', ' for', ' the', ' and'];
+    if(typeof this.config.wordsToRemove == 'undefined'){
+        this.config.wordsToRemove =  analyzerConfig.wordsToRemove;
+    }
+    if(typeof this.config.stopWords == 'undefined'){
+        this.config.stopWords = analyzerConfig.stopWords;
     }
```
omarreiss: Nice improvement, but initialization is now getting cluttered with defaults. Could you please put this in a `setDefaults` function?

js/analyzer.js
```diff
@@ -113,25 +117,53 @@ Analyzer.prototype.subheaderChecker = function() {
...
-        if(wordString.length > 0){ wordString += '|'; }
-        wordString += '('+this.config.wordsToRemove[i]+')\\b';
+ * removes strings from array and replaces them with keyword.
+ * @param textString
+ * @param stringsToRemove []
+ * @param replacement (default == space)
+ * @returns {textString}
+ */
+Analyzer.prototype.stringReplacer = function(textString, stringsToRemove, replacement){
+    if(typeof replacement == 'undefined'){replacement = ' '};
```
omarreiss: isn't it possible in JS to set a default in the arguments list like in the following?

```
Analyzer.prototype.stringReplacer = function(textString, stringsToRemove, replacement = ' '){
```

js/analyzer.js
```diff
@@ -96,7 +100,7 @@ Analyzer.prototype.subheaderChecker = function() {
     }else {
         var foundInHeader = 0;
         for (var i = 0; i < headers.length; i++) {
-            var formattedHeaders = this.stripKeywords(headers[i]);
+            var formattedHeaders = this.stringReplacer(headers[i], this.config.wordsToRemove);
```
omarreiss: Please extract `this.stringReplacer(headers[i], this.config.wordsToRemove)` to something more verbose like `this.removeStopWords(headers[i]);`

js/analyzer.js
```diff
@@ -113,25 +117,53 @@ Analyzer.prototype.subheaderChecker = function() {
     return result;
 };
 
+/**
+ * check if the keyword contains stopwords
+ */
+Analyzer.prototype.stopwordChecker = function(){
+    matches = this.stringCounter(this.config.keyword, this.config.stopWords);
```
omarreiss: Please extract `this.stringCounter(this.config.keyword, this.config.stopWords)` to something more verbose like `this.matchStopWords(this.config.keyword)`

js/analyzer.js
```diff
@@ -113,25 +117,53 @@ Analyzer.prototype.subheaderChecker = function() {
...
+    return yst_pp.stripSpaces(textString);
+}
+
+/**
+ * matches string with given array of strings to match.
+ * @param textString
+ * @param stringsToMatch
+ * @returns {matches}
+ */
+Analyzer.prototype.stringCounter = function(textString, stringsToMatch){
```
omarreiss: This method doesn't really count anything. It matches an array of strings to a string. Please rename it to `matchStrings` or `matchStringsArray`.

omarreiss:
A general thing that I've noticed. You're generally naming methods as if they were objects, i.e. `stringReplacer`. 

Now, I know that in JS _everything_ is an object, but from an OOP perspective it makes more sense to view the actual prototypes as instantiable objects and their methods as the messages they accept and communicate with. 

That way I am not having my `Analyzer` have a `stringReplacer` that has an implicit understanding of how to replace strings. Naming it like that also raises the expectation that the `stringReplacer` is a prototype in itself, on which the `Analyzer` is depending. That's not the case. 

Replacing strings is part of the interface of the `Analyzer`. I could have it `replaceStrings`. Semantically speaking, it then becomes something my Analyzer knows about. 

Distinguishing between objects and messages makes your code more intuitive to read and more consistent with how other code is written at Yoast. Please apply that distinction in your code as well.

js/analyzer.js
```diff
@@ -113,25 +117,53 @@ Analyzer.prototype.subheaderChecker = function() {
...
+    return textString.match(this.regexStringBuilder(stringsToMatch));
+}
+
+/**
+ * builds regex from array with strings
+ * @param stringArray
+ * @returns {RegExp}
+ */
+Analyzer.prototype.regexStringBuilder = function(stringArray){
```
omarreiss: This might be renamed to `stringArrayRegex`

omarreiss:
CR 1 done

js/analyzer.js
terw-dan: This is changed in https://github.com/Yoast/js-text-analysis/tree/stories/DT/fleschReading. The init function is split up in multiple functions to avoid clutter.

js/analyzer.js
terw-dan: This function is moved to the stringhelper as a generic function

js/analyzer.js
terw-dan: This function is moved to the stringhelper as a generic function; stringMatcher

js/analyzer.js
terw-dan: Unfortunately not, so we need to do a check if the argument is set or not.

js/analyzer.js
terw-dan: Renamed it to matchString

js/analyzer.js
terw-dan: this is renamed to 'stringToRegex', since it doesn't output an array

terw-dan:
Renaming of functions is included in this PR: https://github.com/Yoast/js-text-analysis/pull/12

PR #12, opened 2015-04-22
Stories/dt/linkcount #8

### LINKED ISSUE #8: Link count analyzer

terw-dan:
Added linkcounter, including checks for outbound links and nofollow attribute, based on the input textstring in stead of DOMelements. 

Fixes #8


.gruntjshintrc
```diff
@@ -0,0 +1,18 @@
+{
+    "boss": true,
```
andizer: I should use single quote's instead of double ones.

Gruntfile.js
```diff
@@ -0,0 +1,38 @@
+/*jslint es5: false */
+module.exports = function(grunt) {
+    'use strict';
+
+    //require('time-grunt')(grunt);
```
andizer: I would remove this line, because it seems it's not needed anymore?

js/analyzer.js
```diff
@@ -12,28 +12,63 @@ Analyzer = function (args){
...
+/**
+ * initializes required objects.
+ */
+Analyzer.prototype.initRequiredObjects = function(){
     //init preprocessor if not exists
-    if (typeof yst_preProcessor !== 'object' || yst_preProcessor.inputText !== this.config.textString) {
+    if (typeof yst_preProcessor !== "object" || yst_preProcessor.inputText !== this.config.textString) {
         yst_preProcessor = new PreProcessor(this.config.textString);
     }
+    //init helper
```
andizer: Please add a new line above this and add a spaces after `//`. This makes it prettier to read.

js/analyzer.js
```diff
@@ -12,28 +12,63 @@ Analyzer = function (args){
...
+    if(typeof yst_stringHelper !== "object"){
```
andizer: Adding a space before `{` ?

js/analyzer.js
```diff
@@ -53,94 +89,251 @@ Analyzer.prototype.abortQueue = function(){
...
-        }
-        result = {name: 'keywordDensity', result: {keywordDensity: keywordDensity.toFixed(1) }, rating: rating};
-    }else{
-        result = {name: 'keywordDensity', result: null, rating: null};
+Analyzer.prototype.keywordDensity = function() {
+    if (yst_preProcessor.__store.wordcount > 100) {
+        var keywordDensity = this.keywordDensityCheck();
+        var rating = this.keywordDensityRating(keywordDensity);
+        return {name: "keywordDensity", result: {keywordDensity: keywordDensity.toFixed(1)}, rating: rating};
+    } else {
```
andizer: In PHP I would remove the else and return the value within this else directly, because the if already break out the method. Is the same possible in JavaScript?

js/analyzer.js
```diff
@@ -53,94 +89,251 @@ Analyzer.prototype.abortQueue = function(){
...
+        var keywordCount = keywordMatches.length;
+        keywordDensity = (keywordCount / yst_preProcessor.__store.wordcount - (keywordCount - 1 * keywordCount)) * 100;
     }
-    return result;
+    return keywordDensity;
 };
+
+/**
+ * rates the keyword density against given number
+ * @param keywordDensity
```
andizer: What is type type of keywordDensity?

Something like `@param integer keywordDensity` makes it easier to understand.

js/analyzer.js
```diff
@@ -53,94 +89,251 @@ Analyzer.prototype.abortQueue = function(){
...
+};
+
+
+/**
+ * calculate Flesch Reading score
+ * @returns {result object}
+ */
+Analyzer.prototype.fleschReading = function(){
+    var score =  (206.835 - (1.015 * (yst_preProcessor.__store.wordcount / yst_preProcessor.__store.sentencecount)) - (84.6 * (yst_preProcessor.__store.syllablecount / yst_preProcessor.__store.wordcount))).toFixed(1);
+    if(score < 0){score = 0;}else if (score > 100){score = 100;}
```
andizer: Please use some newlines to make this line easier to read.

js/analyzer.js
```diff
@@ -53,94 +89,251 @@ Analyzer.prototype.abortQueue = function(){
...
+    return {name: "subHeadings", result: result, rating: rating};
+};
+
+/**
+ * check if the keyword contains stopwords
+ */
+Analyzer.prototype.stopwords = function(){
+    var matches = yst_stringHelper.matchString(this.config.keyword, this.config.stopWords);
+    var stopwordCount = matches !== null ? matches.length : 0;
+    return {name: "stopWords", result: {count: stopwordCount, matches: matches}, rating:5 };
```
andizer: Please add a space between `{` and `name`;

js/analyzer.js
```diff
@@ -53,94 +89,251 @@ Analyzer.prototype.abortQueue = function(){
...
+ */
+Analyzer.prototype.subHeadingsCheck = function(matches){
+    var foundInHeader;
+    if (matches === null){
+        foundInHeader = -1;
+    }else{
+        foundInHeader = 0;
+        for (var i = 0; i < matches.length; i++) {
+            var formattedHeaders = yst_stringHelper.replaceString(matches[i], this.config.wordsToRemove);
+            if (formattedHeaders.match(new RegExp(this.config.keyword, "g")) || matches[i].match(new RegExp(this.config.keyword, "g"))) {
```
andizer: Please assign `new RegExp(this.config.keyword, "g")` to a var and please that var outside the for-lus.

.gruntjshintrc
terw-dan: Since in JavaScript there is no real difference between singles and doubles I decided to do everything with doubles, so if there is singlequote in a textstring it doesn't have to be escaped.

Gruntfile.js
terw-dan: I removed the comment tags. We actually do want to time grunt, so this needs to be included.

js/analyzer.js
terw-dan: removed this comment completely, since it is already in the function description.

js/analyzer.js
terw-dan: This is possible so I removed the else statement and return directly.

js/analyzer.js
terw-dan: Added types to the @params

js/analyzer.js
terw-dan: added new lines to if-statement

js/analyzer.js
terw-dan: Assigned regexp to var in the config object

js/analyzer.js
terw-dan: added spaces at all objects.

PR #7, opened 2015-04-15
Flesch reading

terw-dan:
Added the Flesch Kincaid reading function, including syllablecounters and sentencecounters.

PR #26, opened 2015-05-11
flesch reading (on master)

omarreiss:
see #7


omarreiss:
@terw-dan I notice you are trying to mimic the singleton pattern with lines like these scattered around your code:

```JS
if(typeof yst_stringHelper !== "object"){
   yst_stringHelper = new StringHelper();
}
```

It would be better to have this logic in one place. You can achieve this by creating getters for these objects that take care of object caching:

```JS
function yst_stringHelper() {
  if(typeof yst_stringHelper !== "object"){
    yst_stringHelper = new StringHelper();
  }
  return yst_stringHelper;
}
```

This way you can always safely call the getter straight away without having to worry about initialization all the time.

js/analyzer.js
```diff
@@ -199,11 +299,47 @@ PreProcessor.prototype.textFormatter = function(){
...
 };
 
 /**
+ * counts the number of syllables in a textstring.
+ * @param textString
+ * @returns syllable count
+ */
+PreProcessor.prototype.syllableCount = function(textString) {
+    var count = 0;
+    var wordsArray = textString.split(/[^aeiouy]/g);
```
omarreiss: I find `wordsArray` a bit confusing. You might want to call it `chunksArray` or `chunkedWordsArray` :stuck_out_tongue_winking_eye:

terw-dan: This function has already changed, so the term wordsArray isn't used anymore.

PR #36, opened 2015-06-17
Stories/dt/paragraphcount #34

### LINKED ISSUE #34: update for paragraph count

terw-dan:
fixes #34 
Now checks on double linebreaks if there are no paragraphs.


andizer:
CR done

Petervw:
The keyword was not placed in the first paragraph. The analysis says it is.

image

Petervw:
The link to Flesch–Kincaid readability tests does not open in a new window, please open the link on a new tab/window.

Petervw:
When I use spaces around the focus keyword, the string is not being trimmed. Is that expected behaviour? I use a focus keyword like: " snippet ", which gives me a different result than "snippet".

Petervw:
A few more things to consider:
 - When I enter the focus keyword "SOLID" and use it in the page title as "Solid", was the focus keyword not found;
 - Adding "> in a small text, the Flesch reading ease is recalculated with a higher score;

Petervw:
When I enter a very long meta description, the output in the snippet preview is not limited. The meta description needs to be limited to 156 chars.

terw-dan:
Fixes for all issues are in #42

PR #38, opened 2015-06-23
CSS fixes :)

jdevalk:


css/style.css
```diff
@@ -69,10 +69,11 @@ html{
     background-color: lightYellow;
 }
```
terw-dan: Maybe use hexcode here in stead of color name to keep the file consistent?

terw-dan:
cr done

jdevalk:
Changed background color to use hex code.

PR #75, opened 2015-08-10
Add JS implementation of add_filter/apply_filters

omarreiss:
Todo:

* [ ] Codestyle
* [ ] Tests

@atimmer @terw-dan could you provide feedback on implementation?

### DIFF
--- a/js/YoastSEO.js
+++ b/js/YoastSEO.js
@@ -6,6 +6,7 @@
 YoastSEO = function( args ) {
     window.YoastSEO_loader = this;
     this.config = args;
+    this.filters = {};
     this.inputs = {};
 	this.loadQueue();
     this.stringHelper = new YoastSEO_StringHelper();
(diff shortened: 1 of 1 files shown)


atimmer:
This is more something I would add in the WordPress context, adding this to js-text-analysis doesn't make much sense given `add_filter` and `apply_filters` is very WordPress'y.

js/YoastSEO.js
```diff
@@ -54,6 +55,57 @@ YoastSEO.prototype.removeFromQueue = function( func ) {
...
+        callChain.sort( function( a, b ) {
+            return a.priority - b.priority;
+        });
+
+        var data = data;
+
+        for (var callableObject in callChain) {
+            var callable = callChain[callableObject].callable;
+
+            data = callable(data, context);
```
atimmer: If `data` is a promise we should call the next one once the promise resolves.

js/YoastSEO.js
```diff
@@ -54,6 +55,57 @@ YoastSEO.prototype.removeFromQueue = function( func ) {
...
+ */
+YoastSEO.prototype.applyFilters = function(filter, data, context) {
+    var callChain = this.filters[filter];
+
+    if ( callChain instanceof Array && callChain.length > 0 ) {
+        callChain.sort( function( a, b ) {
+            return a.priority - b.priority;
+        });
+
+        var data = data;
```
atimmer: This line this superfluous.

js/YoastSEO.js
```diff
@@ -54,6 +55,55 @@ YoastSEO.prototype.removeFromQueue = function( func ) {
...
+        callChain.sort( function( a, b ) {
+            return a.priority - b.priority;
+        });
+
+        for (var callableObject in callChain) {
+            var callable = callChain[callableObject].callable;
+
+            data = callable(data, context);
+        }
+        return data;
```
terw-dan: If multiple callableObjects are in callChain, only the last will return data (since the data variable is overwritten on each execution in this loop).

omarreiss: What I've seen happen, is that the data is updated everytime before the next callable gets to filter it. I think that's what we want.

omarreiss:
closing in favor of #84

PR #79, opened 2015-08-12
Rename Initializer to App, have an initialize function construct it and assign it.

omarreiss:
@atimmer this feels slightly more intuitive to me.

I guess we are initializing a content analysis app here. Let's just name it that, construct it, let it have its init function etc.

No more desolate Globals. Everything now contained by `YoastSEO`.


atimmer:
Looks fine to me. CR done 👍🏻

PR #83, opened 2015-08-13
Changed keyword to match with non latin chars.

### LINKED ISSUE #264: Keyword count at the beginning and end is broken
When having this exact string: `keyword keyword` with `keyword` as the keyword.

To test this you can add the following test to `spec/stringProcessing/wordMatchSpec.js`:
```js
it( "should match at the start and the end of the string correctly", function() {
	expect( wordMatch( "keyword keyword keyword", "keyword" ) ).toBe( 3 );
	expect( wordMatch( "keyword keyword", "keyword" ) ).toBe( 2 );
});
```

terw-dan:
Fixes Yoast/wordpress-seo#264, Yoast/wordpress-seo#681, Yoast/wordpress-seo#1236, Yoast/wordpress-seo#915, Yoast/wordpress-seo#1103 and Yoast/wordpress-seo#613
The keyword density is now checked with a regex so it matches non latin chars. Also the diacritics are replaced on the keyword so they match with the text that has the diacritics replaced.


atimmer:
CR Done 👍🏻

atimmer:
CR2 Done 👍🏻

PR #85, opened 2015-08-18
Change the way the app should be initialized

### LINKED ISSUE #82: Improve initialisation flow
I really want to initialise the analyser like this:

```js
var analyzerApp = new YoastSEO.App({
    contentFunction: contentCallback,
    titleFunction: titleCallback,
    queue: [ "step1", "step2" ]
});
analyzerApp.analyze();
```

We should define what the bare minimum is that the analyser requires and document this in the `README`. There should be no object required to pass in the `App`.

We should document all possible parameters and parameter types in the `README`.

atimmer:
Completely remove the dependency on passing a `source` and calling functions on that object. Following the "Tell don't ask" principle in the `App` object we except several relevant callbacks to be passed and that is all we need. These are properly documented and have a clear use case.

Should be just in conjunction with Yoast/wordpress-seo#2855

Fixes #82


js/app.js
```diff
@@ -4,10 +4,69 @@
 YoastSEO = ( "undefined" === typeof YoastSEO ) ? {} : YoastSEO;
```
terw-dan: Please change to typeof YoastSEO === "undefined" since that notation is used everywhere in the analyze

terw-dan:
CR Done

terw-dan:
CR done

terw-dan:
CR done

PR #90, opened 2015-09-21
Namespace jed

omarreiss:
We want to ship Jed with YoastSEO.js but we don't want to claim it just for ourselves. The solution: namespace it under `YoastSEO`.


grunt/config/concat.js
```diff
@@ -1,9 +1,16 @@
 module.exports = {
 	options: {
-		separator: ";"
+		separator: ";",
+		process: function( src, filepath ) {
+			"use strict";
+			if ( filepath === "node_modules/jed/jed.js" ) {
```
atimmer: Please add a comment why we are doing this.

PR #92, opened 2015-09-23
1.0-RC

terw-dan:
- adds unformattedData to snippetPreview so when an element is clicked in the snippet preview it is filled with the unformatted value so this can be editted
- adds wordboundary to regex for the keyword highlighting
- fixed bug where empty strings where counted as 1 word


omarreiss:
Just one problem. This pull breaks backwards compat

PR #93, opened 2015-09-23
Make sure url keyword check works for the entire url

omarreiss:
Tested in both WordPress and Drupal.

The issue was that the slug is not included in the url and therefore was not analyzed.


omarreiss:
I see I broke the build...

terw-dan:
CR Done

PR #109, opened 2015-10-19
Stories/dt/fixkeywordregex 108

### LINKED ISSUE #108: filter \ in keywordregex
currently the function to create the keyword regex can break if a \ is used in the keyword. 
Replace this (and any other characters that can break regex) to prevent breaking the regex.

terw-dan:
Fixes #108. 
Adds filter to the regexes to filter out unwanted characters (- [ ] / { } ( ) * + ? . \^ $ |) that could break the regex, or return unexpected results since these are keycharacters in the regex.


dist/yoast-seo.js
```diff
@@ -2182,13 +2186,17 @@ YoastSEO.SnippetPreview.prototype.getPeriodMatches = function() {
...
 /**
  * formats the keyword for use in the snippetPreview by adding <strong>-tags
+ * strips unwanted characters that could break the regex or give unwanted results
  * @param textString
  * @returns textString
  */
 YoastSEO.SnippetPreview.prototype.formatKeyword = function( textString ) {
 
+	// removes characters from the keyword that could break the regex, or give unwanted results.
+	var keyword = this.refObj.rawData.keyword.replace( /[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "" );
```
CarolineGeven: Is it possible to save this in a class variable so you don't need to set the regex everytime? In this file alone, this exact regex occurs 3 times. If you need to change the regex, you have to change it in 3 spots as well. It'd be better to have 1 place where this is set and changed.

CarolineGeven:
CR1 done

CarolineGeven:
CR2 done :+1:

diedexx:
Acceptance Done. @CarolineGeven Could you please check and confirm?

dist/yoast-seo.js
```diff
@@ -2481,6 +2492,11 @@ YoastSEO.StringHelper.prototype.stripAllTags = function( textString ) {
 	return textString;
 };
 
+YoastSEO.StringHelper.prototype.cleanRegex = function( textString ) {
```
omarreiss: I don't know which regex this cleans, but to me it seems like this function sanitizes a string. @atimmer please rename accordingly.

dist/yoast-seo.js
```diff
@@ -2197,12 +2206,14 @@ YoastSEO.SnippetPreview.prototype.formatKeyword = function( textString ) {
...
  * <strong>-tags
+ * strips unwanted characters that could break the regex or give unwanted results
  *
  * @param textString
  * @returns {XML|string|void}
  */
 YoastSEO.SnippetPreview.prototype.formatKeywordUrl = function( textString ) {
-	var replacer = this.refObj.rawData.keyword.replace( " ", "[-_]" );
+	var keyword = this.refObj.stringHelper.cleanRegex( this.refObj.rawData.keyword );
+	var replacer = keyword.replace( " ", "[-_]" );
```
omarreiss: What is a replacer? This is a keyword, and it's sanitized so maybe something like `sanitizedKeyword` will do. @atimmer please rename accordingly.

dist/yoast-seo.js
```diff
@@ -2182,13 +2187,17 @@ YoastSEO.SnippetPreview.prototype.getPeriodMatches = function() {
...
  * @returns textString
  */
 YoastSEO.SnippetPreview.prototype.formatKeyword = function( textString ) {
 
+	// removes characters from the keyword that could break the regex, or give unwanted results, includes the -
+	var keyword = this.refObj.rawData.keyword.replace( /[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "" );
+
 	//matches case insensitive and global
-	var replacer = new RegExp( "\\b" + this.refObj.rawData.keyword + "\\b", "ig" );
+	var replacer = new RegExp( "\\b" + keyword + "\\b", "ig" );
```
omarreiss: Isn't this a regex?

omarreiss:
@atimmer please fix comments and merge.

PR #118, opened 2015-10-27
Add an API to register custom analyzer tests

atimmer:
I chose to add a property on the analyser called `analyses` where we can save custom analyses and in the future save our own.

This PR will add the custom tests to the analyzer every time just before the queue is executed.

The refactor regarding switching the responsibilities of the analyser and the scoring will have to wait.


js/pluggable.js
```diff
@@ -294,6 +361,40 @@ YoastSEO.Pluggable.prototype._applyModifications = function( modification, data,
...
+ *
+ * @param {YoastSEO.Analyzer} analyzer
+ * @param {Object}            pluginTest
+ * @param {string}            pluginTest.name
+ * @param {function}          pluginTest.callable
+ * @param {Object}            pluginTest.scoring
+ * @private
+ */
+YoastSEO.Pluggable.prototype._addPluginTest = function( analyzer, pluginTest ) {
+	analyzer.addAnalysis( {
```
terw-dan: This always adds the test, even when a custom queue is given, for example the queue when no keyword is found. 
Perhaps we could check here (or in the addAnalysis function in the analyzer if the function could/should be added).

PR #129, opened 2015-11-04
renamed metadescription to metadescriptionlength

### LINKED ISSUE #123: I am only seeing 3 tests when no keyword is given
screen shot 2015-11-02 at 20 39 32

terw-dan:
renamed metadescription to metadescriptionlength so this is included in the no keyword queue and generates output. 

6 tests are run, not all are always shown. 
- fleshreading only triggers if the content has any words
- url stopwords only shows when there are stopwords in the URL
- keywordcheck, pagetitlelength, wordcount and metadescriptionlength should always be shown.

fixes #123


andizer:
CR done :)

PR #140, opened 2015-11-06
Translate SEO Score screen reader text

### LINKED ISSUE #137: Make "seo score" translatable
Currently "seo score" is hardcoded, this should be translatable.

Also I really want to have a public function to do this because we need to have access to this in the different post scrapers.

atimmer:
Fixes #137 

@terw-dan Can you review?


js/scoreFormatter.js
```diff
@@ -110,3 +120,39 @@ YoastSEO.ScoreFormatter.prototype.scoreRating = function( score ) {
...
+ *
+ * @return {string}
+ */
+YoastSEO.ScoreFormatter.prototype.getSEOScoreText = function( scoreRating, i18n ) {
+	var scoreText = "";
+
+	switch ( scoreRating ) {
+		case "na":
+			scoreText = i18n.dgettext( "js-text-analysis", "No keyword" );
+			break;
```
terw-dan: na is used for when the score is not available. Currently we use this only when no keyword is specified, but this could change in de future.

terw-dan:
CR Done

Issue #104, opened 2015-10-06
Matching keywords in non-latin alphabets doesn't work properly


terw-dan:
Currently matching keywords in non-latin alphabets (ie Hebrew or Cyrillic) doesn't work properly, since wordboundaries aren't recognized in these languages. 

The problem is the fact JavaScript cannot match wordboundaries in these languages, so words that match partially are also found as match (so 'ab' not only matches 'ab', but also 'abc' or 'dab', where it should only match 'ab'). 
see: http://blog.stevenlevithan.com/archives/fixing-javascript-regexp

A possible solution would be to convert all spaces or periods to a specific character and use that as word boundaries. 
Another solution is converting every character to a alphanumeric char (a-z0-9), but this limits to only 36 characters, and some languages have (many) more. 
Also, not in every language it would be desirable to use spaces as wordboundaries, as certain prefixes of words would still make a good match of the keyword.

atimmer:
This has been fixed.

PR #145, opened 2015-11-10
Add wordboundaries to regex to make sure matching is correct. 

### LINKED ISSUE #143: "The focus keyword appears in the url of the page"
screen shot 2015-11-09 at 14 10 27

We should only match the focus keyword in the url as a separate word.

terw-dan:
Fixes #143 and Fixes https://github.com/Yoast/wordpress-seo/issues/3160
Added wordboundaries to keywordRegex to make sure it doesn't match parts of words, and highlight these incorrectly in the snippet. 

disabled the tests for keyword regex with special characters, these don't work since JavaScript doesn't work properly with non-latin tests (see https://github.com/Yoast/YoastSEO.js/issues/104)


tacoverdo:
Fixes #145 and https://github.com/Yoast/wordpress-seo/issues/3160.

PR #147, opened 2015-11-10
when the meta is empty, make the text grey

terw-dan:
When the meta description is generated from text, because it is empty the text should be grey in the snippetPreview. 

If a text is filled out, it should be darker.


atimmer:
CR Done 👍🏻

CarolineGeven:
Acceptance done

PR #150, opened 2015-11-12
Updating scoring array to comply with scoring text for keyword density.

terw-dan:
Fixes #https://github.com/Yoast/wordpress-seo/issues/3077

Changes the range of the keyword density check so 2.5% is included in the good score.


atimmer:
CR Done 👍🏻

PR #152, opened 2015-11-12
Fix off by one error in meta description length

atimmer:
A meta description of exactly 156 characters should be inside the bounds

Introduce a unit test

Fixes https://github.com/Yoast/wordpress-seo/issues/3072


CarolineGeven:
CR and acceptance done

PR #153, opened 2015-11-12
Fix keyword in alt detection with multiple images

atimmer:
Previously when two images were present and only one had the keyword
the analyzer reported that there were no images with alt tags, this
fixes that. Also adds unit test.

Fixes https://github.com/Yoast/wordpress-seo/issues/3123


terw-dan:
CR Done

PR #155, opened 2015-11-13
Fixes matching with hyphens. 

terw-dan:
Fixes #https://github.com/Yoast/wordpress-seo/issues/3202

Make sure the hyphens are correctly processed by the analyzer so these words match.


atimmer:
CR Done.

CarolineGeven:
## Acceptance
This issue is fixed for words with a hyphen, but not with an apostrophe and hyphen. For example:
`ninety-nine` is fixed. However `'s-Heerenberg` is not

PR #159, opened 2015-11-16
Fix keyword density with alt-counts. 

terw-dan:
Fixes #https://github.com/Yoast/wordpress-seo/issues/3194
Fixes #https://github.com/Yoast/wordpress-seo/issues/3214

Checks on the correct string for counting the keywords.


CarolineGeven:
## Acceptance
Issues are not fixed.

spec/keywordCountSpec.js
```diff
@@ -0,0 +1 @@
+
```
diedexx: Empty file

PR #163, opened 2015-11-16
Improved matching for keywords in URLs

terw-dan:
Improved the checking for keywords in links, by removing everything between < > in an anchor to make sure we only have the textnode. 

Fixes https://github.com/Yoast/wordpress-seo/issues/3242


atimmer:
CR Done 👍🏻

PR #168, opened 2015-11-17
Fix matching spaces in keywords

atimmer:
Matching spaces was broken after 65a1b0c, fix this. Also adds unit tests
.


omarreiss:
CR :+1:

PR #174, opened 2015-11-17
Stop rendering scores that aren't displayed

atimmer:
This should be fixed in a better way, for now just exclude scores with
an empty text.

Fixes https://github.com/Yoast/wordpress-seo/issues/3264


omarreiss:
CR :+1:

diedexx:
Acceptance done :+1:

PR #176, opened 2015-11-17
Meta case sensitivity with hyphens

### LINKED ISSUE #172: adding a hyphenated keyword with first letter in caps to meta description doesn't work
When adding a hyphenated keyword with the first letter in caps to the meta description, the analysis claims there's no focus keyword, but the keyword is highlighted.

schermafbeelding_2015-11-17_om_15_49_39

terw-dan:
Fixes #172

Match the meta description with lowercase regex to make sure capitals are matched as well, also when hyphens are included on the page.


CarolineGeven:
CR done

omarreiss:
I've created a branch on wordpress-seo to test this: `test/meta-case-sensitivity`

PR #178, opened 2015-11-18
Keywords with multiple spaces

terw-dan:
Fixes #https://github.com/Yoast/wordpress-seo/issues/3299

Add global flag to the replacement of whitespaces for the matching of the URL. Now it replaces all spaces with dashes so it matches the keyword in the URL when it has multiple spaces.


CarolineGeven:
CR done

CarolineGeven:
## Acceptance
When the focus keyword consists of more than three words, the error 'The focus keyword does not appear in the URL' still shows up. Since we allow up to 10 words as focus phrase, we should be able to detect up to 10 words in the URL.

atimmer:
CR & Acceptance OK.

PR #183, opened 2015-11-19
Check output target before transversing DOM

seripap:
In newer versions of YoastSEO, if I remove the metabox

```php
remove_meta_box( 'wpseo_meta', 'tabs', 'normal' );
```

The main JS file will throw an error as it is checking to input content into something that doesnt exist. This checks the output target before setting or manipulating values inside of the target.


atimmer:
This should be fixed on the WordPress side of things.

I've made an issue on wordpress-seo tracking this: https://github.com/Yoast/wordpress-seo/issues/4238

PR #194, opened 2015-11-24
Match keywords with diacritics in alt-tags. 

### LINKED ISSUE #187: Keyword not recognised in image alt tag
The keyword `maïs` (note the diaeresis on the i), is not recognised in an alt tag.

terw-dan:
Fixes #187
If a diacritic was used in an alt-tag and keyword it didn't match the keyword in the alt-tag, even if it was there. Made sure the diacritics are replaced in the alt-tag to make sure we show a match. 
Added a number of tests for the image alt-tags.


diedexx:
Acceptance Done :+1:

PR #195, opened 2015-11-24
Remove word boundary matching to match non-latin languages

atimmer:
Fixes https://github.com/Yoast/wordpress-seo/issues/3369 and https://github.com/Yoast/wordpress-seo/issues/3298


CarolineGeven:
## Acceptance
When adding an image with the focus keyword in the alt text in a non-latin language, the text analysis claims:
> The images on this page are missing alt tags.

PR #208, opened 2015-12-31
Render the snippet editor with a lodash template

atimmer:
The previous method was using javascript functions, this was not very
readable or maintainable. This approach has as added bonus that we can
easily switch template language later.

I added the browserify dependency to be able to test with the standalone version. This is in place temporary until the whole library is modularised and we move the browserifying to the consumer of the library.


andizer:
Looks good :+1: 

I see you are introducing lodash and browserify. It is not clear for me what these to new things are doing. Can you add documentation somewhere(readme/wiki or something like that) to explain what these things are doing. This makes it more clear.

stramel:
:+1:

Issue #205, opened 2015-12-22
Bower support?


briebug:
Any thoughts on adding this to bower as a library?

briebug:
Additionally, I am unable top use bower to install using:

    git install https://github.com/Yoast/YoastSEO.js.git

 because the version number is wrong in the bower file.   I believe it is supposed to be 1.0.0, not 1.0

PR #212, opened 2016-01-12
Fix bower version

aarmour:
Match version in package.json
[#205]


aarmour:
`version` is [deprecated](https://github.com/bower/spec/blob/master/json.md#version), so perhaps it would be better to remove it.

PR #221, opened 2016-01-19
Fix SVG icons in Safari

atimmer:
Safari got weird with the width: auto. Set the with to the correct value
to force Safari to render it correctly.

Fixes https://github.com/Yoast/wordpress-seo/issues/3720


omarreiss:
CR and acceptance :heavy_check_mark:

Issue #225, opened 2016-01-20
Text analysis snippet editor errors


CarolineGeven:
When there is a title and a meta description that does contain the focus keyword, the content analysis still claims there is no page title, the focus keyword does not appear in the page title and the meta description does not contain the focus keyword.

![schermafbeelding_2016-01-20_om_12_25_35](https://cloud.githubusercontent.com/assets/8614579/12449510/2294ef48-bf7e-11e5-8b47-e3ed02daa631.png)

These errors do not appear when the title and the meta description in the snippet editor are edited

PR #229, opened 2016-01-21
Apply modifications on analyzer data

atimmer:
We would send empty data to the analyzer, fix this.

Fixes https://github.com/Yoast/YoastSEO.js/issues/225
Fixes https://github.com/Yoast/wordpress-seo/issues/3748


jcomack:
Except for the conflicts, it looks good to me. CR done 👍

jcomack:
Acceptance done.

Whilst testing, I noticed that the analyzer result `The page title contains the focus keyword, at the beginning which is considered to improve rankings.` contains a typo. The comma should come after the word 'beginning'.

I also noticed that although the keyword is not the first word in the title, the same message still persists. Only once you edit the keyword does the analyzer check if the keyword is the first word in the title. This is incorrect behaviour.

jcomack:
Acceptance 2 done

Issue #227, opened 2016-01-20
The keyword density is 0.5%, which is great; the focus keyword was found undefined times.


CarolineGeven:
When the keyword density is calculated, the content analysis shows the following message:

> The keyword density is 0.5%, which is great; the focus keyword was found undefined times.

jcomack:
This is related to https://github.com/Yoast/wordpress-seo/issues/3746, which has been tested and accepted.

PR #231, opened 2016-01-21
Store the keyword match count in the store for BC

atimmer:
The `.refObj.__store.keywordCount` option in scoring.js needs that we
have the keyword count int he analyzer __store. The analyzer __store
isn't used for any other purpose so we want to get rid of this but not
now.

Fixes https://github.com/Yoast/YoastSEO.js/issues/227
Fixes https://github.com/Yoast/wordpress-seo/issues/3746


jcomack:
CR done 👍

jcomack:
Acceptance done

Issue #226, opened 2016-01-20
Change meta description edit display in snippet editor


CarolineGeven:
When editing the meta description in the snippet editor, currently there is no realtime feedback to the user about the limitations of the meta description. 
This means a user can keep writing without any feedback about the character limit.
![schermafbeelding 2016-01-20 om 14 38 29](https://cloud.githubusercontent.com/assets/8614579/12451366/430be618-bf89-11e5-900f-e6935389d368.png)

When the user clicks outside the meta description edit box, there appears a red border around the field. However this is not sufficient enough for for example (color)blind users.
![schermafbeelding 2016-01-20 om 14 38 33](https://cloud.githubusercontent.com/assets/8614579/12451377/558ddf30-bf89-11e5-80c8-e2af9c839330.png)

The current solution is a regression compared to 3.0.7. In 3.0.7 the text in the snippet editor became red when the total amount of characters was exceeded.
<img width="528" alt="schermafbeelding 2016-01-20 om 15 15 34" src="https://cloud.githubusercontent.com/assets/8614579/12451277/b986cd72-bf88-11e5-9c2c-8fe75da27984.png">

However, this is still a regression from 2.3.5, where there was a character countdown. 
![schermafbeelding 2016-01-20 om 15 22 10](https://cloud.githubusercontent.com/assets/8614579/12451417/95f275fe-bf89-11e5-99cd-935713aca2eb.png)

As seen in this ticket https://github.com/Yoast/wordpress-seo/issues/3430 and as discussed with @tacoverdo, quite some people in support are requesting for at least the character countdown to come back. If we cannot do this for the upcoming milestone, please make sure we are not creating a regression from 3.0.7

PR #232, opened 2016-01-21
Give feedback by modifying the text color

atimmer:
This was the case in the previous version so we want to maintain this.

Fixes https://github.com/Yoast/YoastSEO.js/issues/226


jcomack:
CR done

jcomack:
Acceptance done

PR #233, opened 2016-01-21
Don't analyze the base url.

atimmer:
The analyzer expects that the `url` key contains only the url path so
only send that part.

Fixes https://github.com/Yoast/wordpress-seo/issues/3768


jcomack:
CR done

jcomack:
Acceptance done

PR #235, opened 2016-01-22
Fixes adding date to empty metadata

### LINKED ISSUE #234: The meta description is under 120 characters when meta description is generated
When the meta description gets auto generated, the content analysis says there are less than 120 characters in the meta description, while there are in reality none (as it is auto generated). Perhaps we should hide this message if the meta description is auto generated.

schermafbeelding_2016-01-21_om_12_38_54

terw-dan:
Fixes #234 
A date should only be added to the metadescription if the metadescription isn't empty.


jcomack:
CR done 👍

PR #242, opened 2016-01-25
Fixes incorrect placement of meta template

terw-dan:
When there is a template defined for the metadescription, it was only added when the text was empty. 
This fix checks if there is a placeholder defined for the meta description (that isn't the default in the snippetPreview) and uses this as metaText before it takes the text from the excerpt, or generates it from the text (like it was in 2.3.5). 
If all options are still empty, it uses the default placeholder text in the snippetPreview.


atimmer:
CR Done 👍🏻

jcomack:
Acceptance done

PR #248, opened 2016-01-27
Adds dash to regex so it matches in url

terw-dan:
Fixes https://github.com/Yoast/wordpress-seo/issues/3816

Adds a dash to the regex for the keyword highlighting in the snippetPreview, so a dash is used as a wordboundary for matching the keyword in the URL.


atimmer:
This really needs unit tests.

jcomack:
CR done

boblinthorst:
Acceptance: works for dashes now :+1:

PR #249, opened 2016-01-27
Strip whitespaces for progressbar in snippet preview

terw-dan:
Removes extra whitespaces in the title and metadescription before using the length for the progressbar. Since the spaces are stripped in the output, the length and progressbar didn't match. 
Also removes these whitespaces in the check for the metadescription and title before adding the --invalid class to these elements in the snippet preview. 
Includes a CSS fix to make sure the title turns red when the --invalid class is applied. 

Fixes https://github.com/Yoast/wordpress-seo/issues/3813


css/analyzer.css
```diff
@@ -171,8 +171,9 @@
     color: #f00 !important;
 }
 
+.snippet-editor__form .snippet-editor__field--invalid,
```
atimmer: We shouldn't have to do this, there is probably an underlying problem we should instead solve. The point of BEM is reducing the specificity of the CSS selectors.

Please find the source of CSS that makes you write this line and solve it in a better way.

css/analyzer.css
```diff
@@ -171,8 +171,9 @@
...
 .snippet-editor__field--invalid {
-    color: #f00;
+    color: #f00
```
atimmer: This requires a semicolon.

js/snippetPreview.js
```diff
@@ -828,13 +828,13 @@ SnippetPreview.prototype.validateFields = function() {
 	var metaDescription = getAnalyzerMetaDesc.call( this );
 	var title = getAnalyzerTitle.call( this );
 
-	if ( metaDescription.length > YoastSEO.analyzerConfig.maxMeta ) {
+	if ( YoastSEO.getStringHelper().stripSpaces( metaDescription ).length > YoastSEO.analyzerConfig.maxMeta ) {
```
atimmer: Please use the module instead of using the deprecated stringHelper.

js/snippetPreview.js
```diff
@@ -847,8 +847,8 @@ SnippetPreview.prototype.validateFields = function() {
 SnippetPreview.prototype.updateProgressBars = function() {
 	var metaDescriptionRating, titleRating, metaDescription, title;
 
-	metaDescription = getAnalyzerMetaDesc.call( this );
-	title = getAnalyzerTitle.call( this );
+	metaDescription = YoastSEO.getStringHelper().stripSpaces( getAnalyzerMetaDesc.call( this ) );
```
atimmer: Please move the `stripSpaces` function to the `getAnalyzerMetaDesc` function, this makes sure it's always applied.

atimmer:
CR Done

jcomack:
CR done 👍

diedexx:
Acceptance Done :+1:

PR #257, opened 2016-02-04
Removes spaces in stripHTMLtags module.

terw-dan:
The stripHTMLtags module replaces htmltags with spaces, but this might give double spaces when multiple tags are removed. 
Adds stripSpaces after replacing the HTMLtags so no double spaces are included in the outputString. 

Includes spec for stripHTMLTags


spec/stringProcessing/stripHTMLTagsSpec.js
```diff
@@ -0,0 +1,16 @@
+var stripHTMLTags = require( "../../js/stringProcessing/stripHTMLTags.js" );
+
+describe( "strips the HTMLtags from a string", function(){
+	it( "returns a string without HTMLtags", function(){
```
diedexx: I wonder what it does with a string with multiple consecutive spaces 
(like: "this &nbsp;&nbsp; &lt;b&gt; is &nbsp;&nbsp; a &lt;/b&gt; teststring")
please add tests for this.

diedexx:
CR 1 done :+1:

terw-dan:
Added a test with multiple consecutive spaces by suggestion of @diedexx

terw-dan:
the stripHTMLtags module is correct, even though it doesn't show up in the latest commit, where only the test was added.

PR #265, opened 2016-02-09
Fix automatic meta description generation

atimmer:
The automatic meta description generation based on the focus keyword was
broken on trunk. This fixes it.


omarreiss:
When I choose a focus keyword that doesn't occur in the text, the metadescription becomes empty.

omarreiss:
The current matching is case sensitive. When I have "Avocado" as my focus kw, it will not match the phrase "Everybody loves avocado"

omarreiss:
CR :-1:

jcomack:
Had to delete the code below as it resulted in a permanent loop.
Posting this as a possible back up example of how things used to be.

```
// Generate a meta description based on where in the text the keyword is found first
	if ( this.refObj.rawData.keyword !== "" && this.refObj.rawData.text !== "" ) {
		var curStart = 0;
		var indexMatches = this.getIndexMatches();
		var periodMatches = this.getPeriodMatches();

		metaText = metaText.substring( 0, analyzerConfig.maxMeta );

		if ( indexMatches.length > 0 ) {
			var j = 0;

			while ( j < periodMatches.length ) {
				if ( periodMatches[ 0 ] < indexMatches[ 0 ] ) {
					curStart = periodMatches.shift();
				} else if ( curStart > 0 ) {
					curStart += 2;
				} else {
					break;
				}
			}
		}
	}
```

atimmer:
Closed in favor of #387

PR #266, opened 2016-02-09
Introduce default values for all snippet fields

atimmer:
Add settable defaults for all the snippet editor fields. These values
will be used for the analyzer, rendering and the progress bars.

The placeholders will no longer be passed to the analyzer or be used for
the progress bar.

This fixes the bug that the progress bar would be 100% while the user didn't set a meta description.


omarreiss:
The title is not empty, still the progress indicator shows no progress: 
screen shot 2016-02-09 at 15 57 54

omarreiss:
Same goes for the meta
screen shot 2016-02-09 at 16 01 06

omarreiss:
When I change the title template to something that resolves to empty, the title field in the snippet preview simply vanishes:
screen shot 2016-02-09 at 16 03 29

We should probably have a default.

omarreiss:
"Please provide an SEO title by editing the snippet below."

PR #281, opened 2016-02-11
Match apostrophes in keyword in the URL

### LINKED ISSUE #273: Slug match is not found with quote in focus keyword
The slug `joost-de-valks-wordpress-plugin` should match the keyword `Joost de Valk's Wordpress Plugin`.

It does match when the quote is removed from the keyword `Joost de Valks Wordpress Plugin`

Test copy:

> Let's say my focus keyword is "Joost de Valk's Wordpress Plugin". That's also my title and that exact phrase also appears within my text in two separate places.
> 
> The text analysis always says, "The focus keyword does not appear in the URL for this page. If you decide to rename the URL be sure to check the old URL 301 redirects to the new one!". The only thing missing in the URL is the apostrophe, which a URL cannot contain anyway. Google knows this and I think the plugin should too.
> 
> The text analysis sometimes also says, "The focus keyword 'Joost de Valk's Wordpress Plugin' does not appear in the page title." I cannot determine why it happens sometimes and not other times as of yet.

Focus keyword:
> Joost de Valk's Wordpress Plugin

Title:
> Joost de Valk's Wordpress Plugin

Slug:
> joost-de-valks-wordpress-plugin

See wordpress-seo issue [#3694](https://github.com/Yoast/wordpress-seo/issues/3694)

atimmer:
When there is an apostrophe in the keyword and the URL doesn't have the
apostrophe the analyzer and snippet preview should still match it.

Fixes #273


andizer:
Can you fix the travis errors.

atimmer:
Done

andizer:
CR 2 done

andizer:
Acceptance done.

PR #282, opened 2016-02-11
Don't strip apostrophes in getAlttagContent

### LINKED ISSUE #280: Focus keyword with quote is not matched in alt text
See screenshot:

screen shot 2016-02-11 at 10 11 03

atimmer:
In getAlttagContent apostrophes would be stripped by the function, this
can be prevented by using a regex capture group to get the alt attribute
content.

Fixes #280


andizer:
CR done :+1:

boblinthorst:
Acceptance:

It's not working 100% properly yet. This is the behaviour i've noticed:

it works if you start in the visual editor and create your image with alt text (featuring the keyword with an apostrophe) there. 
it doesn't work if you start in the text editor. 
if you started in the text editor and switch to the visual editor it starts behaving properly (fixes itself).
if you then switch back from visual to text. it doesn't break. so it keeps working properly. (what i had expected). 

Try to fix it so that it'll work properly for the text editor immediately.

moorscode:
Acceptance:
When using single quotes everything works fine.
When using double quotes the keyword is not matched.

Tested on featured image with alt text and keyword set to: `I told him "single quotes work"`

PR #283, opened 2016-02-11
Expose everything as a module

atimmer:
Removes all prototypes from the global scope and puts them into modules, `browser.js`, which is browserified, exposes them to the global scope for BC.


terw-dan:
CR and acceptance :+1:

PR #286, opened 2016-02-17
Introduces valueobject

terw-dan:
This PR introduces a valueObject to pass arguments to all analyses. 
This object is now a basic name:value object, but will be extended with methods for checks we want to do in this object. 
The valueObject replaces this.config, and is introduced this way to keep backwards compatibility.


atimmer:
Closed in favour of https://github.com/Yoast/YoastSEO.js/pull/288

PR #287, opened 2016-02-18
Introduced the Score value object

jcomack:
This is a first step into migrating the scoring mechanism.


atimmer:
CR & Acceptance done 👍🏻

PR #288, opened 2016-02-18
Introduced the Paper value object

jcomack:
This will be responsible for mananging the keyword, text, meta description etc.


js/values/Paper.js
```diff
@@ -0,0 +1,30 @@
...
+var Paper = function( keyword ) {
+	this.keyword = keyword || "";
+};
+
+/**
+ * Check whether a keyword is available
+ * @returns {boolean}
+ */
+Paper.prototype.hasKeyword = function() {
+	return ( typeof this.keyword !== "undefined" && this.keyword !== "" );
```
atimmer: This is redundant because of line 7. Furthermore, I rather have an `isEmpty` check here.

js/values/Paper.js
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Return the associated keyword or an empty string if no keyword is available
+ * @returns {string}
+ */
+Paper.prototype.getKeyword = function() {
+	if ( this.hasKeyword() ) {
+		return this.keyword;
+	}
+
+	return "";
```
atimmer: This is redundant because of line 7.

js/analyzer.js
```diff
@@ -45,6 +46,13 @@ Analyzer.prototype.checkConfig = function() {
  * YoastSEO.Analyzer initialization. Loads defaults and overloads custom settings.
  */
 Analyzer.prototype.init = function( args ) {
+
+	if ( typeof args.Paper === "undefined" ) {
+		args.Paper = new Paper( args.keyword );
+	}
+
+	this.Paper = args.Paper;
```
atimmer: We should lowercase to start object properties with.

atimmer:
We should replace all occurrences of `this.config.keyword` in the analyser with `this.paper.getKeyword()`.

atimmer:
CR Done 👷🏻

atimmer:
CR & Acceptance done 👍🏻

PR #291, opened 2016-02-19
Remove diacritics in subheadings when matching keyword

### LINKED ISSUE #101: It can't find words with special characters in the content.
I encounter this error on this post: http://www.scroll.com.tr/8-gb-ramli-kapakli-telefon-geliyor-mu/

The focus keyword is "kapaklı". As you can see it contains a special character (dotless i, Turkish character). Because of dotless i, analyze page shows that there is no focus keyword in the first paragraph but you can see it in the first paragraph.

Here is two screenshots from admin panel. First one is the General tab. It says there is 3 occurences of focus keyword in the content and it's right.

general_tab

But on Page Analysis tab it says that "The keyword doesn't appear in the first paragraph of the copy, make sure the topic is clear immediately."

page_analysis_tab

If I change the focus keyword from "kapaklı" to "telefon", I get a green score, but without change it's orange.

I hope it's clear enough.

Some data:
WordPress version: 4.3.1 (single site)
Yoast SEO Premium version: 2.3.5
Server operationg system: CentOS
Web servers: Apache, nginx (proxy)
PHP version: 5.3.3

terw-dan:
When checking for a keyword in a subheading the diacritics weren't removed, so they didn't match the keyword when that has a diacritic (since it was being replaced there). This makes sure the keyword is matched in the subheading if the keyword has a diacritic. 

Fixes #101


jcomack:
CR done 👍

jcomack:
Acceptance done 👍

PR #292, opened 2016-02-22
Introduces metadescription to paper

terw-dan:
Adds a metaValues object to the paper. This now contains the metadescription, but can contain other meta values (like URL, pageTitle etc). 
meta values are added as 1 object in stead of arguments to reduce the number of arguments, since the meta values aren't necessary to evaluate a paper.


js/values/Paper.js
```diff
@@ -1,15 +1,18 @@
 /**
  * Construct the Paper object and set the keyword property
- * @param {string} keyword
+ * @param {string} keyword The keyword to use for analyses.
+ * @param {object} metas The object containing all metavalues.
  * @constructor
  */
-var Paper = function( keyword ) {
+var Paper = function( keyword, metas ) {
```
jcomack: As far as I know, metas isn't a word. Suggestion: use metadata instead. This is automatically more descriptive as well.

js/analyzer.js
```diff
@@ -67,9 +67,12 @@ Analyzer.prototype.checkConfig = function() {
  * YoastSEO.Analyzer initialization. Loads defaults and overloads custom settings.
  */
 Analyzer.prototype.init = function( args ) {
+	var metaValues = {
```
jcomack: This object isn't being used anywhere else. Might as well directly inject it into Paper as such:

`args.paper = new Paper( args.keyword, { metaDescription: args.meta } )` or just pass the `args.meta` and let Paper handle the additional wrapping of `args.meta`.

jcomack:
CR done. Please see comments.

js/analyzer.js
terw-dan: You're right, but because there will be more arguments in the metaData object for readability it will be clearer to define them in an object. This way we can also change the metaData in a valueobject if we want to.

atimmer: I vote for `args.paper = new Paper( args.keyword, { metaDescription: args.meta } )`. 🚲🏠

js/values/Paper.js
```diff
@@ -1,25 +1,35 @@
+var isUndefined = require( "lodash/lang/isUndefined" );
```
CarolineGeven: please document this var

js/values/Paper.js
```diff
@@ -35,10 +45,26 @@ Paper.prototype.hasText = function() {
...
  */
 Paper.prototype.getText = function() {
 	return this._text;
 };
 
+/**
+ * Check whether a metaDescription is available
+ * @returns {boolean} Returns true if metaDescription isn't empty
+ */
+Paper.prototype.hasMetaDescription = function() {
```
CarolineGeven: Why would you need to check if metaDescription is available when getMetaDescription returns an empty string if there is none? It seems like this is duplicate now.

CarolineGeven:
cr done

js/values/Paper.js
terw-dan: This is a dependency, we've decided that these don't need to be documented.

js/values/Paper.js
terw-dan: There is has function for every value so we can use this as a very clean and easy way in our code to check if a certain value is set or not.

jcomack:
Acceptance done :+1:

PR #297, opened 2016-02-24
introduces attributes object to paper

### LINKED ISSUE #294: Introduce object containing arguments
Currently certain, less important, arguments are directly being passed into Paper.
It would be clearer to use a named object instead.

terw-dan:
Creates an attributes object with defaults for keyword and description. 

Fixes #294


js/values/Paper.js
```diff
@@ -1,22 +1,14 @@
...
-
-	// if there is no metaData object given, use an empty object so checks on metaData keep working
-	if ( isUndefined( metaData ) ) {
-		metaData = {};
-	}
-
-	this._metaDescription = metaData.metaDescription || "";
+	this._attributes = attributes || {};
+	this._keyword = this._attributes.keyword || "";
+	this._metaDescription = this._attributes.metaDescription || "";
```
jcomack: This 'check' should probably be handled somewhere else in the code, seeing as it's not a hard requirement.

jcomack:
CR done :+1:

js/values/Paper.js
```diff
@@ -1,22 +1,14 @@
...
+var Paper = function( text, attributes ) {
 	this._text = text || "";
-
-	// if there is no metaData object given, use an empty object so checks on metaData keep working
-	if ( isUndefined( metaData ) ) {
-		metaData = {};
-	}
-
-	this._metaDescription = metaData.metaDescription || "";
+	this._attributes = attributes || {};
```
atimmer: We could also handle this with `defaultsDeep`

terw-dan: I've used `defaults` now, since we only have a single level, so `defaultsdeep` isn't necessary.

jcomack:
CR 2 done :+1:

jcomack:
Acceptance done. Will merge once builds are completed.

PR #298, opened 2016-02-24
Introduced the Collection data type object

jcomack:
This will allows for Assessors to have a Collection of assessments that need to be processed.

### DIFF
--- a/js/datatypes/Collection.js
+++ b/js/datatypes/Collection.js
@@ -1,4 +1,4 @@
-/* global modules */
+/* global module */
 var isUndefined = require( "lodash/lang/isUndefined" );
 
 /**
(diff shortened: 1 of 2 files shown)


js/datatypes/Collection.js
```diff
@@ -0,0 +1,131 @@
...
+ */
+Collection.prototype.removeItem = function( index ) {
+	if ( indexExists( this._items, index ) === false ) {
+		return false;
+	}
+
+	this._items.splice( index, 1 );
+	this._length = calculateLength( this._items );
+
+	return true;
```
terw-dan: If you remove the item where the current iterator is pointed to, the iterator will point the a next item. Something we should check for.

terw-dan:
CR 1 Done

jcomack:
After discussing with @atimmer, we came to the conclusion that it might be easier to use a native `Array` until `Set` methods are available in Babel or something similar to Babel.

atimmer:
We can close this, the code will be preserved.

PR #316, opened 2016-03-04
Fix highlighting of keywords with diacritics

### LINKED ISSUE #314: title highlighting doesn't work in snippetpreview
When using the keyword Slovníček pojmû it doesn't highlight in the snippet preview.
image

It is found in the analysis (The page title contains the focus keyword, at the beginning which is considered to improve rankings.).

terw-dan:
Keywords with diacritics are now highlighted in the snippetpreview, this wasn't done because the regex always removed the diacritics, which should not be done for highlighting, or the keyword won't match. 

Fixes #314


moorscode:
CR done :+1:

jcomack:
Acceptance failed. Highlighting doesn't seem to work.

jcomack:
Acceptance done :+1:

PR #319, opened 2016-03-07
Fixes undefined grey bullet

### LINKED ISSUE #318: no keyword isn't handled correctly in paper
When a keyword is given a grey bullet shows up with undefined

image

This seems to come from 'no keyword' when checking the screenreader span of the bullet.

terw-dan:
Fixes #318
The check for the undefined score was wrong so an undefined score was parsed. This fixes that.


jcomack:
CR done :+1:

jcomack:
Acceptance done :+1:

PR #329, opened 2016-03-14
Fixes indenting for jslint

terw-dan:
This fixes the indenting that generate warnings in JSlint. Mostly added an extra space after // and changed space-indenting to tab indenting. 
This fixes a bunch of warnings in jslint.


atimmer:
CR Done and no functionality has been changed.

PR #332, opened 2016-03-14
Introduced the Researcher

### LINKED ISSUE #326: Create Researcher object
Currently all 'Researches' are called within the Analyzer class. These calls need to be moved to their own class and rewritten to support the new way of analyzing the text.

jcomack:
Applied modifications where necessary and refactored some of the code in the assessment

Fixes #326


js/researcher.js
```diff
@@ -0,0 +1,92 @@
...
+
+	this.customResearches = {};
+};
+
+/**
+ * Update the Paper associated with the Researcher.
+ * @param {Paper} paper The Paper to use within the Researcher
+ * @throws {InvalidTypeError} Parameter needs to be an instance of the Paper object.
+ */
+Researcher.prototype.updatePaper = function( paper ) {
```
atimmer: Shouldn't we call this `validatePaper`?

js/app.js
```diff
@@ -412,6 +422,15 @@ App.prototype.runAnalyzer = function() {
 
 	this.analyzerData.keyword = keyword;
 
+	// The new researcher
+	if ( isUndefined( this.researcher ) ) {
+		this.researcher = new Researcher( this.paper );
+		this.pluggable._addPluginTests( this.researcher );
+	} else {
+		this.researcher.updatePaper( this.paper );
+		this.pluggable._addPluginTests( this.researcher );
```
atimmer: This won't work on the researcher, the function `_addPluginTests` tries to call a `addAnalysis` function on the object passed as an argument, but that function doesn't exist on the Researcher.

js/researcher.js
```diff
@@ -0,0 +1,92 @@
...
+ * @returns {Object} An object containing all available researches.
+ */
+Researcher.prototype.getAvailableResearches = function() {
+	return merge( this.defaultResearches, this.customResearches );
+};
+
+/**
+ * Return the Research by name.
+ * @param {string} name The name to reference the research by.
+ * @returns {boolean|callback}
```
atimmer: I think this should be `{*}`

terw-dan: missing jsdoc

jcomack: Doesn't that imply that anything could be returned?

js/app.js
jcomack: Noted. Will remove this call.

spec/assessments/countWordAssessmentSpec.js
```diff
@@ -36,7 +37,8 @@ var repeatString = function(string, times) {
 describe( "A word count assessment", function(){
 	it( "assesses a single word", function(){
 		var mockPaper = new Paper( "sample" );
-		var assessment = wordCountAssessment( mockPaper, i18n );
+		var researcher = new Researcher( mockPaper );
```
atimmer: Currently with this test we are testing both the word count research and the word count assessment. We could only test the assessment by mocking the researcher:

```
var researcher = {
    getResearch: function() { return 1; }
};
```

js/researcher.js
jcomack: You're not 'validating' anything here. What happens is that the Paper object gets updated whenever text has been updated in the interface. Because the Researcher isn't continuously updated, this method was introduced.

atimmer: Fair enough, maybe use the more standard `setPaper`?

terw-dan:
CR done

terw-dan:
CR2 done

terw-dan:
Acceptance done :+1:

PR #345, opened 2016-03-18
Altered stringToRegex

jcomack:
Now it uses default values as a fallback. Also changed the double-negative dontReplaceDiacritics with a doReplaceDiacritics which defaults to true


atimmer:
CR Done :+1:

terw-dan:
Acceptance :+1:

PR #381, opened 2016-03-23
Removes score for metadescription keyword if there is no metadescription

terw-dan:
The researcher returns -1 when there is no meta description. If there is no metadescription we cannot score for keywords in the metadescription.


jcomack:
CR done :+1:

PR #385, opened 2016-03-25
Improves the hasScore for the assessmentresult.

terw-dan:
Adds a boolean isScoreSet to the assessmentResult. This way we can check if a score is set, and it is possible to set a score to 0.

### DIFF
--- a/js/values/AssessmentResult.js
+++ b/js/values/AssessmentResult.js
@@ -29,7 +29,7 @@ AssessmentResult.prototype.getScore = function() {
 
 /**
  * Set the score for the assessment.
- * @param {number|string} score The score to be used for the score property
+ * @param {number} score The score to be used for the score property
  * @returns {void}
  */
 AssessmentResult.prototype.setScore = function( score ) {
--- a/js/values/AssessmentResult.js
+++ b/js/values/AssessmentResult.js
@@ -7,6 +7,7 @@ var isString = require( "lodash/isString" );
  * @constructor
  */
 var AssessmentResult = function() {
+	this.isScoreSet = false;
 	this.score = 0;
 	this.text = "";
 };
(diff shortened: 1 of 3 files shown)


js/values/AssessmentResult.js
```diff
@@ -29,14 +29,15 @@ AssessmentResult.prototype.getScore = function() {
 
 /**
  * Set the score for the assessment.
- * @param {number|string} score The score to be used for the score property
+ * @param {number} score The score to be used for the score property
  * @returns {void}
  */
 AssessmentResult.prototype.setScore = function( score ) {
-	if ( !isNumber( score ) && !isString( score ) ) {
+	this.isScoreSet = true;
```
atimmer: Why not?

```js
if ( !isNumber( score ) ) {
    this.score = score;
    this.isScoreSet = true;
}
```

js/values/AssessmentResult.js
```diff
@@ -1,12 +1,12 @@
 var isUndefined = require( "lodash/isUndefined" );
 var isNumber = require( "lodash/isNumber" );
-var isString = require( "lodash/isString" );
 
 /**
  * Construct the AssessmentResult value object.
  * @constructor
  */
 var AssessmentResult = function() {
+	this.isScoreSet = false;
```
atimmer: Why not call this `hasScore`, the method is called this way.

atimmer:
CR Done 👷🏻

js/values/AssessmentResult.js
terw-dan: changed this to 
if ( isNumber( score ) ) {
    this.score = score;
    this.isScoreSet = true;
}

js/values/AssessmentResult.js
terw-dan: renamed it to _hasScore :)

atimmer:
CR 2 & Acceptance done 👍🏻

PR #393, opened 2016-03-29
Introduces Assessment requirements

jcomack:
This allows us to define requirements on an Assessment that will be checked by the Assessor prior to running the Assessment.


js/assessor.js
```diff
@@ -71,35 +71,56 @@ Assessor.prototype.getAvailableAssessments = function() {
...
 /**
+ * Checks whether or not the Assessment is applicable.
+ * @param {Object} assessment The Assessment object that needs to be checked.
+ * @param {Paper} paper The Paper object to check against.
+ * @param {Researcher} [researcher] The Researcher object containing additional information.
+ * @returns {boolean} Whether or not the Assessment is applicable.
+ */
+Assessor.prototype.isApplicable = function( assessment, paper, researcher ) {
+	if ( assessment.hasOwnProperty( "isApplicable" ) ) {
+		return assessment.isApplicable( paper, researcher ) === true;
```
atimmer: `=== true` isn't necessary here.

js/assessor.js
```diff
@@ -71,35 +71,56 @@ Assessor.prototype.getAvailableAssessments = function() {
...
 	this.results = [];
-	forEach( assessments, function( assessment, assessmentName ) {
-		this.results.push(
-			{
-				name: assessmentName,
-				result: assessment( paper, researcher, this.i18n )
-			}
-		);
+
+	forEach( assessments, function( assessment, name ) {
```
atimmer: This could be much neater if we use `filter` and `map` here.

```
var applicableAssessments = filter( assessments, function( assessment ) {
    return this.isApplicable( assessment, paper, researcher );
}.bind( this ) );

this.results = map( applicableAssessments, function( assessment, name ) {
    return {
        name: name,
        result, assessment.getResult( paper, researcher, this.i18n )
    };
}.bind( this ) );

atimmer:
CR & Acceptance done.

PR #396, opened 2016-03-29
Check for empty string in keyword density

terw-dan:
Fixes a bug where an empty string could give errors. 
If after sanitizing the string was empty (ie there is only an imagetag in the text) it generated an error. 

Now when the text is empty and a keyword is entered, it no longer generates javascript errors.


spec/researches/keywordDensitySpec.js
```diff
@@ -23,5 +23,7 @@ describe("Test for counting the keyword density in a text", function(){
 		expect( keywordDensity( mockPaper ) ).toBe( "0.0" );
 		mockPaper = new Paper( "a string of text with the key&word in it, density should be 7.7%", {keyword: "key&word"} );
 		expect( keywordDensity( mockPaper ) ).toBe( "7.7" );
+		mockPaper = new Paper( "<img src='http://plaatje.com/plaatje.png'>", {keyword: "key&word"} );
```
jcomack: Please use English examples in tests, as some of the other users / people that will be working with this don't know what "plaatje" means :)

jcomack:
CR done.

PR #399, opened 2016-03-29
Snippet editor open close buttons better accessibility

### LINKED ISSUE #398: Snippet editor open close buttons better a11y
See https://github.com/Yoast/wordpress-seo/issues/4054

afercia:
In order to avoid a focus loss on the snippet editor open/close buttons I'd propose to keep the buttons always visible. Properly set an `aria-expanded` attribute. At this point, I'd remove the "Snippet editor" heading which looks redundant since the button above is always visible and already says the same thing.

Fixes #398


terw-dan:
CR Done
@afercia, could you fix the issues in the build, there are a few places where you use single quotes, the codestyle of YoastSEOjs requires to use double quotes everywhere. If you change these, the build should no longer fail

afercia:
@terw-dan done, hopefully :)

terw-dan:
CR done :+1:

terw-dan:
Acceptance :+1:

PR #404, opened 2016-03-30
Adds locale check to fleschreading

### LINKED ISSUE #402: Add locale check for fleschreading
We need to add a check for the locale on the fleschreading so it will run if the paper has an english locale

terw-dan:
Adds a locale to the fleschreading assessment so it only executes when it has an english locale.

Fixes #402


omarreiss:
CR :+1:

jcomack:
Acceptance done :+1:

PR #408, opened 2016-03-31
Modulize scoreformatter

### LINKED ISSUE #400: Split up score formatter
@atimmer requested that the Score Formatter gets split up in smaller modules to support assessments implemented in the other plugins.

jcomack:
This PR introduces an AssessorPresenter class for generating the ratings (both the individual and overall rating) and thus deprecates ScoreFormatter.

This also splits off some of the scoring code into its own module, allowing it to be used throughout the entire code base, if necessary.

Fixes #400

### DIFF
--- a/js/interpreters/scoreToRating.js
+++ b/js/interpreters/scoreToRating.js
@@ -1,8 +1,8 @@
 /**
  * Interpreters a score and gives it a particular rating.
+ *
  * @param {Number} score The score to interpreter.
  * @returns {string} The rating, given based on the score.
- * @constructor
  */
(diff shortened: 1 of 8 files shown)


js/interpreters/scoreToRating.js
```diff
@@ -0,0 +1,28 @@
+/**
+ * Interpreters a score and gives it a particular rating.
+ * @param {Number} score The score to interpreter.
+ * @returns {string} The rating, given based on the score.
+ * @constructor
```
atimmer: This is not a constructor, it's a pure function.

PR #412, opened 2016-04-04
Updates scoring in the links assessment

terw-dan:
Creates a seperate assessment for the countlinks and the links statistics. 
The linkstatistics is only fired when a keyword is entered. 

Fixes https://github.com/Yoast/wordpress-seo/issues/3212


js/researches/countLinks.js
```diff
@@ -0,0 +1,16 @@
...
+
+var getLinks = require( "./getLinks" );
+
+/**
+ * Checks a text for anchors and returns the number found.
+ *
+ * @param {object} paper The paper object containing text, keyword and url.
+ * @returns {number} The number of links found in the text.
+ */
+module.exports = function( paper ) {
```
andizer: You only need a text, why passing a whole object?

andizer:
Cr done, one question.

js/researches/countLinks.js
terw-dan: The paper is always passed to a research, this is done because is research is called by the researcher and always passes the paper as an argument.

jcomack:
Acceptance failed, `countWords` is never executed.

PR #417, opened 2016-04-05
Use the passed paper to analyze the keyword

atimmer:
This makes sure we can test different papers with the same assessor.


moorscode:
CR Done :+1:

PR #427, opened 2016-04-07
Add a separate Taxonomy count words assessment

jcomack:
This PR also alters some of the functionality of the `Assessor`, allowing us to extend it and remove assessments and also allows us to inject a different `Assessor`, if need be.


examples/standalone/example-plugin-test.js
```diff
@@ -2,7 +2,7 @@
 	addEventListener( "load", function() {
 		// Wait for YoastSEO to be loaded
 		setTimeout(function() {
-			addPlugin();
+//			addPlugin();
```
atimmer: Should not commit this.

atimmer:
CR Done 👍🏻

PR #430, opened 2016-04-07
Adds keyword requirement for used keywords plugin

### LINKED ISSUE #421: Focus keyword
When not setting a focus keyword, I expect the content analysis to tell me there's no focus keyword and do no further keyword related analysis.

### What happened instead?
The content analysis alerted me that there's no focus keyword set and also informed me that I've used this focus keyword 2 times before.
schermafbeelding_2016-04-06_om_10_32_07

Do not set a focus keyword for a post.

### Technical info
* Platform: WordPress
* Platform version: 4.4.2
* Yoast SEO version: Beta 3.2-1

terw-dan:
The previously used keywords plugin in the bundled plugins was always fired because there was no isApplicable set to this assessment. 

Added this so it will only trigger when keyword is present. 

Fixes #421


atimmer:
Acceptance done 👍🏻

PR #448, opened 2016-04-15
Improve the snippet preview elements accessibility

### LINKED ISSUE #127: Snippet preview is confusing for a screenreader
For example, in the following screenshot the screenreader will just say "Twitter analytics, twitter analytics, twitter analytics" without it being clear what's it all about.
screen shot 2015-11-03 at 16 12 46

afercia:
Adds screen reader text to explain what the snippet preview elements are.
Also, if we want to use a `section` HTML element (actually not sure about this but I'm not so worried about abstract semantic issues) I think the heading should be inside the section.

First pass, not sure about a couple things. Would greatly appreciate some feedback :)

Fixes #127


css/_snippet-editor.scss
```diff
@@ -115,7 +115,7 @@ $color_buttons: #555555;
 	}
 
 	&__preview {
-		padding: 15px 20px;
+		padding: 0 20px  15px;
 	}
```
terw-dan: remove double space

terw-dan:
CR done, removed 1 double space.

diedexx:
### Acceptance:
Somehow I get the following when trying to read the "Seo title preview:" span;




---

I'll leave it up to you if you want to address this in this pull, but the progress does not have a value set to it for screen readers.

adding `aria-valuenow="[value]" aria-valuemin="0" aria-valuemax="65"` might be a good solution.

(using VoiceOver in both safari and chrome)

afercia:
@diedexx thanks! About the progress bars, there's a specific issue I've opened on the wpseo repo, guess I should move it here :) https://github.com/Yoast/wordpress-seo/issues/4135

afercia:
About the screen reader text announced as 3 separate words, it really depends on couple of things:
- on VoiceOver, it happens when using arrow down/up to navigate. Doesn't happen when using Control-Option-Right Arrow (or Leeft Arrow) to navigate
- the screen reader CSS class has a 1 x 1 pixels size, sometimes screen readers understand this and read out words as they would appear visually: they would wrap in a new line. Adding `white-space: nowrap;` would fix this but may lead to other unexpected results.

afercia:
Below, how the 3 snippet preview elements get read out using Firefox + NVDA:

snippet preview 127 nvda

I'd say what is missing here is a clear explanation about what happens when clicking on the elements. They're announced as "clickable" but users don't have a clue what the click is meant for.

Instead of adding further text to each element, I'd consider to add a short hidden text explanation  after the "Snippet preview" heading. Something like:
```
<h3>Snippet preview</h3>
<p class="screen-reader-text">
You can click on each element to jump to the Snippet Editor.
</p>
```

terw-dan:
CR Done :+1:

terw-dan:
Acceptance: If I use NVDA in Firefox (Windows) it reads all the texts from the snippetpreview as I would suspect, but the "You can click on each element to jump to the Snippet Editor."-text is skipped. 
Should NVDA read these as well?

afercia:
@terw-dan yes it should read it, using arrow down to navigate trough all elements, right after "headlng level 3 Snippet preview". It works for me on Win 10, Firefox + NVDA.

terw-dan:
Acceptance done :+1: 
It works on my Windows 10 system with firefox and nvda. 

Could someone test this on a OSX system as well?

js/templates.js
```diff
@@ -38,7 +38,7 @@
   var undefined;
 
   /** Used as the semantic version number. */
-  var VERSION = '4.6.1';
+  var VERSION = '4.11.1';
```
moorscode: This seems like a big version jump, develop version is at 4.7.0 at the moment.

moorscode:
Acceptance done on OSX
+ Merged develop for merge conflicts

PR #457, opened 2016-04-21
Remove %% from the text strings

atimmer:
It is very confusing for translators to have to keep the double
percentage-signs in the string and this has bitten us in the 3.2 release
of Yoast SEO.


terw-dan:
CR Done :+1:

atimmer:
I've tested this.

PR #465, opened 2016-04-26
Add new translatable string for the Content tab. First pass.

### LINKED ISSUE #462: Consider to use new translatable strings for the Content tab traffic light
See related: https://github.com/Yoast/wordpress-seo/issues/4055
See also the related PR https://github.com/Yoast/wordpress-seo/pull/4297

The traffic light tab link needs text to report the current SEO score, the related PR is currently concatenating two strings:
- a fixed part "Content:" 
- the SEO score string from the "Assessor presenter" config `presenter.js`

I think it is not too bad, since the first string ends with a colon. By the way, concatenation should be avoided when it comes to translatable strings. Maybe a better solution would be adding new, extended, strings to `presenter.js` e.g.:
```
"Content: Focus keyword not set"
"Content: Bad SEO score"
"Content: OK SEO score"
"Content: Good SEO score"
```
These new strings would make translators happy. On the other hand, it is very likely they would be used in just one place.
I don't have a strong opinion about this, I think the first solution is not ideal but could work, the second one is better but maybe overkill. Would appreciate some l10n thoughts :)

afercia:
First pass for the new, fully translatable string to be used for the Content tab.
- adds a new `fullText` string
- changes the `feedback` string in "No SEO score available"
- makes the "k" in the "Ok" score string uppercase
Not sure about some things, and have no idea about what to do with Jasmine tests :) any feedback more than welcome.
Fixes #462


languages/yoast-seo.pot
```diff
@@ -396,21 +420,37 @@ msgid ""
...
-#: js/config/presenter.js:14
-msgid "Bad SEO score"
+#: js/config/presenter.js:11
+msgid "Content Analysis: No SEO score available"
 msgstr ""
 
-#: js/config/presenter.js:18
-msgid "Ok SEO score"
+#: js/config/presenter.js:15
+msgid "Bad SEO scorexx"
```
moorscode: Strange 'xx' addition?

moorscode:
Could you also merge `develop` into this PR for merge ability?

afercia:
Ah! :) Sorry for the "xx", debugging leftover.

atimmer:
@afercia Awesome, only thing this will go wrong is that the text in the feedback key is also used in the list of feedback. I think the only assessment this is currently used in is the stop words assessment, which checks if there are stop words in the keyword.

I think we should split `feedback` into `feedback` and `noScore`.

afercia:
OK, I see now the `<span class="screen-reader-text">Feedback</span>`. Will do, thanks.

afercia:
@atimmer I've realized that https://github.com/Yoast/wordpress-seo/pull/4390/files is going to hardcode a specific `screenReaderText ` string for the "no keyword" case, see [line 580](https://github.com/Yoast/wordpress-seo/pull/4390/files#diff-2345f544b22b4e0069fccb2eceada862R580)
Following that logic should we hardcode also a `fullText` string and don't worry to split `feedback`?

atimmer:
I think that is best for now.

js/config/presenter.js
```diff
@@ -7,19 +7,23 @@ module.exports = function ( i18n ) {
 	return {
 		feedback: {
 			className: "na",
-			screenReaderText: i18n.dgettext( "js-text-analysis", "Feedback")
+			screenReaderText: i18n.dgettext( "js-text-analysis", "Feedback"),
+			fullText: i18n.dgettext( "js-text-analysis", "Content Analysis: Needs feedback")
```
atimmer: This should be "Has feedback" instead of "Needs feedback".

atimmer:
CR Done 👍🏻

terw-dan:
Acceptance done :+1:

PR #470, opened 2016-04-28
Fixes scoring of alttags

terw-dan:
When there are multiple images in a text, but only 1 has a keyword, it was reported that images did not have the keyword. Since it isn't necessary for every image to have the keyword in the alt-tag, this pull fixes the order of scoring.


andizer:
CR done :+

jcomack:
Acceptance done 👍

PR #486, opened 2016-05-09
Rename the pageTitle to metaTitle

### LINKED ISSUE #386: Way to analyze custom metabox?
I have a metabox with `Subtitle` and I want the `Focus Keyword` to match either the post title or post subtitle to suppress this warning:

> The focus keyword 'any-title' does not appear in the page title.

I couldn't find any API in the docs, care to provide any example to pass custom metabox value through Page Analyzer?

andizer:
The pageTitle wasn't used consequently. Now, it should be done properly. Also the naming was kind of confusing: In fact it's the meta title. Thus `pageTitle` is renamed into `metaTitle`.

Fixes #386


atimmer:
@andizer You haven't actually fixed #386, because https://github.com/Yoast/YoastSEO.js/blob/AM/rename-pageTitle-to-metaTitle/js/app.js#L496 is still the same.

jcomack:
Acceptance done.

While testing the code, I noticed that after I had registered a plugin that modifies the title, the modification is applied to the `data.title` property and thus the keyword assessment fails. 
When I modify this to be applied to the `data.pageTitle`, omit a page title in WordPress and set the keyword to be a word that is in my custom plugin, the assessment succeeds.
 
Additionally, I noticed that none of the content in the snippet preview is being update. I'd expect the meta title to update to reflect the modification that my plugin introduces and display said modification.

Example plugin used: 
```
(function () {
	var module = {
		// Initialize
		init: function() {
			jQuery(window).on('YoastSEO:ready', module.load);
		},

		// Load plugin and add hooks.
		load: function() {
			YoastSEO.app.registerPlugin('examplePlugin', {status: 'ready'});
			YoastSEO.app.registerModification( 'title', function( oldTitle ) {
				var newTitle = oldTitle + 'extra title';
				
				return newTitle;
			}, 'examplePlugin' );

			module.bind();
		},

		// Update Yoast SEO analyzer when fields are updated.
		// Use debounce technique, which triggers refresh only when keys stop being pressed.
		bind: function() {
			var timeout;

			function refresh() {
				clearTimeout(timeout);
				timeout = setTimeout(function() {
					YoastSEO.app.refresh();
				}, 250);
			}
		}
	};

	module.init();
})();
```

atimmer:
Acceptance done 👍🏻

PR #489, opened 2016-05-09
Fixed a rounding issue and altered the result text slightly.

### LINKED ISSUE #477: We should update the output string of the sentence length assessment
We should add rounding to percentage in output string of sentence length assessment
We should also update the string from 'within the recommended range' to 'less than the recommended maximum of...'

jcomack:
Text is now more uniform. Also, no longer will the percentage be something along the lines of `33.333333`.

Fixes #477 

Small side-note: I noticed the following test result being expected, which seems off to me.

```
expect( assessment.getText() ).toEqual ( "25% of the sentences contain more than 20 words, which is less than the recommended maximum of 25%." );
```

@IreneStr, could you please elaborate on why this is the way it is?

### DIFF
--- a/js/assessments/sentenceLengthInTextAssessment.js
+++ b/js/assessments/sentenceLengthInTextAssessment.js
@@ -19,19 +19,21 @@ var calculateSentenceLengthResult = function( sentences, i18n ) {
 	var score = calculateTooLongSentences( percentage );
 
 	if ( score >= 7 ) {
-		return{
+		return {
 			score: score,
 
-			// Translators: %1$s expands to number of sentences.
-			text: i18n.sprintf( i18n.dgettext( "js-text-analysis", "%1$s of the sentences contain more than %3$s words, " +
+			// Translators: %1$s expands to the percentage of sentences, %2$d expands to the maximum percentage of sentences.
+			// %3$s expands to the recommended amount of words.
+			text: i18n.sprintf( i18n.dgettext( "js-text-analysis", "%1$s of the sentences contain more than %3$d words, " +
 				"which is less than the recommended maximum of %2$s." ), percentage + "%", maximumPercentage + "%", recommendedValue )
 		};
 	}
-	return{
+	return {
 		score: score,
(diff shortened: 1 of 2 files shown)


js/assessments/sentenceLengthInTextAssessment.js
```diff
@@ -13,25 +14,26 @@ var calculateSentenceLengthResult = function( sentences, i18n ) {
...
 	var score = calculateTooLongSentences( percentage );
 
 	if ( score >= 7 ) {
 		return{
 			score: score,
 
-			// translators: %1$s expands to number of sentences.
-			text: i18n.sprintf( i18n.dgettext( "js-text-analysis", "%1$s of the sentences contain more than 20 words, " +
-				"which is within the recommended range." ), percentage + "%" )
+			// Translators: %1$s expands to number of sentences.
```
terw-dan: Please add translator comments for %3$s, also shouldn't this be %3$d, since we only put a digit in there.

js/assessments/sentenceLengthInTextAssessment.js
```diff
@@ -13,25 +14,26 @@ var calculateSentenceLengthResult = function( sentences, i18n ) {
...
+				"which is less than the recommended maximum of %2$s." ), percentage + "%", maximumPercentage + "%", recommendedValue )
 		};
 	}
 	return{
 		score: score,
 
-		// translators: %1$s expands to number of sentences.
-		text: i18n.sprintf( i18n.dgettext( "js-text-analysis", "%1$s of the sentences contain more than 20 words, " +
-			"which is more than the recommended maximum of %2$s. Try to shorten your sentences." ), percentage + "%", maximumPercentage + "%" )
+		// Translators: %1$s expands to number of sentences.
```
terw-dan: translator comments should be expanded with %2$s and %3$s

terw-dan:
CR 1 done.

terw-dan:
CR done :+1:

PR #490, opened 2016-05-10
Use debounce to refresh the analyzer

### LINKED ISSUE #487: Deprecate analyzeTimer and use debounce in refresh function
analyzeTimer is a horrible interface. We should just call refresh straight away from anywhere, make sure refresh internally debounces analysis runs and call refresh from everywhere.

andizer:
- Overwrite `refresh` function with `debounce( refresh ) ` call.
- Deprecate the analyzeTimer function
- Moved the deprecated function to the bottom and specify the version since it is deprecated.

Fixed #487


jcomack:
CR done :

jcomack:
Acceptance done 👍

PR #492, opened 2016-05-10
Add identifier to all assessments

atimmer:
This makes it possible to refer to them later when implementing decorations/marks.

How to acceptance test this: Check if the standalone version still works.


andizer:
CR done 👍

atimmer:
I need this and it's trivial, so I am merging this.

PR #498, opened 2016-05-12
Add two helpers to determine if value is too long or too short

### LINKED ISSUE #467: Calculation of textValue too long or too small should be done in helpers
In my 'fight' to reduce code duplication I would suggest to replace the methods `isParagraphTooLong`, `isTextTooLong` and probably some other ones (that are doing the same) in one single helper function like `isValueTooLong` 

```
**
 * The function to use as a filter for too long paragraphs.
 * @param {number} recommendedValue The recommended maximum length of the text value.
 * @param {number} valueLength The number of words within a text value.
 * @returns {boolean} Returns true if valueLength exceeds recommendedValue.
 */
 var isValueTooLong = function( recommendedValue, valueLength ) {
    return valueLength > recommendedValue;
 };
```

The opposite function of this that is calculating if a 'value' is too short, should also become a helper function.

andizer:
Replaces the existing functions with different naming but same logic.

Fixes #467

### DIFF
--- a/js/assessmentHelpers/checkForTooLongSentences.js
+++ b/js/assessmentHelpers/checkForTooLongSentences.js
@@ -1,14 +1,5 @@
 var filter = require( "lodash/filter" );
-
-/**
- * Checks if a sentence is longer than the recommended value.
- * @param {number} recommendedValue The recommended maximum length of the sentence.
- * @param  {number} sentenceLength The length of the sentence.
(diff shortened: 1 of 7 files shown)


terw-dan:
CR: 
can you fix the build, and also include this in de paragraphsTooLong Assessment

terw-dan:
CR done :+1:

terw-dan:
Acceptance :+1:

PR #501, opened 2016-05-13
Change page title into SEO title

andizer:
It is confusing when there is a title field (in wordpress) and an SEO field. But the SEO field is the one being used.


atimmer:
👍🏻

PR #502, opened 2016-05-13
Add framework for marking text based on markers

### LINKED ISSUE #485: Create decoration framework
As a user, I want to be able to click an assessment result and have the issue marked in the text I am editing.

We need a framework that makes it simple to implement for every assessment.

We need to be able to do this in tinyMCE for MVP.

Eventually we also want this for plain text area's and CKEditor.

atimmer:
The implementation of the actual marking of the text can be passed to
the App with the marker argument. Each assessment can have a `getMarks`
function that returns a list of marks for a given paper.

Implement the first marker in the subheading length assessment.

With this is should be possible to add a `getMarks` function to an assessment and be able to mark it in the text.

Fixes #485


atimmer:
CR by @terw-dan, @IreneStr and @jcomack.

Acceptance @terw-dan, @IreneStr and @jcomack.

PR #506, opened 2016-05-18
Add removeMarks function to remove added marks

atimmer:
This function is able to remove marks added by `addMark`.

I also moved both functions to a `markers` folder so they are grouped together.


terw-dan:
CR Done :+1:

andizer:
'Acceptance' done 👍

PR #513, opened 2016-05-19
Hide progress bars from screen readers

### LINKED ISSUE #466: Snippet editor progress bars accessibility
Moving here issue wrongly opened on wordpress-seo, see https://github.com/Yoast/wordpress-seo/issues/4135

As also @diedexx commented, see https://github.com/Yoast/YoastSEO.js/pull/448#issuecomment-213405785 the progress bars value is not announced by VoiceOver. Reporting here the screenshot from #448 :

236550f0-0896-11e6-9852-d9c5c8a5081f

While the implementation is technically correct, seems not all browsers / screen readers combinations get it right. The native HTML5 `<progress>` element with `value` and `max` attributes should suffice because it already carries the necessary semantics and information. But, while webkit supports the `<progress>` element, probably VoiceOver doesn't yet. In my testing, to make VoiceOver work with both Safari and Chrome it is necessary to add a `role="progressbar"` and the aria attributes @diedexx mentioned.

The reason why it announces "indeterminate" is because it needs the "max" value set as attribute (not as property) to be present since the beginning.

Worth noting the value announced is the result of the internal calculation screen readers do based on the min, max and current value so we can't solve the rounding issue and values will be announced with decimals, unless we base all on a scale from 0 to 100, maybe. Needs more testing with other screen readers.

Screenshot of my testing so far:

screen shot 2016-04-27 at 19 09 50

Still to evaluate:
- not sure if setting values as properties is OK, for example `element.value = value` same for "max"
- add an aria label to explain "what" these progress bar are about
- maybe add an aria-live region to announce the progress value when users stop typing

afercia:
Tested using Safari+VoiceOver, Firefox+NVDA, IE 11 and JAWS. The progress bars are now completely ignored by screen readers. People who use screen readers will be better server with the planned `aria-live` messages for the Content Analysis.
Fixes #466


atimmer:
CR Done 👍🏻

PR #518, opened 2016-05-23
Use a custom mark element to mark the document

atimmer:
This makes sure we can always delete all marks safely from a given text.
With <mark>-tags we aren't sure if we are the ones that added the tags
or other parties added them.


terw-dan:
CR 1 Done. 
Please fix the build.

andizer:
CR 2 Done

andizer:
Acceptance done 👍

PR #522, opened 2016-05-23
Fixes '/' in focus keyword

rensw90:
Updated regex so it matches slashes in the focus keyword.

Fixes https://github.com/Yoast/wordpress-seo/issues/3839


atimmer:
CR Done 👍🏻

terw-dan:
Acceptance :+1:

PR #532, opened 2016-05-24
Add save content score callback

atimmer:


atimmer:
Signed off by @moorscode

PR #534, opened 2016-05-25
Implemented a new marker button and CSS

### LINKED ISSUE #503: Improve styling of mark buttons
We should use the following icon for the buttons: http://fontawesome.io/icon/eye/

They should stay blue when that specific thing has been marked in the text.

jcomack:
Modularized some of the JavaScript and CSS as well so it can be used more easily throughout the code base.

Fixes #503


rensw90:
CR Done!

PR #535, opened 2016-05-26
Make removeMarks more robust

atimmer:
It should now remove the marks even when they are jumbled by tinyMCE or
if other HTML attributes are added.

You can test this by running the unit tests


terw-dan:
CR Done :+1:

terw-dan:
Acceptance :+1:

Issue #483, opened 2016-05-09
Create transliteration module or use existing one.


omarreiss:
To be able to transliterate special characters in URL's as requested in #293, we need a module that can transliterate special characters in strings.

### Our requirements
- We need a module that takes a string and a locale and returns the transliterated version of the string, based on that locale.
- We want the module to strive for full coverage of every available locale. Of course it's fine if the module doesn't cover everything yet, but modules that were created only to fix it for i.e. cyrillic are not interesting to us (except to compare our own data with).
- We don't want defaults or reuse of characters in any way, shape or form. Transliteration tables should be specific for every language.

### Reuse or build our own?

It seems there are already many transliteration modules. Still, they all seem to be reinventing the wheel and many are not transparent about where they base their transliteration tables on. All of them seem to do it for a couple of languages, but not for all. Some have a lot of languages, but are not localized as well or are incomplete for some languages. 

If we can find a module that fits our requirements, then we should use that. If we can find a module that fits nearly all the requirements, we should consider if we can contribute to it in a way that gets us very far very fast. 

There are quite some advantages to owning the module ourselves:

- We can develop it more rapidly.
- We can control the specification.
- We can control the way it is packaged and modularized.

Since transliteration tables are not likely subject to a lot of change, the risk of high maintenance is relatively low.

PR #536, opened 2016-05-26
Introduce transliterations

### LINKED ISSUE #531: Match keyword with transliterations
Everywhere where we match with a keyword we should match the transliterated keyword as well.

terw-dan:
Fixes #531 

This prepares for the use of transliterations of keyword matching. On the places we currently use replace diacritics we now can match with transliterations. Currently the transliteration match is the same as the default keyword, since we don't have the transliterations themselves. These will be fixed in #483 and can be included in the matchStringWithTransliteration.


js/stringProcessing/matchTextWithTransliteration.js
```diff
@@ -0,0 +1,48 @@
...
+	var keywordRegex = toRegex( keyword );
+	var matches = text.match( keywordRegex ) || [];
+
+	text = text.replace( keywordRegex, "" );
+
+	var transliterateKeyword = transliterate( keyword );
+	var transliterateKeywordRegex = toRegex( transliterateKeyword );
+	var transliterateMatches = text.match( transliterateKeywordRegex ) || [];
+
+	var combinedArray = [].concat( matches, transliterateMatches );
```
atimmer: Why not `matches.concat( transliterateMatches)`?

js/stringProcessing/addWordboundary.js
```diff
@@ -5,7 +5,7 @@
  *
  * @param {string} matchString The string to generate a regex string for.
  * @param {string} [extraWordBoundary] Extra characters to match a word boundary on.
```
andizer: Why are there `[]` around `extraWordBoundary` ???

terw-dan: This indicates an optional parameter

js/stringProcessing/findKeywordInUrl.js
```diff
@@ -13,7 +15,7 @@ module.exports = function( url, keyword ) {
 
 	if ( formatUrl !== null ) {
 		formatUrl = formatUrl[ 0 ].replace( /<.*?>\s?/ig, "" );
-		if ( formatUrl.match( keywordRegex( keyword ) ) !== null ) {
+		if ( matchTextWithTransliteration( formatUrl, keyword ).length > 0 ) {
```
andizer: I would return `( matchTextWithTransliteration( formatUrl, keyword ).length > 0 )` directly instead of doing and instead of `return keywordFound` in the end of the function just `return false`;

js/stringProcessing/matchTextWithTransliteration.js
andizer: What about this feedback @terw-dan ???

andizer:
CR done

js/stringProcessing/addWordboundary.js
andizer: Okay 👍

andizer:
CR 2 Done 👍

andizer:
Acceptance done.

PR #538, opened 2016-05-27
Add transliterations

### LINKED ISSUE #483: Create transliteration module or use existing one.

IreneStr:
Add transliterations for:

- Spanish
- Polish
- German
- Nynorsk
- Bokmål
- Swedish
- Finnish
- Danish
- Turkish
- Latvian
- Icelandic
- Faroese
- Czech
- Russian
- Esperanto
- Afrikaans
- Catalan
- Asturian
- Aragonese
- Aymara
- English
- French
- Italian
- Dutch
- Bambara
- Ukrainian

Fixes #483 

For testing check the unit tests.


js/config/transliterations.js
```diff
@@ -0,0 +1,622 @@
...
+ * The function returning an array containing transliteration objects, based on the given locale.
+ *
+ * @param {string} locale The locale.
+ * @returns {Array} An array containing transliteration objects.
+ */
+module.exports = function( locale ) {
+	if ( isUndefined( locale ) ) {
+		return [];
+	}
+		switch( getLanguage( locale ) ) {
```
andizer: Indentation it too much.

andizer:
CR done. One big feedback thing:
I would choose for a big array with all the transliterations. This array should be outside the module.exports thing. 

```js
var transliterations = { en : { }  }; 

module.exports = function() {
    if ( isUndefined( locale ) ) {
        return [];
    }

    var language = getLanguage( locale );
 
    if ( check here if the language exists in the transliterations object ) {
        return transliterations[ language ]; 
    }

    return [];
```

andizer:
CR done 👍

rensw90:
Acceptance done 💯

PR #539, opened 2016-05-27
Mark duplicate words correctly

atimmer:
Apply exactly one <yoastmark> to each word. Do this by removing duplicated marks from the array of marks and then applying each mark globally in the text instead of on one occurence.

How to test this: Have duplicate words in your text with more than 4 syllables and test if they are all marked correctly with one <yoastmark> element.


andizer:
CR done

andizer:
Acceptance done.

PR #546, opened 2016-05-27
Update transition word list

### LINKED ISSUE #525: Check transition word list
We should check the list of transition words for words that should be removed because they often are not really a transition word ('about' etc)

IreneStr:
Removes words from the transition word list that only function as transition words in very specific situations. Also removes 'or' because of its very high frequency, causing high transition word percentages for almost every text. (For that same reason 'and' was already excluded from the list.)

Removed words:
about,  absolutely, again, definitely, eternally, expressively, expressly, extremely, immediately, including, instantly, namely, naturally, next, not only, notably now, or, ordinarily, positively, truly, ultimately, uniquely, usually, where

Added word: 
'after that' as replacement for 'after' (because 'after' is only in specific situations a transition word, but 'after that' always indicates a transition)

Fixes #525


terw-dan:
CR Done

rensw90:
Acceptance done 👍

PR #554, opened 2016-05-30
Catch exception thrown by assessments

### LINKED ISSUE #456: Gracefully fail on assessment failure
If an assessment throws an error such as happened in #... we should gracefully fail and not show that assessment or show an error instead of breaking the complete JavaScript.

This makes our code much more robust.

atimmer:
How to test this: Introduce an error in an assessment and make sure the
assessment is now a gray cirlce with an error message the trace is in
the console and the rest of the content analysis still works.

Fixes #456

### DIFF
--- a/js/assessor.js
+++ b/js/assessor.js
@@ -157,8 +157,8 @@ Assessor.prototype.executeAssessment = function( paper, researcher, assessment )
 
 		result.setScore( 0 );
 		result.setText( this.i18n.sprintf(
-			this.i18n.dgettext( "js-text-analysis", "An error occured in the '%1$s' assessment, " +
-				"the exact error can be found in the console." ),
+			/* translators: %1$s expands to the name of the assessment. */
+			this.i18n.dgettext( "js-text-analysis", "An error occured in the '%1$s' assessment" ),
 			assessment.identifier,
 			assessmentError
 		) );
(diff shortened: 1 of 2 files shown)


js/assessor.js
```diff
@@ -125,19 +126,45 @@ Assessor.prototype.assess = function( paper ) {
...
 			result.setMarker( this.getMarker( assessment, paper, researcher ) );
 		}
+	} catch ( assessmentError ) {
+		showTrace( assessmentError );
+
+		result = new AssessmentResult();
+
+		result.setScore( 0 );
+		result.setText( this.i18n.sprintf(
+			this.i18n.dgettext( "js-text-analysis", "An error occured in the '%1$s' assessment, " +
```
moorscode: Do we want to confuse people with "the console"?
I guess just putting it in the console and leaving it there for support to ask for it would be a bit more user friendly.

js/helpers/errors.js
```diff
@@ -0,0 +1,23 @@
+var isUndefined = require( "lodash/isUndefined" );
+
+/**
+ * Shows and error trace of the error message in the console if the console is available.
+ *
+ * @param {string} [errorMessage=""] The error message.
+ */
+function showTrace( errorMessage ) {
+	if ( isUndefined( errorMessage ) ) {
+		errorMessage = "";
```
moorscode: Why not just exit when there is no message?

moorscode:
CR Done :construction_worker:

js/helpers/errors.js
atimmer: Because I want to be able to call this function without an error message and still put a trace in the console.

moorscode:
CR Done 👍

terw-dan:
Acceptance :+1:

PR #556, opened 2016-05-30
Mark only when having marks

### LINKED ISSUE #528: Don't show mark button when there is nothing to mark
When there is nothing to mark, the mark button is being visible also. We should hide these buttons.

yoast_seo_content_analysis

andizer:
Only show the markers when there is something to mark.

Fixes #528

Testing:
See the specs for how to test

### DIFF
--- a/js/values/AssessmentResult.js
+++ b/js/values/AssessmentResult.js
@@ -17,6 +17,7 @@ var emptyMarker = function() {
 var AssessmentResult = function() {
 	this._hasScore = false;
 	this._identifier = "";
+	this._shouldMark = false;
 	this._marker = emptyMarker;
 	this.score = 0;
 	this.text = "";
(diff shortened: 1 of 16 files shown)


js/values/AssessmentResult.js
```diff
@@ -17,6 +17,7 @@ var emptyMarker = function() {
 var AssessmentResult = function() {
 	this._hasScore = false;
 	this._identifier = "";
+	this._shouldMark = false;
```
atimmer: I think we should to with `_hasMarks` instead of `_shouldMark`. You have already intuitively named the variable in the assessment `hasMarks` so it feels right to name it that way everywhere.

atimmer:
CR Done 👍🏻

terw-dan:
Acceptance :+1:

PR #573, opened 2016-05-31
Massively improve performance of passive voice

atimmer:
The passive voice makes extensive use of giant list of words that need
to be transformed to regexes. We were transforming these each time we
needed the regex. Now we cache them using lodash's memoize.

This can be much improved over time by shipping a data structure that is
already optimized instead of relying on the memory to save time.

To test this: Just make sure everything still work, I have already
tested this. If you really want to test this: Grab a giant text and put
it in the standalone version.


moorscode:
CR Done 👍
Acceptance Done 👍 
Awesomeness added 👍 
💯  points for improvement.

PR #584, opened 2016-06-01
Icon eye buttons focus and active style

### LINKED ISSUE #571: Icon eye marks buttons focus and active style
The new "icon eye" buttons to set or remove the marks in the editor need a focus style for accessibility. I'd propose to use the same "circular focus" style used for the Help buttons. Something like this:

screen shot 2016-05-31 at 16 57 38

This would require to make the clickable area a bit bigger (which is always a good thing) and then the positioning of the buttons should be adjusted a bit I guess.

afercia:
First pass for new icon-eye buttons style. Fixes #571 

We may need to adjust the position and vertical alignment of the buttons. The buttons are taller now and in order to align the "eye" with the text on the left we need some margin between the list items. Notice that when used in WordPress, list items already inherit some bottom margin. By the way, when the assessment text is in just one line, the margin between items can be uneven:

screen shot 2016-06-01 at 13 07 41
 
One way to fix this would be making the buttons "overlap" a bit (backgrounds in the screenshot below are just for testing purposes):

screen shot 2016-06-01 at 13 06 20

Or maybe try to make the buttons a bit smaller, but I'd rather accept the slightly uneven margin.

### DIFF
--- a/css/_analysis.scss
+++ b/css/_analysis.scss
@@ -92,8 +92,9 @@ li.score {
 		&.icon-eye {
 			&-inactive, &-active {
 				border: 0;
-				width: 16px;
-				height: 16px;
+				width: 28px;
+				height: 28px;
+				padding: 4px;
 				outline: none;
 				background: no-repeat center;
 				background-size: 16px;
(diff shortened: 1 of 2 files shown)


afercia:
Preview in the metabox:

screen shot 2016-06-01 at 13 37 12

forgot to remove the list top padding.

afercia:
There's the need to have predictable element sizes so we should reset the white space between inline elements. Switched to floats, which is easier and used `calc()` for the score text width. Open to suggestions for other methods, maybe in the future we could use flexbox. Screenshot to highlight the elements size:

screen shot 2016-06-01 at 15 43 48

andizer:
CR done 👍

andizer:
Acceptance done

PR #598, opened 2016-06-03
Optimized find transition words and optimized word boundary code

jcomack:
This shaves off a lot of time when dealing with a large amount of sentences (see screenshots)

before

after


andizer:
👍

PR #628, opened 2016-06-07
Split paragraphs on heading tags

### LINKED ISSUE #576: Paragraph detection is broken
The detection of paragraphs is broken:

schermafbeelding 2016-06-01 om 09 24 44
schermafbeelding 2016-06-01 om 09 25 01

Text in picture:
``<h3>1. Evaluate your categories</h3>``
As your blog is growing, it could well be that you’re blogging about one thing far more than other things. That’s just the way blogging goes. You should therefore critically evaluate your categories every few months, asking yourself whether or not one category is growing much faster than another category. If some parts of your blog are growing much faster than other parts, you could divide such a category in two separate categories.

Imagine yourself having an Italian food blog for example with the categories pasta, salads, meat dishes and desserts. You should critically evaluate these categories every now and then. It could well be that the category pasta becomes a really large category, because you love spaghetti so much. Perhaps you could make a separate spaghetti category, if you tend to write a lot of spaghetti posts. You would then have five categories: pasta, spaghetti, meat dishes and desserts.
``<h3>2. Add sub categories and tags</h3>``
If you have lots and lots of posts, it becomes harder to make sure you link to (all) similar content. There will be too many posts to choose from. As a result of that, a lot of blog posts will ‘get lost’ in a structure that is too flat. There will be few links to these posts, making it hard to be found by both your audience and Google!

### LINKED ISSUE #621: Incorrect sentence detection in case of parentheses
Parenthesis in the middle of a sentence to be not viewed as a sentence terminator.

### What happened instead?
schermafbeelding 2016-06-07 om 14 11 08

Write a sentence containing a transition word (for example 'because'). Write somewhere after the transition words something between parentheses. Hit the mark transition words button. 

### Technical info
* Platform: WordPress
* Platform version: 4.5.2
* Yoast SEO version: trunk

atimmer:
Fixes #576

### DIFF
--- a/js/helpers/html.js
+++ b/js/helpers/html.js
@@ -5,19 +5,19 @@ var inlineElements = [ "b", "big", "i", "small", "tt", "abbr", "acronym", "cite"
 	"samp", "time", "var", "a", "bdo", "br", "img", "map", "object", "q", "script", "span", "sub", "sup", "button",
 	"input", "label", "select", "textarea" ];
 
-var blockElementsRegex = new RegExp( "^(" + blockElements.join( "|" ) + ")$", "im" );
-var inlineElementsRegex = new RegExp( "^(" + inlineElements.join( "|" ) + ")$", "im" );
+var blockElementsRegex = new RegExp( "^(" + blockElements.join( "|" ) + ")$", "i" );
+var inlineElementsRegex = new RegExp( "^(" + inlineElements.join( "|" ) + ")$", "i" );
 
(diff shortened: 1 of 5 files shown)


jcomack:
CR done 👍

terw-dan:
Acceptance:
If I console.log the output of the paragraphs, it sees header tags as a seperate paragraph. We should not treat header tags as seperate paragraphs. Maybe these should be filtered out?

terw-dan:
Acceptance :+1:

PR #631, opened 2016-06-07
Alignment of the icon

andizer:
Align the background 3pixels from the left instead of in the center.

Fixes https://github.com/Yoast/wordpress-seo/issues/4795

### DIFF
--- a/css/_analysis.scss
+++ b/css/_analysis.scss
@@ -137,7 +137,7 @@ li.score:after {
 		padding: 4px;
 		border-radius: 100%;
 		outline: none;
-		background: url(svg-icon-eye-slash($color_marker_inactive)) no-repeat center;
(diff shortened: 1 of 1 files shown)


terw-dan:
CR done :+1:

terw-dan:
Acceptance:
Alignment is better, but it is still a bit off if the top item has a marker. This isn't aligned properly (probably due to the clear all marks icon above it). 
image

If the first item is only a single line, it has a much larger spacing than the other lines.

andizer:
As far as I can see, the screenshots represents the old situation. Did you do a `grunt build` ?

terw-dan:
Yes I did, the alignment of the icons is improved, only the top item is a bit strange.

atimmer:
I think we should close this in favor of https://github.com/Yoast/YoastSEO.js/pull/632

andizer:
I agree!

PR #635, opened 2016-06-08
Fix block detection with newlines

atimmer:


atimmer:
Signed off by @moorscode

PR #648, opened 2016-06-08
RW/empty content wrong score

rensw90:
Resolves the problem when no content score is showed in the publish box with an empty post.

Testing:
1. Click on add new post.
2. Check if the content score in the publish box shows: `na`


terw-dan:
CR Done :+1:

PR #653, opened 2016-06-09
Fixes passive detection

### LINKED ISSUE #605: Passive voice doesn't handle text between quotation marks well
It looks like quotation marks interfere with passive voice detection. 

In the following sentence, the research should detect 'get lost' as a passive:
`As a result of that, a lot of blog posts will 'get lost' in a structure that is too flat.`
However, it doesn't.

terw-dan:
This pull fixes the passive detection. Only the last subsentence was pushed into an array with passives in stead of all of them, so a sentence was only passive if it was the last part. 

Includes another fix where the index is determined on a lowercased sentence.

For testing: See the added specs for an example of passive sentences. 

Fixes #605 (the issue wasn't the quotation marks).

### DIFF
--- a/js/researches/getPassiveVoice.js
+++ b/js/researches/getPassiveVoice.js
@@ -314,7 +314,6 @@ module.exports = function( paper ) {
 			passive = determinePassives( subSentence );
 			if ( passive === true ) {
 				passiveSentences.push( sentence );
-				passive = false;
 			}
 		} );
 	} );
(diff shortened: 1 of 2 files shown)


js/researches/getPassiveVoice.js
```diff
@@ -311,11 +312,11 @@ module.exports = function( paper ) {
 		var passive = false;
 		forEach( subSentences, function( subSentence ) {
 			passive = determinePassives( subSentence );
+			if ( passive === true ) {
+				passiveSentences.push( sentence );
+				passive = false;
```
IreneStr: This is probably unnecessary

IreneStr:
CR done

IreneStr:
CR 2 done 👍

IreneStr:
Acceptance done 💯

PR #661, opened 2016-06-09
Make sure the sentence only gets added to the list once

moorscode:
Added specs for multiple subsentences with passive and non-passive content
Optimized forEach when a subsentence is passive to stop doing hard work for following subsentences


andizer:
CR done, please resolve the conflicts.

terw-dan:
Acceptance :+1:

PR #663, opened 2016-06-09
Introduce accessible tooltips for the mark buttons

### LINKED ISSUE #623: Give mark button a title attribute / popup thingy
Some text to indicate what pressing this button would do.

Paging @afercia  :)

afercia:
Moving here the tooltips initially implemented in https://github.com/Yoast/wordpress-seo/pull/4813. First time handling Sass files here on YoastSEO.js please let me know if I'm doing something wrong :)

Tooltips implemented just on the mark buttons for now, and in the "South" position. Example inside the metabox:

screen shot 2016-06-09 at 14 34 32

Fixes #623

### DIFF
--- a/css/_tooltips.scss
+++ b/css/_tooltips.scss
@@ -2,6 +2,11 @@
 	position: relative;
 }
 
+button.yoast-tooltip {
+	/* IE 11 needs overflow: visible on buttons. */
+	overflow: visible;
+}
+
 .yoast-tooltip::after {
 	display: none;
 	position: absolute;
--- a/css/_tooltips.scss
+++ b/css/_tooltips.scss
@@ -38,16 +38,6 @@
 	pointer-events: none;
 }
 
-@-webkit-keyframes yoast-tooltip-appear {
-	from {
-		opacity: 0;
-	}
-
-	to {
-		opacity: 1;
-	}
-}
-
 @keyframes yoast-tooltip-appear {
 	from {
 		opacity: 0;
--- a/css/_tooltips.scss
+++ b/css/_tooltips.scss
@@ -1,0 +1,224 @@
+.yoast-tooltip {
+	position: relative;
+}
+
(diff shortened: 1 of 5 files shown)


afercia:
To do:
- test in IE
- even if it's not required for now it would be nice to test the "multiline" tooltip

css/_tooltips.scss
```diff
@@ -0,0 +1,224 @@
...
+	width: 0;
+	height: 0;
+	border: 5px solid transparent;
+	opacity: 0;
+	color: rgba(0,0,0,0.8);
+	content: "\00a0";
+	pointer-events: none;
+}
+
+@-webkit-keyframes yoast-tooltip-appear {
```
andizer: This isn't necessary because the grunt-task `auto-prefixer` will add prefixes on `build:css`

css/_tooltips.scss
```diff
@@ -0,0 +1,224 @@
...
+	from {
+		opacity: 0;
+	}
+
+	to {
+		opacity: 1;
+	}
+}
+
+@keyframes yoast-tooltip-appear {
```
andizer: This isn't necessary because the grunt-task `auto-prefixer` will add prefixes on `build:css`

andizer:
Two small comments :) 👍

afercia:
Found an easy way to hide the tooltips when the mark button is "active". GH does a similar thing for the tooltips, check for example the "+" Create new... icon on the top right.

afercia:
Was not working on IE 11, scratched my head for a couple of hours. Fun. When tooltips are used on buttons, IE 11 needs an `overflow: visible` on the button itself. Reduced test case: http://s.codepen.io/afercia/debug/YWwNGM

Tested also with screen readers:
- Firefox + NVDA
- IE 11 + JAWS
works as expected.

terw-dan:
Acceptance :+1:

PR #679, opened 2016-06-10
Disable 'got' in passive voice irregular list

### LINKED ISSUE #678: We should disable 'got' in the passive voice irregulars list
We should disable 'got' in the passive voice irregulars list to avoid false positives. 
See #677.

IreneStr:
Disables 'got' in the passive voice irregular list.

Also adds uppercase to translator comments, and adds comments to passive voice spec indicating what is the passive part of the sentences.

Fixes #678


andizer:
Cr done

rensw90:
Acceptance done 👍

PR #681, opened 2016-06-10
Simplify assessment to ranges

atimmer:
This removes the scores inside the ranges but in the end the only thing that matters is a `green`, `orange` or `red` indicator.


andizer:
CR done 👍

terw-dan:
Acceptance :+1:

PR #684, opened 2016-06-10
Add an assessment that checks if you have content

atimmer:
This makes sure we always have something in the analysis.


atimmer:
To test this: If you have more than 50 characters the assessment shouldn't be in the results.

jdevalk:
CR: 👍

PR #693, opened 2016-06-12
Changes subheadings assessments and scoring

### LINKED ISSUE #692: Change subheading assessment workings
Right now, when you have a text with 1 subheading, you get 3 assessment results, whereas you get 1 result when you have no subheadings. This means your overall score could be lower because you have 1 subheading then when you have 0. That's odd.

Fix:
- don't use `subheadingPresence` at all;
- don't use `subheadingLength`;
- use `subheadingsTooLong`, which is based on the distribution, as the overall check. Make it return 3 when there are _no_ subheadings. 

This means we'll have 1 instead of 3 checks for subheadings.

jdevalk:
Fixes #692 by disabling two subheadings scores and merging the presence one into the distribution one.

Also: 

- changed scoring to a slightly more balanced method.
- changed `NegativePoints` to `PenaltyPoints`. Because they're positive numbers, because "Penalty" describes what they are a bit better and matches the terminology used in our explainer post.

### DIFF
--- a/js/assessments/fleschReadingEaseAssessment.js
+++ b/js/assessments/fleschReadingEaseAssessment.js
@@ -26,7 +26,7 @@ var calculateFleschReadingResult = function( fleschReadingScore, i18n ) {
 
 	if ( inRange( fleschReadingScore, 70, 80 ) ) {
 		return {
-			score: 8,
+			score: 9,
 			resultText: i18n.dgettext( "js-text-analysis", "fairly easy" ),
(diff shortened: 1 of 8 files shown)


atimmer:
CR Done 👍🏻

rensw90:
Acceptance done, everything seems ok. Just one test fails, if this is fixed the PR can be merged.

rensw90:
Acceptance done 👍

PR #703, opened 2016-06-13
Handle refreshes more intelligently

atimmer:
Instead of debouncing for a very long period we now apply both a
debounce of 400ms and a throttle of 3000ms. This means that the
analysis will only run once every 3 seconds, but will also feel
'instantly' to users that change their focus keyword. The 400ms makes
sure that the analysis isn't run on the first keystroke.

Fixes https://github.com/Yoast/wordpress-seo/issues/4831

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -10,6 +10,7 @@ var MissingArgument = require( "./errors/missingArgument" );
 var isUndefined = require( "lodash/isUndefined" );
 var forEach = require( "lodash/forEach" );
 var debounce = require( "lodash/debounce" );
+var throttle = require( "lodash/throttle" );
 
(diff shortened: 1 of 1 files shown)


js/app.js
```diff
@@ -347,6 +350,10 @@ App.prototype.getData = function() {
  * @returns {void}
  */
 App.prototype.refresh = function() {
+	this._pureRefresh();
+};
+
+App.prototype._pureRefresh = function() {
```
terw-dan: Please add documentation.

terw-dan:
CR done.

terw-dan:
CR Done :+1:

terw-dan:
Acceptance :+1:

PR #720, opened 2016-06-15
Don't apply transition and passive to non-english

### LINKED ISSUE #717: Hide passive voice and transition word assessment if locale is not en_
The passive voice and transition word assessment to be hidden if the locale is not en_.

### What happened instead?
They were not hidden.

Set your locale to something that is not English. Look at the content assessor.

atimmer:
These assessment only make sense for english.

Fixes #717


terw-dan:
CR Done :+

terw-dan:
Acceptance :+1:

PR #721, opened 2016-06-15
RW/add nomarker option

rensw90:
Fixes https://github.com/Yoast/wordpress-seo/issues/4807
Requires: https://github.com/Yoast/wordpress-seo/pull/4902

Testing:
1. Create a post and write some text so some marker buttons show up.
2. Test if the marker buttons are disabled in text editor and enabled in visual editor.
3. Also test if the marker buttons are disabled when you refresh the page, with the text editor enabled. They also need to re-enable when you again switch to the visual editor.


jcomack:
Acceptance done 👍

Issue #675, opened 2016-06-10
Images in gallery recognized as outbound links


tacoverdo:
# Problem description
When a gallery (WP default, no plugin) is added to a page, the images are considered to be outbound links. 

# Example
![edit_post_ _clean_ _wordpress](https://cloud.githubusercontent.com/assets/5147598/15961210/d1812db0-2f03-11e6-9d19-bf0283e9b3fb.png)

# Steps to reproduce
* Create a new post.
* Add a gallery (Add media -> Create Gallery -> Select Images -> Create Gallery -> Insert Gallery).
* Check the content analysis.

### Technical info
* Platform: WordPress
* Platform version: 4.5.2
* Yoast SEO version: 3.3-beta5

PR #722, opened 2016-06-16
Competing keywords execution being executed on links to self

jcomack:
Implemented a check to exclude external links to itself from being assessed for competing keywords

Fixes https://github.com/Yoast/wordpress-seo/issues/4348
Fixes https://github.com/Yoast/YoastSEO.js/issues/675


js/researches/getLinkStatistics.js
```diff
@@ -21,6 +22,37 @@ var keywordInAnchor = function( keyword, anchor, locale ) {
...
+};
+
+/**
+ * Matches the URL retrieved from an anchor with the current url.
+ *
+ * @param {string} url The URL to look for.
+ * @param {string} currentUrl The current URL to match against.
+ * @returns {boolean} Whether or not the URLs match.
+ */
+var matchAnchorUrlToCurrentUrl = function( url, currentUrl ) {
```
atimmer: I would call this `isLinkToSelf`

js/stringProcessing/extractUrl.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * Retrieves all URLs from a string, based on the found anchors.
```
atimmer: Why would you try to match all href's in a URL, you are only throwing one anchor in this function so wouldn't it make more sense to only match once and just return '' or the matched url?

atimmer:
CR Done 👷🏻

jcomack:
Feedback has been processed. Regarding `extractUrl.js`, @atimmer and I came to an agreement to leave this as is for the time being.

diedexx:
CR Done 👍

terw-dan:
Acceptance:

My post has the URL: `http://local.wordpress.dev/banana` and keyword `banana`
If I link to this url with the keyword `banana` I still get the error I am linking to another page with my keyword.

jcomack:
Thanks to @terw-dan we found out that the getLinkStatistics wasn't matching against a proper URL (only a slug).

Added functionality to the Wordpress SEO plugin that gets a permalink that needs to be used to fix this PR. Please review https://github.com/Yoast/wordpress-seo/pull/4998 prior to attempting this PR.

rensw90:
CR-Done 👍

rensw90:
Acceptance done, couldn't get it to work. Used the same input as Danny mentioned before.

terw-dan:
CR Done :+1: 
Fixed the build by using the correct quotes.

boblinthorst:
Acceptance: Still doesn't seem to work. Used same testcase as previous testers. (http://local.wordpress.dev/banana/ with keyword banana and linking to this url from an instance of the keyword in the content.

boblinthorst:
acceptance ok 👍

PR #725, opened 2016-06-20
Increases negative point treshold for contentscore 

terw-dan:
To make sure we don't give too many penalty points when only a few tests are run, this PR increases the treshold a bit on bad scores. 

With 3 scores; 
1 bad score and 1 ok score should result in a red bullet.
1 bad score and 2 good scores should result in an orange bullet.
2 ok scores and a good score should result in an orange bullet.
1 ok score and 2 good scores should result in a green bullet. 

I increased the penalty from a bad score from 3 to 4, and increased the total allowed penaltypoints so with 2 ok scores you have 4 penaltypoints, which scores an orange bullet, in stead of a red bullet. 

For testing: Make sure to test this with a non-english locale

Fixes #https://github.com/Yoast/wordpress-seo/issues/4957

- update, created a seperate score for non-english, so the english scoring will remain completely unchanged.

### DIFF
--- a/js/contentAssessor.js
+++ b/js/contentAssessor.js
@@ -66,7 +66,7 @@ ContentAssessor.prototype.calculatePenaltyPoints = function () {
 		// Convert the ratings to negative 'points'.
 		switch ( rating ) {
 			case "bad":
-				weight = 3;
+				weight = 4;
 				break;
 
 			case "ok":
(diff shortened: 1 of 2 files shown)


atimmer:
CR Done 👍🏻

atimmer:
Acceptance done 👍🏻

PR #732, opened 2016-06-27
Refactored some of the code to support disabling of analyses

jcomack:
This applies to both content and SEO analysis.

This means that things like Assessors, Presenters etc. don't get loaded unless these are marked as being active.


js/app.js
```diff
@@ -264,6 +254,46 @@ var App = function( args ) {
 	this.refresh();
 };
 
+App.prototype.getDefaultOutputElement = function( args ) {
+	if ( args.keywordAnalysisActive ) {
```
atimmer: You should add `keywordAnalysisActive` to the default arguments of the `App`.

spec/appSpec.js
```diff
@@ -78,19 +78,6 @@ describe( "Creating an App", function(){
 		});
 	});
 
-	it( "throws on a missing output element ID", function() {
```
atimmer: When changing the behaviour you shouldn't delete the test case that tested this behaviour, instead you should change the expected behaviour of the test.

In this case the spec has changed from "throws on a missing output element ID" -> "works when no output element ID has been passed".

js/app.js
```diff
@@ -264,6 +257,46 @@ var App = function( args ) {
...
+	}
+
+	if ( args.contentAnalysisActive ) {
+		return args.targets.contentOutput;
+	}
+
+	return "";
+};
+
+App.prototype.initializeAssessors = function( args ) {
```
terw-dan: Missing JSdoc

js/app.js
```diff
@@ -264,6 +257,46 @@ var App = function( args ) {
...
+	if ( args.keywordAnalysisActive ) {
+		this.initializeSEOAssessor( args );
+	}
+
+	if ( args.contentAnalysisActive ) {
+		this.initializeContentAssessor( args );
+	}
+};
+
+App.prototype.initializeSEOAssessor = function( args ) {
```
terw-dan: Missing JSdoc

js/app.js
```diff
@@ -264,6 +257,46 @@ var App = function( args ) {
...
+	// Set the assessor
+	if ( isUndefined( args.seoAssessor ) ) {
+		this.seoAssessor = new SEOAssessor( this.i18n, { marker: this.config.marker } );
+	} else {
+		this.seoAssessor = args.seoAssessor;
+	}
+};
+
+App.prototype.initializeContentAssessor = function( args ) {
```
terw-dan: Missing JSdoc

js/app.js
```diff
@@ -264,6 +257,46 @@ var App = function( args ) {
 	this.refresh();
 };
 
+App.prototype.getDefaultOutputElement = function( args ) {
```
terw-dan: Missing JSdoc

terw-dan:
CR Done

terw-dan:
CR done :+1:

boblinthorst:
Acceptance tests: mergeable with knowledge of some minor bugs

PR #753, opened 2016-07-04
Assess title based on width in pixels in stead of characters

### LINKED ISSUE #740: Make sure title width is measured based on pixels
The pixels should be leading on the progress bars and in the analysis. If this is not yet the case, fix it.

terw-dan:
This measures the title in pixels, so we can assess based on the viewable area in stead of number of characters, which is less accurate. 

Fixes #740

This changes the bars in the snippetpreview to change based on the width of the title, not on the number of characters used. 

For testing: Create a page title should be between 400 and 600 px, above 600px is too long.


js/app.js
```diff
@@ -475,6 +475,7 @@ App.prototype.runAnalyzer = function() {
 		description: this.analyzerData.meta,
 		url: this.analyzerData.url,
 		title: this.analyzerData.metaTitle,
+		titleWidth: this.snippetPreview.data.titleWidth,
```
atimmer: Why not introduce a `getTitleWidth` function on the snippet preview instead of grabbing it from the internals of the snippet preview?

js/values/Paper.js
```diff
@@ -103,6 +104,22 @@ Paper.prototype.getTitle = function() {
 };
 
 /**
+ * Check whether an title width in pixels is available
+ * @returns {boolean} Returns true if the Paper has a title.
+ */
+Paper.prototype.hasTitleWidth = function() {
+	return this._attributes.titleWidth !== "";
```
atimmer: This line is not covered by the specs

atimmer:
CR Done 👍🏻

boblinthorst:
Acceptance ok 👍

PR #757, opened 2016-07-04
Change feedback string from keyword density assessment

### LINKED ISSUE #751: Improve keyword density string in case of 0.0% density
When the keyword density is 0%, the following message appears: 
schermafbeelding 2016-07-04 om 09 05 54

We should change 'which is a bit low' to something stronger.

IreneStr:
Change keyword density feedback string from 
`The keyword density is %1$s, which is a bit low.`
to
`The keyword density is %1$s, which is too low.`

This results in a less sarcastic comment in case of a 0% keyword density.

Fixes #751


terw-dan:
CR + Acceptance :+1:

PR #772, opened 2016-07-08
Strip numbers before counting in flesch reading

### LINKED ISSUE #741: Handle digits correctly in Flesch Reading Ease
Currently, digits are included in the word count, but counted as 0 syllables. 

This leads to unwanted consequences in the Flesch Reading Ease formula, that contains both word count and syllable count:
FLE (English) = 206.835 - 1.015( <sup>total words</sup>/<sub>total sentences</sub> ) - 84.6( <sup>total syllables</sup>/<sub>total words</sub> ).

Therefore, for the Flesch Reading Ease we should exclude digits from the word count as well. 

In all other assessments digits could remain included in the word count.

terw-dan:
We stripped the numbers after counting the number of words. Since we don't want to include them in the FleschReading at all, it makes more sense to remove them before counting the words. 

Fixes #741

### DIFF
--- a/js/researches/calculateFleschReading.js
+++ b/js/researches/calculateFleschReading.js
@@ -20,6 +20,8 @@ module.exports = function( paper ) {
 		return 0;
 	}
 
+	text = stripNumbers( text );
+
 	var numberOfSentences = countSentences( text );
 
 	var numberOfWords = countWords( text );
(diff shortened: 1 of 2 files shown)


atimmer:
This deserves a few specs

atimmer:
CR Done 👍🏻

atimmer:
Acceptance done 👍🏻

PR #783, opened 2016-07-11
Replaces hidden element for titlewidth in snippet preview.

### LINKED ISSUE #776: Fix measurement of snippet title width in snippet preview
There is something wrong with the measurements in the snippet preview for the title. When I fill the title field with a lot of Ws the title is filled sooner than the title progress bar. Preferably we don't measure the width of a hidden/invisible element, but we measure the width of the element that is actually displayed to the user.

See the following screenshot:
screen shot 2016-07-11 at 10 17 04

terw-dan:
This replaces the hidden element used for measuring the title width. This makes sure the CSS isn't overwritten and differs from the visible element. 
Added a span in the snippet template that sizes with the content. The assessment and progressbars now use the visible snippet title, making it consistent. 

Added some small tweaks to the title width assessment, where a title of exactly 600px wide would not be seen as too long or as a nice size. 

For testing, make sure the title width works as expected in the snippet preview, as soon as the snippet preview starts abbreviating the title, it should report it is too long and the progressbars should turn orange. 
`wwwWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWW` is exactly 600px wide (on my machine) to test the treshold. 

Fixes #776

### DIFF
--- a/css/_snippet-editor.scss
+++ b/css/_snippet-editor.scss
@@ -284,6 +284,3 @@ $snippet_width: 600px;
 @include progress-bar-color(".snippet-editor__progress--ok", $color_ok);
 @include progress-bar-color(".snippet-editor__progress--good", $color_good);
 
-.hidden-title {
-	font-size: 18px;
-}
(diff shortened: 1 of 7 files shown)


atimmer:
CR & Acceptance done 👍🏻

PR #786, opened 2016-07-11
Fixes sentence detection by removing nbsp's

terw-dan:
This strips out non breaking spaces for the sentence checks, to make sure we don't end up with a sentence consisting of only a space.

Fixes #https://github.com/Yoast/wordpress-seo/issues/4805


atimmer:
CR & Acceptance done 👍🏻

PR #787, opened 2016-07-12
Fixes error in textlength assessments

terw-dan:
When a text had exactly 300 words it would break, since it wouldn't return a valid assessment result. 
Also fixes the taxonomy length, that had the same issue with 150 words. 

Fixes https://github.com/Yoast/wordpress-seo/issues/3226


atimmer:
CR & Acceptance Done 👍🏻

PR #792, opened 2016-07-12
Don't assume there are multiple words for sentence beginnings. 

### LINKED ISSUE #790: sentenceBeginnings assessment throws an error
The sentenceBeginnings assessment to not throw an error.

### What happened instead?

> An error occurred in the 'sentenceBeginnings' assessment

This probably has something to do with spaces/newlines.

Input:
```
<span class="caption">A
  fan poses during China's premiere of the film "Warcraft" at a
  theatre in Shanghai, China June 7, 2016.</span>
```

### Technical info
* Platform: Standalone
* Platform version: 1.4 develop
* Yoast SEO version: 1.4 develop

terw-dan:
In the sentence beginnings we append the second word in a sentence if the first word is an exclusion word. 
But if the sentence only consists of 1 exclusionword, and no more, it would break. 

This fix makes sure we don't assume there is a second word. If there is no second word, we use only the first word as a sentence beginning. 

Fixes #790

### DIFF
--- a/js/researches/getSentenceBeginnings.js
+++ b/js/researches/getSentenceBeginnings.js
@@ -58,12 +58,13 @@ module.exports = function( paper ) {
 	var sentences = getSentences( paper.getText() );
 	var firstWordExceptions = getFirstWordExceptions( paper.getLocale() )();
 	var sentenceBeginnings = sentences.map( function( sentence ) {
+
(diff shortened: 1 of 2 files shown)


atimmer:
CR Done 👍🏻

atimmer:
CR & Acceptance done 👍🏻

PR #794, opened 2016-07-13
Adds changelog for 1.4

terw-dan:
Changelog for 1.4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,16 +9,16 @@ We will follow [Semantic Versioning](http://semver.org/) from version 2 and onwa
 
 ### Changed
     * Improves feedback texts of assessments.
-    * Improves Russian transliteraion.
-    * Determine length of title based on pixels in stead of on number of characters.
+    * Improves Russian transliteration.
+    * Determine length of title based on the width in pixels instead of on number of characters.
     * Words comprised of only digits are no longer counted in the Flesch Reading assessment.
-    * Improves passive voice detection by ommiting HTML tags.
-    * Non breaking spaces are replaced with normal spaces
-    * Improves the competing links check in post for linking to the same post.
+    * Improves passive voice detection by omitting HTML tags.
+    * Non breaking spaces are replaced with normal spaces in word boundaries.
 ### Fixed
-    * Improved keyword density assessment by rounding the scores.
-    * ¿ and ¡ are now used as valid sentence beginnings.
+    * Improved keyword density assessment by scoring on the rounded result.
+    * ¿ and ¡ are now accepted as sentence beginnings.
     * Fix a bug where the text assessment would fail with exactly 300 words.
+    * Fix a bug in the competing links assessment where a link to the same post would be counted as an outbound link.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,36 @@ This changelog is according to [Keep a Changelog](http://keepachangelog.com).
 All notable changes to this project will be documented in this file.
 We will follow [Semantic Versioning](http://semver.org/) from version 2 and onwards.
 
+## 1.4: July 19th, 2016
+
+### Changed
+    * Improves feedback texts of assessments.
+    * Improves Russian transliteraion.
```
atimmer: There is a typo here.

CHANGELOG.md
```diff
@@ -5,6 +5,36 @@ This changelog is according to [Keep a Changelog](http://keepachangelog.com).
...
+    * Determine length of title based on pixels in stead of on number of characters.
```
atimmer: `instead` is one word.

atimmer: Maybe try: `based on the width in pixels instead of on the number of characters.`

CHANGELOG.md
```diff
@@ -5,6 +5,36 @@ This changelog is according to [Keep a Changelog](http://keepachangelog.com).
...
+    * Words comprised of only digits are no longer counted in the Flesch Reading assessment.
+    * Improves passive voice detection by ommiting HTML tags.
```
atimmer: `omitting` is written with two ts and one m.

CHANGELOG.md
```diff
@@ -5,6 +5,36 @@ This changelog is according to [Keep a Changelog](http://keepachangelog.com).
...
+    * Non breaking spaces are replaced with normal spaces
```
atimmer: I miss when this happens, before each assessment, before the whole analysis?

CHANGELOG.md
```diff
@@ -5,6 +5,36 @@ This changelog is according to [Keep a Changelog](http://keepachangelog.com).
...
+    * Improves the competing links check in post for linking to the same post.
```
atimmer: I think this should be in the `Fixed` category and should say something like: `Fix a bug in the competing links assessment where a link to the same post would be counted as an outbound link.`

CHANGELOG.md
```diff
@@ -5,6 +5,36 @@ This changelog is according to [Keep a Changelog](http://keepachangelog.com).
...
+### Fixed
+    * Improved keyword density assessment by rounding the scores.
```
atimmer: I miss when the rounding happens, we were already rounding the scores but now we do it earlier to make the textual result make sense.

CHANGELOG.md
```diff
@@ -5,6 +5,36 @@ This changelog is according to [Keep a Changelog](http://keepachangelog.com).
...
+    * ¿ and ¡ are now used as valid sentence beginnings.
```
atimmer: I think `accepted` is better instead of `used`.

PR #800, opened 2016-07-15
Prevents setting the width to 0 when not empty

terw-dan:
If you switch tabs, the title width would be reset to 0. 
The only time when the title width should be 0 is if there is an empty text string. If the text is not empty, we should not updat the data.titleWidth. 

fixes #https://github.com/Yoast/wordpress-seo/issues/5168

For testing:
In the standalone, test that the title width assessment is working like expected
In WP, when you switch tabs (from seo to readability and back), it should not set the value to zero.


boblinthorst:
CR: ok 👍

PR #809, opened 2016-07-18
Fix markers on passive voice with HTMLtags

### LINKED ISSUE #803: Passive voice markings made impossible by <i> tags
The passive voice marker to mark passive sentences that contain `<i>` tags.

### What happened instead?
It did not mark sentences containing `<i>` tags. The marks did not show up in the HTML. 

N.B. :

- The markings did work when I removed the `<i>` tags.

- This was only the case in WP. Everything seems to work fine in the standalone.

- The marks for sentence length, transition words, and sentence beginnings do work in combination with `<i>` tags. 

Mark passive voice in this text:

```
The first mention of Nijmegen in history is in the 1st century BC, when the Romans built a military camp on the place where Nijmegen was to appear; the location had great strategic value because of the surrounding hills, which gave a good view over the Waal and Rhine valley.

By 69, when the Batavians, the original inhabitants of the Rhine and Maas delta, revolted, a village called <i>Oppidum Batavorum</i> had formed near the Roman camp. This village was destroyed in the revolt, but when it had ended the Romans built another, bigger camp where the Legio X <i>Gemina</i> was stationed. Soon after, another village formed around this camp.

In 98, Nijmegen was the first of two settlements in what is now the Kingdom of the Netherlands to receive Roman city rights.

In 103, the X <i>Gemina</i> was re-stationed to Vindobona, modern day Vienna, which may have been a major blow to the economy of the village around the camp, losing around 5000 inhabitants. In 104 Emperor Trajan renamed the town, which now became known as <i>Ulpia Noviomagus Batavorum</i>, Noviomagus for short (the origin of the current name Nijmegen).
```

### Technical info
* Platform: WordPress 
* Platform version: 4.5.3
* Yoast SEO version: release 3.4

terw-dan:
Fixes #803

For correct matching we stripped the tags from the sentences, but did not store the original sentence to push in the array that would be used for markers. 
This creates a new variable for testing, but pushes the unaltered sentence so it can be marked.

For testing, create a passive voice sentence with a HTMLtag, this should be marked.

replaces #808

### DIFF
--- a/js/researches/getPassiveVoice.js
+++ b/js/researches/getPassiveVoice.js
@@ -314,9 +314,9 @@ module.exports = function( paper ) {
 
 	// Get subsentences for each sentence.
 	forEach( sentences, function( sentence ) {
-		sentence = stripHTMLTags( sentence );
+		var strippedSentence = stripHTMLTags( sentence );
 
-		var subSentences = getSubsentences( sentence );
+		var subSentences = getSubsentences( strippedSentence );
 
 		var passive = false;
 		forEach( subSentences, function( subSentence ) {


IreneStr:
CR & acceptance done :+1:

PR #810, opened 2016-07-18
Removed the assumption that the loading dialog is always present when trying to remove it

jcomack:
Required to fix https://github.com/Yoast/wordpress-seo/issues/5194


boblinthorst:
CR: ok 👍

boblinthorst:
Acceptance: ok 👍

PR #814, opened 2016-07-19
Added a missing if around the seoAssessorPresenter

jcomack:
If the SEO analysis is disabled, the console will throw an error when attempting to disable the markers.


terw-dan:
CR done :+1:

PR #815, opened 2016-07-19
Checks the correct element for defining width of pagetitle

terw-dan:
We should check the rendered element for the page title, not the input title because that can stay empty (if the snippet title is generated from the page title, and not filled in in the snippet itself). 

Fixes the first issue of https://github.com/Yoast/wordpress-seo/issues/5198


terw-dan:
CR by Omar

jcomack:
Acceptance done 👍

PR #817, opened 2016-07-19
Refreshes the snippetPreview before measuring

terw-dan:
This make sure the snippet is refreshed before the title is measure, because otherwise it could fail due to rendering of replacable variables. 

This we we get the correct width to use for the assessments.

Fixes #https://github.com/Yoast/wordpress-seo/issues/5198


jcomack:
CR & acceptance done 👍

PR #822, opened 2016-07-19
Fixed a possible division by zero issue

### LINKED ISSUE #819: An [enter] as content results in NaN percentage of transition words.
0% of the sentences contain a transition word, which is bad.

### What happened instead?
NaN% of the sentences contain a transition word, which is great.

1. create a new post. 
2. put the yoast metabox tab on readability.
3. Have the content of the post be a single [enter]

### Technical info
* WordPress version: 4.5.3
* Yoast SEO version: 3.4.

yoast_seo_content_analysis

jcomack:
This PR fixes an issue where the user only inserts an enter into the content, leading into transition words returning NaN due to a division by zero error.

Fixes #819


terw-dan:
CR Done :+1:

terw-dan:
Acceptance :+1:

PR #829, opened 2016-07-21
Improves keyword density check by not relying on wordboundary

### LINKED ISSUE #203: The wordboundary regex is broken with consecutive keywords
The wordboundary regex we use isn't working when keywords are used consecutively in a sentence. 
This issue was also present in 2.3.5

The phrase "a keyword and a keyword" will match keyword twice (as it should)
When using "a keyword keyword keyword" will match the keyword only twice, in stead of 3 times. 

This is due to the fact the regex matches the whitespaces around the keyword as well. This screenshot taken from regexr shows it how it is matched. 
screen shot 2015-12-15 at 15 13 44

Only the first and third occurence of the keyword is matched, not the second one.

terw-dan:
Fixes #203

Simply changing the wordboundary regex would give unpredictable results with empty keywords and spaces. 
This PR implements a check where we match per word, like we already did for the passive voice analysis. This makes sure a keyword is matched if it is followed directly by another keyword. 

For testing: Use a sentence with the keyword 2 or more times after eachother. Make sure you use 100+ words to trigger the keyword density check.

### DIFF
--- a/js/researches/getKeywordDensity.js
+++ b/js/researches/getKeywordDensity.js
@@ -1,11 +1,7 @@
 /** @module analyses/getKeywordDensity */
 
-var filter = require( "lodash/filter" );
-
 var countWords = require( "../stringProcessing/countWords.js" );
-var transliterate = require ( "../stringProcessing/transliterate.js" );
-
-var getWords = require( "../stringProcessing/getWords.js" );
+var countWordOccurrences = require( "../stringProcessing/countWordOccurrences.js" );
 
 /**
  * Calculates the keyword density .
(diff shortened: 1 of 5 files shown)


js/researches/getKeywordDensity.js
```diff
@@ -17,6 +21,10 @@ module.exports = function( paper ) {
...
 		return 0;
 	}
-	var keywordCount = matchWords( text, keyword, locale );
-	return ( keywordCount / wordCount ) * 100;
+	var words = getWords( text );
+	var keywordCount = filter( words, function( word ) {
+		return ( word === keyword || transliterate( word, locale ) === keyword );
+	} );
+
+	return ( keywordCount.length / wordCount ) * 100;
```
jcomack: Shouldn't this result be properly formatted before returning it?

jcomack:
CR done 🚧

boblinthorst:
Acceptance: It is indeed found three times in code, but the assessment still says 2, the assessment is incorrect.

jcomack:
CR 2 done 👍

jcomack:
Acceptance done 👍

PR #839, opened 2016-07-27
Improves language check in assessments

### LINKED ISSUE #838: Create a way for checking availability in languages for assessments
We need a function to check if an assessment is available for the current language. 
Currently, we split the locale on the `_` and return true if it matches the language string, which can become quite large with multiple languages. 

Create a function that matches an array of available languages so we can use a generic way for all assessments.

terw-dan:
Fixes #838 
This creates a uniform way of getting the availability of languages used in assessments. 

Replaces the current checks in the isApplicable of the assessment where we would match a bunch of languages. 

For testing: Change the locale in the stand alone and see if all the assessments are available as expected.

### DIFF
--- a/js/assessments/fleschReadingEaseAssessment.js
+++ b/js/assessments/fleschReadingEaseAssessment.js
@@ -1,6 +1,10 @@
 var AssessmentResult = require( "../values/AssessmentResult.js" );
 var inRange = require( "lodash/inRange" );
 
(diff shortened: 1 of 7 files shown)


boblinthorst:
CR: ok 👍

jcomack:
Acceptance done 👍

PR #841, opened 2016-07-28
Add exclusionwords for english syllable count

terw-dan:
Adds 'colonel' and 'wednesday' to the exclusionwords, since they both cannot be counted with regular syllable counters.


jcomack:
CR done 👍

jcomack:
Acceptance done 👍

PR #843, opened 2016-07-29
Improves removing punctuation.

### LINKED ISSUE #835: Remove « and » from text before counting syllables.
The left-pointing double angle quotation mark («) and right-pointing double angle quotation mark (»), which are used in for example French, should be removed before words and syllables are counted for the Flesh Reading Ease. Adding « and » to the sentenceTerminators list should fix this.

terw-dan:
The remove terminators and remove punctuation are the same functions, but using a slightly different list of punctuation marks to remove. 

This uses the remove punctuation in favor of the remove terminators (since it was only used in 1 place). 
Merged the 2 specs together. 

Added the «, » and - to the punctuations to remove. 

Fixes #835 

After https://github.com/Yoast/YoastSEO.js/pull/872 is merged we should meet the treshold for the coverage in this PR.


js/stringProcessing/removePunctuation.js
```diff
@@ -1,18 +1,18 @@
 // Replace all other punctuation chars at the beginning or at the end of a word.
-var punctuationRegexString = "[\-()_\\[\\]’“”\"'\/]";
+var punctuationRegexString = "[\\–\\-\\(\\)_\\[\\]’“”\"'.?!:;,¿¡«»]+";
 var punctuationRegexStart = new RegExp( "^" + punctuationRegexString );
 var punctuationRegexEnd = new RegExp( punctuationRegexString + "$" );
 
 /**
- * Replaces punctuation chars from the given text string.
+ * Replaces punctuation chars from the given word string.
```
jcomack: Please don't use abbreviations.

spec/stringProcessing/removePunctuationSpec.js
```diff
@@ -23,3 +23,17 @@ describe("a test removing punctuation from a string", function(){
 	});
 
 });
+
+describe( "Removing terminators at the begin and end of a word", function(){
```
jcomack: Shouldn't `terminator(s)` be `punctuation`? This also applies to line 28.

jcomack:
CR done 👍

jcomack:
CR 2 done 👍

js/stringProcessing/removePunctuation.js
```diff
@@ -1,18 +1,18 @@
...
+// Replace all other punctuation characters at the beginning or at the end of a word.
+var punctuationRegexString = "[\\–\\-\\(\\)_\\[\\]’“”\"'.?!:;,¿¡«»]+";
 var punctuationRegexStart = new RegExp( "^" + punctuationRegexString );
 var punctuationRegexEnd = new RegExp( punctuationRegexString + "$" );
 
 /**
- * Replaces punctuation chars from the given text string.
+ * Replaces punctuation characters from the given word string.
  *
- * @param {String} text The text to remove the punctuation chars for.
```
atimmer: Does this function require a word to be input to work? I think `text` is correct in the documentation.

jcomack:
Acceptance failed. 

Oddly enough these characters are still being processed and not ignored. On top of that, the flesch reading score is now 0.5 lower than it was before.

terw-dan:
@jcomack The characters are removed from the text, if this is the only character in a word the word count will be altered, and so will the Flesch Reading score, this is to be expected.

jcomack:
Acceptance 2 done. Looks good now :)

PR #881, opened 2016-09-02
Escape the keyword in stead of sanitizing. 

terw-dan:
## Summary

Escape the keyword rather than sanitizing it in the paper so the keyword can be matched correctly. 

* In stead of stripping of certain characters that could interfere with regexes, they are now escaped. 

## Test instructions

* Try matching `$keyword`, this should now work. 

Fixes https://github.com/Yoast/wordpress-seo/issues/5244


js/values/Paper.js
```diff
@@ -21,7 +21,7 @@ var defaultAttributes = {
  * @returns {object} The attributes passed to the Paper.
  */
 var sanitizeAttributes = function( attributes ) {
-	attributes.keyword = sanitizeString( attributes.keyword );
+	attributes.keyword = escapeRegExp( attributes.keyword );
```
boblinthorst: Perhaps change method name if you change the action in the function is changed to escape instead of sanitize

atimmer:
We shouldn't do the escaping when passing the value to the Paper. We should escape as late as possible before using the keyword in a regex. Because that will be less trivial, I am moving this to the 1.6 milestone.

terw-dan:
Re-did this PR. In stead of escaping the keyword in the paper, we do it in the researches where necessary.

boblinthorst:
CR: ok 👍

jcomack:
Acceptance done 👍

PR #889, opened 2016-09-05
Swap the snippet editor icons for RTL

### LINKED ISSUE #886: The snippet editor "caret" icon needs to be swapped for RTL
Needed for automated RTL CSS generation, see Yoast/wordpress-seo#2693

The arrows that appear on hover and focus in the snippet editor need to be swapped for RTL languages. It's about the sass mixin `svg-caret-before` that uses a sass function `svg-icon-caret-right` to generate the svg icon. As far as I see there's already a swapped icon ready to be used, see: `svg-icon-caret-left`.

screen shot 2016-09-05 at 12 18 20

afercia:
Fixes #886 

I think we just need to change the svg background image, the other properties will be handled by the grunt task.

Needed for automated RTL CSS generation, see Yoast/wordpress-seo#2693

### DIFF
--- a/css/_snippet-editor.scss
+++ b/css/_snippet-editor.scss
@@ -18,6 +18,10 @@ $snippet_width: 600px;
 	background-size: 25px;
 }
 
+@mixin svg-caret-before-rtl($color) {
+	background-image: url(svg-icon-caret-left($color));
+}
+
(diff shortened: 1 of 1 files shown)


atimmer:
CR & Acceptance Done 👍🏻

PR #899, opened 2016-09-15
Replace any target=new with target=_blank

### LINKED ISSUE #877: Review usage of target='new' vs target='_blank'
Some links use `target='new'`. Unless there's a specific reason I'm not aware of, they should probably use `target='_blank'` instead. What's the difference?

References for HTML 4 and 5:
https://www.w3.org/TR/html401/types.html#type-frame-target
https://www.w3.org/TR/html5/browsers.html#browsing-context-names

- `target="new"` is valid but doesn't have any special meaning, it could be `target="imabittiredtoday"`, it just gives a name to the new tab/window so it can be referenced by other links that use the same `target="imabittiredtoday"` and they will open in that same tab/window (if it's still open)
- `target="_blank"` is valid and always opens a new tab/window
- `target="_new"` is not valid

What happens in practice and how to reproduce?

- edit a post and enter a focus keyword with a stop word, you should get something like in the screenshot below with two links that use `target="new"`

screen shot 2016-08-31 at 10 25 12

- click on the link "this article"
- the link opens in a new tab
- don't close the new tab, just switch back to the previous one
- now click on the link "stop words"
- the link opens in the same tab that was used before for the "this article" link

To clarify: technically this is not wrong :) but maybe can be confusing for users. Also for consistency, I'd say either we always use `target="something"` or `target="_blank"`.

afercia:
Fixes #877


terw-dan:
CR :+1:

terw-dan:
Acceptance :+1:

PR #915, opened 2016-10-03
Make class for yoast-section.

andizer:
Change the id based styling into class based.

### DIFF
--- a/css/_analysis.scss
+++ b/css/_analysis.scss
@@ -143,7 +143,7 @@ li.score:after {
 
 	&__mark-container {
 		display: inline-block;
-		float: left;
+		float: right;
 		width: 33px;
 		min-height: 1px;
 	}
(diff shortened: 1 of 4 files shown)


terw-dan:
CR done 👍

terw-dan:
Acceptance, when using this PR the snippet preview the snippet preview has no longer a maximum width of ~600 px. 

image

We should keep the snippet preview the same width as it had before.

andizer:
@rensw90 Why did you set this pull to scheduled without giving feedback?

rensw90:
Because of the note from @terw-dan about this PR that has to be addressed.

andizer:
@rensw90 If you take  As far as I know, the styling issue should be gone by doing `grunt build:jst` and that file is committed already by @terw-dan

terw-dan:
@andizer @rensw90 
The last thing that went wrong was the missing border-box in the CSS. I've added that, so please review and test.

rensw90:
Acceptance done 👍

PR #918, opened 2016-10-11
Generic screen reader text for the Readability tab

### LINKED ISSUE #917: Better screen reader text for the Readability tab
See https://github.com/Yoast/wordpress-seo/issues/4991

The Readability tab in the meta box needs a better hidden text for the readability score. Currently, it still uses the screen reader text for the SEO score so the actual text announced by screen readers is a bit confusing:
```
Readability Bad SEO score.
Readability OK SEO score.
Readability Good SEO score.
```

The generic (content) tab should use a generic text for the score text, I'd propose to add a new item in the "indicator" object to provide new generic strings:
```
Needs improvement
OK
Good
```

afercia:
Fixes #917


rensw90:
CR-Done 👍

rensw90:
Acceptance done 👍

PR #929, opened 2016-11-04
Move the overallScore CSS out from the main stylesheet.

### LINKED ISSUE #928: Move the overallScore CSS rules to the example pages
See https://github.com/Yoast/wordpress-seo/issues/6026

As far as I see the CSS rules and colors related to `overallScore` are only used on the example pages. Ideally, we shouldn't import them in the `yoast-seo` stylesheet to avoid they're included in the plugins where the content analysis is used.

This is part of the ongoing effort to standardise colours and improve colour contrast. See also https://github.com/Yoast/yoast-components/issues/106

afercia:
Fixes #928


atimmer:
Will the CSS be overwritten when running `grunt build:sass` now?

afercia:
@atimmer not sure I fully understand your question, as far as I see the stylesheet for the browserified example has a task that needs to be run manually `grunt sass:example` and has no alias, nor is part of the build process.
The standalone example instead, has no tasks at all.

atimmer:
Sorry, I misread `browserified.scss` as `browserified.css`.

CR Done 👍🏻

terw-dan:
Acceptance 👍

PR #932, opened 2016-11-08
Moved the extending of Assessor in SEOAssessor to support code completion

jcomack:
## Summary
When using YoastSEO.js as a third party library, code completion fails in the SEO Assessor due to a "wrong" extend / export order. This PR solves that.


terw-dan:
CR 👍

PR #935, opened 2016-11-09
Enhances feedback strings for metadescription length assessment

IreneStr:
## Summary

- Made the verb tense more consistent.
- Replaced commas by periods between sentences. 
- Made the feedback string for `score: 9` more descriptive. Before, there was no link between the feedback and the assessment: the length of the description was not mentioned at all. 

This PR can be summarized in the following changelog entry: Enhanced feedback strings for the metadescription length assessment.


terw-dan:
CR done

IreneStr:
Do we want to keep the questions in the feedback string? As far as I know, it’s the only feedback string with this kind of questions. The questions are also rather broad and not really actionable.

PR #941, opened 2016-11-14
Improve keyword highlighting in snippet.

terw-dan:
## Summary

* Instead of stripping out special characters in the keyword, use lodash escape to escape them. 
* Removed the sanitize string from the snippet preview and let stringToRegex handle this (where this was already being done). 
* Updated some specs 

## Test instructions

* Use `Yoast SEO 3.8` as keyword and make a meta with the text `Yoast SEO 3.8` in it. See that it isn't highlighted. 
* Change `Yoast SEO 3.8` to `Yoast SEO` and see the keyword is highlighted. 
* Switch to this branch and see `Yoast SEO 3.8` is highlighted as expected. 

* Keywords with characters like `&` or `*` can now be highlighted.

Fixes the first part of https://github.com/Yoast/wordpress-seo/issues/6042


js/snippetPreview.js
```diff
@@ -733,10 +732,10 @@ SnippetPreview.prototype.getPeriodMatches = function() {
  */
 SnippetPreview.prototype.formatKeyword = function( textString ) {
 	// Removes characters from the keyword that could break the regex, or give unwanted results.
-	var keyword = this.refObj.rawData.keyword.replace( /[\[\]\{\}\(\)\*\+\?\.\^\$\|]/g, " " );
+	var keyword = this.refObj.rawData.keyword;
 
 	// Match keyword case-insensitively.
-	var keywordRegex = stringToRegex( keyword, "", false );
```
atimmer: This line is not necessary, given that we still set `keyword` above this line.

spec/snippetPreviewSpec.js
```diff
@@ -95,6 +95,57 @@ describe( "The SnippetPreview format functions", function(){
...
+			}
+		};
+
+		var snippetPreview = new SnippetPreview({
+			analyzerApp: mockApp,
+			targetElement: mockElement
+		});
+
+		it( "should highlight the keywords", function(){
+			expect( snippetPreview.formatMeta() ).toBe( "This is the<strong> Yoast SEO 3.9 </strong>release")
```
atimmer: I don't think the space should be included within the `strong` tag here.

spec/snippetPreviewSpec.js
```diff
@@ -95,6 +95,57 @@ describe( "The SnippetPreview format functions", function(){
...
+			}
+		};
+
+		var snippetPreview = new SnippetPreview({
+			analyzerApp: mockApp,
+			targetElement: mockElement
+		});
+
+		it( "should highlight the keywords", function(){
+			expect( snippetPreview.formatMeta() ).toBe( "If you like Yoast SEO, please give a<strong> 5* rating</strong>" )
```
atimmer: I don't think the space should be included in the `strong` tag here.

spec/stringProcessing/stringToRegexSpec.js
```diff
@@ -0,0 +1,14 @@
+var stringToRegex = require( "../../js/stringProcessing/stringToRegex.js" );
```
atimmer: Let's call this `createWordRegex`.

spec/stringProcessing/sanitizeStringSpec.js
```diff
@@ -2,7 +2,8 @@ var sanitizeString = require("../../js/stringProcessing/sanitizeString.js");
 
 describe("Test for removing unwanted characters", function(){
 	it("returns cleaned string", function(){
-		expect( sanitizeString( "keyword*" ) ).toBe( "keyword" );
+		// Because regexes are now properly escaped, there is no need to strip characters like * from the keyword.
+		expect( sanitizeString( "keyword*" ) ).toBe( "keyword*" );
```
atimmer: I would put all regex characters here, for good measure.

spec/snippetPreviewSpec.js
terw-dan: I totally agree. I've created a new issue for this https://github.com/Yoast/YoastSEO.js/issues/944
The problem isn't in the highlighting, but in the way we capture keywords, so that should be fixed separately to prevent this issue getting huge.

andizer:
CR Done 👍

PR #954, opened 2016-11-24
Add mobile preview

andizer:
## Summary

* Added a module for handling the toggles

## Test instructions

* Open the browserified example page and see if you can switch preview
* Default should be the mobile preview
* When on the desktop preview when the viewport is made smaller then 640px, it should be automatically switched to the mobile preview (this depends on the size of the viewport, so won't work in WordPress at the moment). 

requires: https://github.com/Yoast/wordpress-seo/pull/6423 when testing in WordPress. This makes sure the mobile preview is responsive in WordPress.


afercia:
Before it goes formally under Code Review, I'd recommend some accessibility improvements if possible. Or please just let me know if splitting a11y improvements in a separate issue is preferable.

terw-dan:
@afercia please feel free to add comments for a11y as a code review. If they are very big changes we can always decide to move those to another PR in stead of this one.

templates/snippetEditor.jst
```diff
@@ -3,36 +3,43 @@
...
-                    <%- metaDescriptionDate %> -
-                </span>
-            <% } %>
-            <span class="desc" id="snippet_meta">
-                <%- rendered.meta %>
-            </span>
-        </div>
+		<div class="snippet-editor__view-toggle">
+			<button class="snippet-editor__view-icon snippet-editor__view-icon-desktop" data-type="desktop" />
+			<button class="snippet-editor__view-icon snippet-editor__view-icon-mobile" data-type="mobile" />
```
afercia: The buttons are empty, there's nothing screen readers can announce.

screen shot 2016-11-24 at 15 28 49

I'd suggest to add some screen-reader-text or `aria-label` ("Desktop preview" - "Mobile preview" or something like that).

Also, there's no visual or audible feedback about which is the currently selected, active, view. For the visual feedback I guess we have to wait for a design proposal as mentioned in https://github.com/Yoast/yoast-components/issues/110
For the audible feedback, I'd propose to use the `aria-pressed` attribute, as done for example in the WordPress Customizer for the device preview buttons. This attribute uses false/true values and conveys the information the buttons are "toggle buttons", mutually exclusive. Meaning only one of them can have a "pressed" state. Screen readers will announce the buttons as "toggle button" and will announce also their state (some screen readers use the word "selected", others the word "pressed"):

screen shot 2016-11-24 at 15 29 42

screen shot 2016-11-24 at 15 29 54

js/snippetPreviewToggler.js
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * Sets the events for the preview toggles to switch between the preview modes.
+ *
+ * @returns {void}
+ */
+SnippetPreviewToggler.prototype.setEvents = function() {
+	forEach( this.previewToggles, function( previewToggle ) {
+		previewToggle.addEventListener( "click", function( evt ) {
+			evt.preventDefault();
+
```
terw-dan: Is this whitespace necessary?

js/snippetPreviewToggler.js
```diff
@@ -0,0 +1,117 @@
...
+};
+
+/**
+ * Constructs the snippet preview toggle.
+ *
+ * @param {string}    previewMode    The default preview mode.
+ * @param {Element[]} previewToggles Array with toggle elements.
+ *
+ * @property {string}   previewMode    The current preview mode.
+ * @propert {Element[]} preivewToggles The array with toggle elements.
```
terw-dan: typo.

js/snippetPreviewToggler.js
```diff
@@ -0,0 +1,117 @@
+
+var forEach = require( "lodash/forEach" );
+var domManipulation = require( "./helpers/domManipulation.js" );
+
+var previewModes = {
+	desktop: "",
```
terw-dan: Maybe we could use `snippet-editor__view--desktop` for this, so it will be more explicit what previewmode is set.

terw-dan:
CR 1 done.

js/snippetPreviewToggler.js
andizer: Not necessary, but it makes it more clear for me.

terw-dan:
Since this will be implemented via react I'll close this PR for now.

atimmer:
We have decided that because of time constraints we are going to implement this solution. So I am reopening this pull request. It does require some more work, which I will detail now:

* The preview should automatically switch to the mobile version. This should occur if the container is too small for the desktop variant.
* The button for the desktop variant should be hidden if there is no room for the desktop variant.
* The mobile variant should automatically resize it self it it's container is resized.

afercia:
Played a bit with this, I'd recommend to use an `aria-pressed` attribute on the buttons, its value should be updated to true/false depending on the button's state. To illustrate what `aria-pressed` does, here's some screenshots. Currently, there's no indication about which of the buttons is the "active" one:

screen shot 2017-01-03 at 15 19 34
screen shot 2017-01-03 at 15 19 41

`aria-pressed` makes assistive technologies announce the buttons as "toggle button" instead of just "button". See below when `aria-pressed` is `false`:

screen shot 2017-01-03 at 15 44 22

and when `aria-pressed` is `true`, it will also make the button be announced as "selected":

screen shot 2017-01-03 at 15 44 31

This change would be very simple and just needs to set the attribute value in `_removeActiveState()` and `_setActiveState()`.

Also, the buttons should have a `type="button"` attribute, this should be set in the JS template and would allow to remove calling `preventDefault()` (and passing the event), since when buttons have a `type="button"` attribute there's no default action to prevent.

afercia:
About:
> The button for the desktop variant should be hidden if there is no room for the desktop variant.

Maybe both button should be hidden? Because when there's no room for the desktop view, then having a button for the mobile view would be a bit pointless (and also defeat the purpose of the toggle buttons with `aria-pressed`).

Thinking at a possible solution, given the I think there's no reliable way to do this with CSS and media queries and it would probably require to check the preview container width with JS on DOM ready, on load, and on any window resize... Since the `resize` event is inconsistent across browsers and fires multiple times on each resize and orientation change, maybe we should implement something similar to WordPress custom event `wp-window-resized`, see in `wp-admin/js/common.js`.

js/snippetPreview.js
```diff
@@ -919,6 +941,15 @@ SnippetPreview.prototype.updateProgressBars = function() {
 };
 
 /**
+ * Gets the width of the Snippet Preview to set the Snippet Preview Toggler visibility.
+ * @returns {void}
+ */
+SnippetPreview.prototype.handleWindowSize = debounce( function() {
```
jcomack: The name of the function seems to (mostly) be called when resizing the window. Consider altering the name to `handleWindowResizing` or similar to better reflect when this is called.

js/snippetPreviewToggler.js
```diff
@@ -0,0 +1,153 @@
...
+};
+
+/**
+ * Sets the events for the preview toggles to switch between the preview modes.
+ *
+ * @returns {void}
+ */
+SnippetPreviewToggler.prototype.setEvents = function() {
+	forEach( this.previewToggles, function( previewToggle ) {
+		previewToggle.addEventListener( "click", function( evt ) {
```
jcomack: Consider moving the anonymous callback into its own function to make this a bit more readable.

js/snippetPreviewToggler.js
```diff
@@ -0,0 +1,153 @@
...
+SnippetPreviewToggler.prototype.initialize = function() {
+	this._setPreviewMode( this.previewMode, this._findElementByMode( this.previewMode ) );
+};
+
+/**
+ * Sets the events for the preview toggles to switch between the preview modes.
+ *
+ * @returns {void}
+ */
+SnippetPreviewToggler.prototype.setEvents = function() {
```
jcomack: `bindEvents` sound a bit more fitting.

js/snippetPreviewToggler.js
```diff
@@ -0,0 +1,153 @@
...
+		previewToggle.addEventListener( "click", function( evt ) {
+			evt.preventDefault();
+
+			this._setPreviewMode( previewToggle.getAttribute( "data-type" ), previewToggle );
+		}.bind( this ) );
+	}.bind( this ) );
+};
+
+/**
+ * Returns the element by given node.
```
jcomack: I believe `node` should be `mode`.

js/snippetPreviewToggler.js
```diff
@@ -0,0 +1,153 @@
...
+ *
+ * @param {string} previewMode The element to search.
```
jcomack: I believe the description is incorrect.

js/snippetPreviewToggler.js
```diff
@@ -0,0 +1,153 @@
...
+ * @param {Element} toggleElement The element that has been triggered.
+ *
+ * @returns {void}
+ * @private
+ */
+SnippetPreviewToggler.prototype._setPreviewMode = function( previewMode, toggleElement ) {
+	this._removeActiveStates();
+	this._setActiveState( toggleElement );
+
+	if( previewMode === this.previewMode ) {
```
jcomack: Shouldn't this check be done prior to removing states?

js/snippetPreviewToggler.js
```diff
@@ -0,0 +1,153 @@
...
+};
+
+/**
+ * Sets the visibility based on the width of the Snippet Preview container.
+ *
+ * @param {number} previewWidth the width of the Snippet Preview container.
+ *
+ * @returns {void}
+ */
+SnippetPreviewToggler.prototype.setToggleVisibility = function( previewWidth ) {
```
jcomack: `setToggleVisibility` seems like a strange name. Aren't you just toggling the visibility? I'd consider changing this to either `toggleVisibility` or `setVisibility`. Please don't forget to update the documentation if you decide to change this name.

js/snippetPreviewToggler.js
terw-dan: Nope, we always want to set the active state of the buttons, otherwise, you can have buttons without an active state.

js/snippetPreview.js
```diff
@@ -15,6 +15,7 @@ var transliterate = require( "./stringProcessing/transliterate.js" );
 var templates = require( "./templates.js" );
 var snippetEditorTemplate = templates.snippetEditor;
 var hiddenElement = templates.hiddenSpan;
+var SnippetPreviewToggler = require( "./snippetPreviewToggler" );
```
terw-dan: This should be capitilized, just like other objects.

jcomack:
CR 2 done 👍

jcomack:
Acceptance done 👍

PR #957, opened 2016-11-29
Improves matching in first paragraph

terw-dan:
## Summary

Improves matching of keywords in the first paragraph in case of empty paragraphs, or paragraphs filled with only shortcodes. 

* Adds a check for the first paragraph. Is this is empty, we will skip to the next paragraph (or use an empty string if all paragraphs are empty). 
This was done because when a plugin like visual composer sometimes adds a bunch of shortcodes in a `<p> ` tag, and reported that the keyword wasn't found in the first paragraph, however, the first paragraph with text did have the keyword. 
see https://github.com/Yoast/wordpress-seo/issues/4515#issuecomment-258885279

## Test instructions

* In the standalone version, add an empty paragraph, and 1 with text.
* Enter a keyword that is found in the 2nd paragraph. 
* The feedback should show the keyword was found in the first paragraph that contains text. 

Fixes #https://github.com/Yoast/wordpress-seo/issues/4515


js/researches/findKeywordInFirstParagraph.js
```diff
@@ -14,7 +16,13 @@ var escapeRegExp = require( "lodash/escapeRegExp" );
...
  */
 module.exports = function( paper ) {
-	var paragraph = matchParagraphs( paper.getText() );
+	var paragraphs = matchParagraphs( paper.getText() );
 	var keyword = escapeRegExp( paper.getKeyword().toLocaleLowerCase() );
-	return wordMatch( paragraph[ 0 ], keyword, paper.getLocale() );
+	var paragraph = reject(
+		paragraphs,
+		function( paragraph ) {
+			return isEmpty( paragraph );
```
andizer: Why don't you do `reject( paragraphs, isEmpty);`

andizer:
Cr done. Awesome 👍

PR #960, opened 2016-11-29
Remove unused files

terw-dan:
This PR removes some files that aren't used anymore. 

Back-compat was a backwards compatibility file we used before everything was converted to modules and is now obsolete. 

The standalone example is succeeded by the browserified example and isn't used anymore. 

Both can be removed.


boblinthorst:
CR: In grunt/config/uglify.js there are still references to the example scraper and it's minified variant. Those should be removed as well

boblinthorst:
Merged after a conversation with @terw-dan

PR #961, opened 2016-11-29
Renames specs

terw-dan:
Rename specs so they have the same filenames. 

Some specs have been merged, because there where 2. I did not change the tests themselves here, just a copy/paste so we don't lose any data.


boblinthorst:
CR: ok 👍

boblinthorst:
Unit-tests were fine 👍

PR #967, opened 2016-12-06
Add alternate source for tooltips.

### LINKED ISSUE #965: Tooltips: introduce the ability to use an alternate source for the tooltip text
See https://github.com/Yoast/wordpress-seo-premium/issues/884

Sometimes, there's the need to display some text in the displayed tooltip but then have a different text, with additional context or info, used for the `aria-label` attribute.

The rationale behind this is that while sighted users can get additional information from the context (other things they see around the UI control), screen reader users benefit from additional info and context added directly on the UI control. A typical example of this is the link suggestions "Copy" button mentioned in the issue above.

afercia:
Fixes #965.

This adds the ability to use a HTML data attribute to specify the text to be used for the tooltips. This way, the tooltip can display a "visual" info while the `aria-label` attribute is available to provide expanded info and context for assistive technologies.

The tooltips text is created via CSS generated content.

Usage:
- by default, it uses the `aria-label` text
- adding a CSS class to the base class, e.g. `yoast-tooltip yoast-tooltip-alt` it will use the text of a `data-label` attribute

### DIFF
--- a/css/_tooltips.scss
+++ b/css/_tooltips.scss
@@ -30,6 +30,14 @@ button.yoast-tooltip {
 	-webkit-font-smoothing: subpixel-antialiased;
 }
 
+/*
+ * Use a data attribute as alternative source for the tooltip text.
+ * Add this selector as modifier of the base selector.
+ */
+.yoast-tooltip-alt::after {
+	content: attr(data-label);
+}
+
 .yoast-tooltip::before {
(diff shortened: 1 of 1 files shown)


terw-dan:
CR Done 👍

terw-dan:
Acceptance 👍

PR #969, opened 2016-12-07
Adds passive voice assessment specs and renames spec files.

IreneStr:
- Adds specs for the passive voice assessment
- Renames some spec files that were missing 'spec' at the end.


terw-dan:
👍

PR #970, opened 2016-12-07
Deletes sentence length percentage score helper

IreneStr:
We no longer use a continuous scale for scoring. This makes the sentence length percentage score helper obsolete.


boblinthorst:
CR: ok 👍

PR #971, opened 2016-12-07
Make assessments for stopwords language dependant.

terw-dan:
## Summary

* Only applies stopword assessments for available languages

* Restricted the stopwords to only be checked in English. Currently we don't have other languages supported in JS than English, so running these checks on other languages don't make any sense. 

## Test instructions

* Set language to English (if that is not already the site language). 
* Create a keyword and slug with a stop word in it
* See that the stopwords assessments give feedback. 
* Set language to something else. 
* See that the stopwords assessments no longer return feedback. 

Related: #972

### DIFF
--- a/js/assessments/keywordStopWordsAssessment.js
+++ b/js/assessments/keywordStopWordsAssessment.js
@@ -1,5 +1,9 @@
 var AssessmentResult = require( "../values/AssessmentResult.js" );
 
+var getLanguageAvailability = require( "../helpers/getLanguageAvailability.js" );
+
+var availableLanguages = [ "en" ];
+
 /**
  * Calculate the score based on the amount of stop words in the keyword.
  * @param {number} stopWordCount The amount of stop words to be checked against.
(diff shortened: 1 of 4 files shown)


boblinthorst:
CR: ok 👍

js/assessments/keywordStopWordsAssessment.js
```diff
@@ -52,6 +56,7 @@ module.exports = {
 	identifier: "keywordStopWords",
 	getResult: keywordHasStopWordsAssessment,
 	isApplicable: function( paper ) {
-		return paper.hasKeyword();
+		var isLanguageAvailable = getLanguageAvailability( paper.getLocale(), availableLanguages );
```
afercia: When I set the site language in (for example) Italian and the user language in English, `paper.getLocale()` returns `en_US` while I guess the analysis should use the site language. I'm afraid this is because https://github.com/Yoast/wordpress-seo/pull/6215. We should probably pass both languages to the config and review what was done in https://github.com/Yoast/wordpress-seo/pull/6215

js/assessments/urlStopWordsAssessment.js
```diff
@@ -48,5 +52,8 @@ var urlHasStopWordsAssessment = function( paper, researcher, i18n ) {
 
 module.exports = {
 	identifier: "urlStopWords",
+	isApplicable: function( paper ) {
+		return getLanguageAvailability( paper.getLocale(), availableLanguages );
```
afercia: Same as above.

afercia:
Acceptance depends on https://github.com/Yoast/wordpress-seo/pull/6295

afercia:
Acceptance OK 👍

PR #978, opened 2016-12-13
Remove unused readability assessments

### LINKED ISSUE #977: Remove unused assessments
After discussing with Joost and Marieke, we've decided to remove the following inactive assessments, because it is unlikely we are going to use them in the foreseeable future:
- sentence length variation
- subheading length
- subheading presence
- subheading distribution too short
- paragraph too short

IreneStr:
## Summary

Removes the following inactive readability assessments:
- sentence length variation
- subheading length
- subheading presence
- subheading distribution too short
- paragraph too short

Also removes the following related helpers and researches:
- Is value too short helper
- Subheading length research
- Subheading presence research
- Sentence variation research

All associated specs are removed as well. 

This PR can be summarized in the following changelog entry: N/A

Fixes #977


IreneStr:
FYI: CodeClimate doesn't like code I haven't touched in this PR.

terw-dan:
CR:
I would suggest not removing the valueTooShort helper function, since we do use it's counterpart, the valueTooLong. 
CR 1 done.

Issue #958, opened 2016-11-29
Improve test coverage by adding missing specs. 


terw-dan:
In order to improve the test coverage we should add specs for the following files

/assessmenthelpers
- [x] checkForTooLongSentences

/assessments
- [x] textPresenceAssessment
- [ ] wordComplexityAssessment --> is currently not in use

/errors
- [x] invalidType
- [x] missingArgument

/interpreters
- [x] scoreToRating

/helpers
- [x] domManipulation
- [x] getLanguage
- [x] getTransitionWords
- [x] inRange
- [x] isValueTooLong
- [x] types

/values
- [x] Sentence

PR #983, opened 2016-12-13
Adds specs for inrange.js

terw-dan:
Add spec for inRange.js

After merge, please update #958


boblinthorst:
inRange function not exactly what I'd expect, but CR 👍

PR #986, opened 2016-12-20
Deletes Analyzer Config

IreneStr:
Two out of three values from the analyzerConfig were never used.
The third value was only used in one single file.


terw-dan:
CR done 👍

PR #987, opened 2016-12-20
Add unittest for invalidType

IreneStr:
Add unittest for invalidType.

After merge, please update #958.

### DIFF
--- a/spec/errors/invalidTypeSpec.js
+++ b/spec/errors/invalidTypeSpec.js
@@ -1,0 +1,10 @@
+var invalidType = require( "../../js/errors/invalidType.js" );
+
+var error = new invalidType( "Error" );
+
+describe( "Invalid type error", function(){
+	it( "throws an error", function(){
+		expect ( error.message ).toBe( "Error" );
+		expect ( error.name ).toBe( "InvalidTypeError" );
+	});
+});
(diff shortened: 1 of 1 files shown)


spec/errors/invalidTypeSpec.js
```diff
@@ -0,0 +1,10 @@
+var invalidType = require( "../../js/errors/invalidType.js" );
+
+var error = new invalidType( "Error" );
+
+describe( "Invalid type error", function(){
+	it( "throws an error", function(){
```
atimmer: It doesn't throw an error, `InvalidType` is an error to be thrown.

spec/errors/invalidTypeSpec.js
```diff
@@ -0,0 +1,10 @@
+var invalidType = require( "../../js/errors/invalidType.js" );
```
atimmer: It is best to keep the variable name consistent with the variable name in the `invalidType` file, so this should probably be `InvalidTypeError`.

Also, we support ES2015 in the unit tests so you can use the new import syntax:
```js
import InvalidTypeError from "../../js/errors/invalidType";
```

spec/errors/invalidTypeSpec.js
```diff
@@ -0,0 +1,10 @@
...
+
+var error = new invalidType( "Error" );
```
atimmer: I would move this statement within the test.

PR #1000, opened 2016-12-29
Fixes the wording of the feedback string for the subheading too long assessment. 

terw-dan:
Fixes #https://github.com/Yoast/wordpress-seo/issues/5089

The wording of the feedback string wasn't correct. This fixes that. This is a string only change.


andizer:
CR done 👍

PR #1021, opened 2017-01-06
Place the measurement elements holder off screen.

### LINKED ISSUE #1006: createMeasurementElements() breaks the responsive view
See https://github.com/Yoast/wordpress-seo/issues/5847

`SnippetPreview.prototype.createMeasurementElements()` creates and appends to the `body` a couple elements with a fixed width. The elements are filled with the meta description content and their width is taken from the `meta_container` div which currently is `598px` even in the responsive view.

Even if the elements are hidden with `visibility: hidden` they still take space in the page so the actual content is wider then the viewport width, causing the content to be scrollable horizontally.

Additionally, when opening modal dialogs that are supposed to have a width of 100%, these modals are wider than expected, as pointed out in https://github.com/Yoast/wordpress-seo/issues/5847

afercia:
## Summary

Move the measurement elements holder `<div>` off-screen so it doesn't break the responsive view when opening the media modal.

I've initially considered to put the required CSS inline in the JS template but this would not allow to swap the `left` property for RTL languages. The CSS needs to be in a file that goes through the RTL files generation.
Also, added an `aria-hidden` attribute because we don't want this element content to be available for assistive technologies.

## Test instructions

- be sure the `yoastseo` module is "yarn linked"
- on wordpress-seo, build the CSS and JS
- emulate a responsive view (e.g. iPhone) using your browser's dev tools
- edit a post and click on "Add Media" to open the WordPress media modal
- ensure the media modal has a correct size and doesn't extend on the right; for reference see the screenshot below

On the left: good. On the right, bad:

screen shot 2017-01-03 at 09 33 44

Fixes #1006


andizer:
Can you resolve the merge conflicts?

afercia:
@andizer: done.

andizer:
Acceptance done 👍

PR #1031, opened 2017-01-13
Remove the rounded border from the snippet preview

### LINKED ISSUE #1030: Remove the rounded borders of the snippet preview
See https://github.com/Yoast/wordpress-seo/issues/6374

This very, very, light border is almost unnoticeable but affects the max-width of the snippet preview which currently is `598px` and should be `600px` instead:

screen shot 2017-01-13 at 13 23 29

afercia:
## Summary

Remove the very light rounded border from the snippet preview, since it affects the snippet preview width. See https://github.com/Yoast/wordpress-seo/issues/6374

## Test instructions

- check the browserified example
- in your browser's dev tools, make the `.form-container` larger so the snippet preview gets wider till its max-width
- check the snippet preview width is actually 598px, it should be 600px (see screenshot below)

screen shot 2017-01-13 at 13 23 29

- switch to this branch and check the border is gone and the width highlighted in the screenshot above is now `600px`

Fixes #1030


andizer:
CR Done 👍

andizer:
Acceptance done 👍

afercia:
Hm yes we could adjust other things like padding etc. but we've quickly discussed this and if I remember correctly there was consensus to remove the borders. Have you noticed any breakage without the borders?

atimmer:
The borders are also used for the other styled sections, so this changes the visuals of the whole metabox.

afercia:
I know :) I think the Social Previews have their own rounded border too, that should be handled separately. There's no hurry though, this is probably a change that needs to be pondered.

Still, I can barely see that border, even on a Retina display.

terw-dan:
@afercia can you fix the merge conflicts. 

I've tested this PR, in the standalone it looks good, however, when using this inside WordPress it still has rounded corners in the metabox. Is there a way we can update this so we have a consistent styling everywhere?

afercia:
@terw-dan done 🍸 
I had troubles to test it inside WordPress too. What I've done was building everything (yoast components, YoastSEO.js) and then build everything in Yoast SEO.
Not sure but I think I've also `yarn link` 'ed everything again.

terw-dan:
I've tested this in conjunction with https://github.com/Yoast/yoast-components/pull/130
Acceptance 👍

PR #1036, opened 2017-01-17
Fix the preview toggle buttons translations.

### LINKED ISSUE #1035: Fix translations of the preview toggle buttons aria-label
The preview toggle buttons use `aria-label` attributes but their translation should be fixed, the js template references wrong property names in the i18n object.

afercia:
## Summary

Fixes the preview toggle buttons translations in the JS template.

## Test instructions

Please notice I could test this just  in the browserified example as I'm getting errors while building the JS on Premium.

Fixes #1035


afercia:
Solved the issues while building the JS in the `premium` dir of the Premium plugin so I could test it inside the meta box: aria labels working correctly:

screen shot 2017-01-17 at 11 07 13

PR #1038, opened 2017-01-17
Make the mark buttons tooltips always visible.

### LINKED ISSUE #1037: Mark buttons tooltips: make them visible even when the mark buttons are active
As sometimes happens when one looks back at things after a few months, I can't find a good reason why the tooltips on the assessment-results mark buttons are hidden when the buttons are active. Hiding the tooltips was [intentionally implemented](https://github.com/Yoast/YoastSEO.js/commit/961019c03bb67189dcbc3847779474d486b4a0a3) as a "feature" but now I tend to think it's better to show them even when the buttons are active.

What happens now:
hovering or focusing a mark button, a tooltip appears and displays the text used for the aria label `Mark this result in the text`:

screen shot 2017-01-17 at 14 11 32

After the button is clicked, it becomes "active", the aria label text changes in "Remove marks in the text" but no tooltip appears:

screen shot 2017-01-17 at 14 19 53

Instead, we should display the tooltip even when the button is active, to help users understand tue functionality of the button has changed:

screen shot 2017-01-17 at 14 11 41

afercia:
## Summary

Make the mark buttons tooltips always visible, even when the buttons are "active".

## Test instructions

You can test this on the browserified example or built the CSS on the Yoast SEO plugin and test there. Things to check:

Hovering or focusing a mark button, a tooltip appears and displays the text used for the aria label `Mark this result in the text`:

screen shot 2017-01-17 at 14 11 32

After the button is clicked, it becomes "active", the aria label text changes in "Remove marks in the text" and the tooltip appears again on hover and focus, showing the updated text:

screen shot 2017-01-17 at 14 11 41

Fixes #1037

### DIFF
--- a/css/_analysis.scss
+++ b/css/_analysis.scss
@@ -87,11 +87,6 @@ li.score:after {
 			&-active {
 				background-image: url(svg-icon-eye($color_marker_active));
 				background-color: #a4286a;
-
-				&.yoast-tooltip::before,
-				&.yoast-tooltip::after {
-					display: none;
-				}
 			}
 
 			&-disabled {


andizer:
CR done

andizer:
Acceptance done

PR #1049, opened 2017-01-20
Snippet preview toggle buttons accessibility improvements

### LINKED ISSUE #1047: Snippet preview toggle buttons accessibility improvements
I'd like to propose three a11y improvements for the new snippet preview toggle buttons.

1
Buttons placement:
the new buttons are on the right of the Edit snippet button so the current logical order is:
"the Snippet preview box with details" > "Edit snippet" > "Mobile preview" > "Desktop preview"
the a11y issue here is that the "Edit snippet" button expands the editor below and, when activated, its `aria-expanded` attribute announces a change of state, giving users the indication that something in the page is now expanded.
For this reason, the expanded panel should immediately follow the button in the source order. Imagine you can't see and:
- you activate the button
- you hear "expanded" announced by your screen reader
- and then what? you have no idea what exactly is now expanded in the page
- the most natural action is going forward with exploring content, hoping to find the "expanded thing"
- instead, the first thing you find are the toggle buttons

Having the "expanded thing" right after the control that was activated solves this issue and ideally should be a pattern to always use in similar cases.
For reference, worth nothing ARIA has also an [aria-controls attribute](https://www.w3.org/TR/wai-aria/states_and_properties#aria-controls) but support is very limited (as far as I know, only JAWS with _some browsers_). Also, the only thing it does is offering users a shortcut to "jump" to the controlled element, so it's not a great option.
For more details about `aria-controls` and the reasons why we should not use it, see this post (notice the very meaningful post title):

Aria-Controls is Poop
http://www.heydonworks.com/article/aria-controls-is-poop

Changing the buttons order would result in a new logical order:
"the Snippet preview box with details" > "Mobile preview" > "Desktop preview" > "Edit snippet"
and visually would be something like this:

screen shot 2017-01-17 at 09 30 25

2
There is no audible feedback about _which_ of the buttons is currently active. This can be easily solved using an `aria-pressed` attribute which would make screen readers announce the buttons as "toggle buttons" and also their selected state:

screen shot 2017-01-20 at 11 39 18

3
Lastly, I'd consider to add tooltips to the buttons. They're icon-only controls and tooltips could help users better understand their purpose.

afercia:
## Summary

- moves the preview toggle buttons on the left, before the "Edit snippet" button
- adds `aria-pressed` attributes
- adds `type="button"` attribute, so `preventDefault()` can be removed
- **removes `overflow-x: auto;` from the snippet preview: it shouldn't be necessary any longer since the snippet should be responsive now (please double check this)**
- CSS adjustments
- removes Arial font-family since the font metrics affect margins set in `em`
- adds tooltips
- adds events and functions to handle the tooltips: this mimics what we are already doing for the tooltips in the metabox left sidebar

Please see screenshots and discussion on the related issue.

About the tooltips: open to any feedback 🙂 we can also remove them and iterate later.

Fixes #1047


andizer:
CR and acceptance done 👍

PR #1078, opened 2017-02-20
Improve the snippet preview toggler

### LINKED ISSUE #1071: The snippet preview toggler fails under certain conditions
For reference, see #954 and #1054.

The snippet preview toggler should automatically switch to the "mobile preview" when the snippet container width doesn't allow to show the "desktop preview". This works well on the first load of the page. Shrink your browser's window, and you will see the snippet preview automatically switches to mobile, thus preserving the responsiveness of the whole metabox.

Note: the relevant viewport width breakpoint is 1249 pixels *when the WP admin menu is expanded*.

Dumping into the console some numbers, see screenshot below:

screen shot 2017-02-15 at 14 59 37

Hope the screenshot helps making clear that, when shrinking the window, the snippet container is allowed to shrink and when its width gets less than 640 pixels, the automatic switch happens.

However, now try this:
- make the browser window larger
- refresh the page
- click on the "mobile" button
- click on the "desktop" button

A couple CSS rules just got applied to some elements:
```
.snippet-editor__view--desktop {
    width: 600px;
    background: #fff;
}

.snippet-editor__view--desktop .snippet_container__title {
    width: 600px;
}
```

Those elements are children of the snippet preview and have a fixed width. Thus, the snippet preview container is not allowed to shrink under 640 pixels (600 width + padding) any longer. The `previewWidth` is always `640` and the condition to make the toggler work properly is never met:

screen shot 2017-02-15 at 15 00 42

To my understanding, the fixed 600 pixels width is probably meant to make the desktop preview always have a... desktop width. But that leads us to the original problem: on small screens, pressing the Desktop button will make the metabox overlap and hide some content on the right.

We should probably clarify what is the exact behavior to implement. Maybe one relevant point is:
should the toggle buttons be hidden when the preview container is under the allowed `minimumDesktopWidth`? Reading the the original PR, see @atimmer [comment](https://github.com/Yoast/YoastSEO.js/pull/954#issuecomment-267362508)
> The button for the desktop variant should be hidden if there is no room for the desktop variant.

Not sure if the plans have changed later, but this hasn't been implemented so on small screens we still have the original problem.

Couple additional issues:

1
Should `handleWindowResizing` run also on the `orientationchange` event? As far as I know, not all platforms/devices/browsers fire the `resize` event, especially on mobile.

3
In the mobile mode, avoid the title to wrap in multiple lines, see https://github.com/Yoast/wordpress-seo/issues/6548 (aside: in the issue, it is correctly reported that the relevant viewport width breakpoint is 1249 pixels)

afercia:
## Summary

Improve the snippet preview desktop and mobile views. Removes automatic switching of the views.

On init, the desktop or mobile view mode is set based on the width of the snippet container. We can't use media queries since the available width depends on several factors such as the WP admin menu expanded/collapsed state or the 1/2 columns layout users can set in the WP Screen Options.

There's no more automatic switching to the mobile view.

The desktop view is always set to `overflow: auto` so it always works and it's up to browsers to show the horizontal scrollbar.

When the window gets resized or the WP admin menu state changes, the width calculation is still used but just to show or hide the "Scroll Hint" message. This part is optional, no objections if we prefer to remove it, just thought it was something useful for users.

Trying to summarize the changes:

- The metabox sidebar and the content sections use a new layout based on float, no more table layout: this is necessary because tables expand their width based on their content and we don't want that.
- Changed the snippet preview template to remove spaces between the slug elements
- When there's not enough space for the desktop view and so it's scrollable, a message "Scroll to see the preview content." gets displayed. Any thoughts for a better wording welcome.
- In the desktop view:
  - title, slug, and meta description have a fixed 600 pixels width
  - title and slug don't wrap in new lines and use ellipsis
  - meta description wraps in new lines
- In the mobile view:
  - title, slug, and meta description have a fluid width
  - slug doesn't wrap in new lines and doesn't have the "down arrow"
  - title and meta wrap in new lines, thus they don't have ellipsis

Note 1: in the mobile view, the title length check can't work correctly since the title wraps in new lines

Note 2: in the mobile view, the snippet elements width isn't 600 pixels anyways, because of the additional 16px + 16px padding

Note 3: as far as I see, in the mobile view Google truncates the url a bit before the max limit, not sure what we can do here.

Todo: evaluate if we want to add a short explanation in the inline help (i.e. the "?" button)

Todo: evaluate if we want to highlight the view switch in the UI, for example changing the "Snippet preview" heading in "Desktop snippet preview" and "Mobile snippet preview"

## Test instructions

- needs https://github.com/Yoast/wordpress-seo/pull/6700
- assets are already built, this should make testing easier, hopefully
- make sure YoastSEO.js is yarn linked
- on wordpress-seo, switch to the `stories/af/6682-snippet-preview-toggler` branch

Fixes #1071


afercia:
Some screenshots:

desktop 

screen shot 2017-02-20 at 12 58 09

desktop shorter slug

shorter url

mobile

screen shot 2017-02-20 at 12 58 29

responsive, desktop view

screen shot 2017-02-20 at 12 52 27

responsive, mobile view

screen shot 2017-02-20 at 12 52 54

scrollbar on Windows are bigger, of course

screenshot 63

thulshof:
CR done 👍

css/_snippet-editor.scss
```diff
@@ -99,39 +101,77 @@ $snippet_width: 600px;
...
 
+			// In the mobile view title, slug, and meta description have a fluid width.
+			// Title and meta description wrap in new lines. Slug doesn't wrap.
 			.snippet_container {
 				display: block;
 				line-height: 20px;
 
 				&__title, &__url, &__meta {
 					width: 100%;
+					// Worth noting inside the yoast-section everything is already border-box.
```
andizer: I guess there is a type in `noting`

css/_snippet-editor.scss
```diff
@@ -99,39 +101,77 @@ $snippet_width: 600px;
...
+					// setting this proprty on specific element is needed just for the browserified example.
```
andizer: typo in `proprty`

css/_snippet-editor.scss
```diff
@@ -99,39 +101,77 @@ $snippet_width: 600px;
...
 		&--desktop {
-		width: $snippet_width;
-		background: #FFF;
-		  .snippet_container {
+			width: $snippet_width;
+			max-width: 100%;
+			overflow: auto;
+		}
+
+		// Used for the desktop view and then overriden for the mobile one:
```
andizer: `overriden` should be `overridden`

afercia:
@andizer thanks! Fixed the typos and tried to make more clear that "noting" is a verb form.

andizer:
@afercia I see what you meant with "noting". My mistake.

CarolineGeven:
There's an issue with the scroll bars, I've added a video to show what I mean. Please advise @afercia 

http://sendvid.com/hguyb575

Tested with Chrome, Safari and Firefox.

afercia:
OK so, I can't fully reproduce what happens on @CarolineGeven's video. What I see here is the _vertical_ scrollbar quickly appearing when hovering on the description in Firefox.

Regardless, the current implementation breaks the "arrows" that are supposed to appear on the left of the title, url, and description when hovering them or when the related fields in the editor have focus. I've made them visible in the screenshot below, to clarify the issue:

screen shot 2017-04-19 at 11 55 06

Title, url, and description are now in a wrapper element with `overflow: auto`. The title has also an `overflow: hidden`. So, the arrows are not visible. When hovering, they still pop-up and seems they cause the scrollbars to appear.

We should probably rework this part. If I remember correctly, we already had some troubles with these arrows and maybe they have proven to be a bit fragile and prone to bugs. Honestly, they don't look so great too 🙂  Maybe worth considering a different design? /cc @moorscode @hedgefield 

However, even an outline would be cut-off by the overflow property, so I'm not really sure what a new design could look like:

screen shot 2017-04-19 at 12 10 54

Maybe an underline? Or background change? The simpler, the better!

moorscode:
We should keep the preview as close to the actual Google representation as possible.

I believe we need to keep the 'marker' to show what we are doing outside of the visible preview, which the arrows do in a nice way.

**Future solution**
The best solution would be to solve this is via a React component so we can solve the HTML Structure problem we're having by making the arrows state-aware and not have them coupled to the actual title/description/link elements.

**Practical solution**
I'm seeing that the `.snippet_container .title` has an `overflow: hidden` but no `height` defined. So if the `overflow` is not applied on the containing object, the problem would be solved because the text won't be displayed when it is too long, but the arrows won't be affected.

afercia:
It's a bit more complicated :) the `overflow: hidden;` on the title is needed to truncate the title with ellipsis when it's too long. But we could move that to some inner elements as already done for the URL.

The real issue is the `overflow: auto` on the `#snippet-preview-view` element, which is needed to allow to horizontally scroll the preview, and it's the key to make all this work.

afercia:
Refreshed CSS to correctly display the arrows. This should be carefully checked, I've changed the width/padding of some main elements of the snippet preview. Hopefully this won't impact on the JS measurements here, but I might have missed something.

Other changes:
- improvements for the mobile snippet preview to more accurately match the big G mobile snippet

andizer:
Cr Done

andizer:
Acceptance done. @afercia Can you resolve the conflicts?

afercia:
@andizer: done!

PR #1096, opened 2017-02-22
Change ranges to include 0.5 keyword density

### LINKED ISSUE #896: Keyword density assessment disappears if density is exactly 0.5%
If the keyword density is exactly 0.5%, it returns an empty text and empty score, because the if statements do not cover this percentage.  

Please also add a test case for a density of 0.5%.

IreneStr:
## Summary

This PR can be summarized in the following changelog entry: Fixes bug where the keyword density assessment would disappear when the density was 0.5%.

* Introduces inRange function that includes both the start as well as the end of the range.

## Test instructions

* Write a text with a keyword density of 0.5%.
* Check whether the keyword density assessment/bullet is present.

Fixes #896


terw-dan:
CR and Acceptance 👍

PR #1118, opened 2017-03-07
Fix example code

alxy:
The postUrl is for a single post, the searchUrl for multiple

## Summary

* Fix the provided example code


atimmer:
👍  Nice find

PR #1148, opened 2017-05-19
Changes the copy for 'keyword in headings' assessment result

CarolineGeven:
Changes the keyword copy to `The focus keyword appears only in {x} (out of {y}) subheadings in your copy. Try to use it in at least one more subheading.`

Fixes https://github.com/Yoast/wordpress-seo/issues/7133


andizer:
CR & Acceptance done 👍

PR #1162, opened 2017-06-08
Change page title to snippet title

### LINKED ISSUE #120: Confusing which page title I have to change
When I enter a titel for my post, the snippet preview takes this title also, which is nice. But the text analysis is giving the following message:
edit_page_ _theme_ _wordpress

So I changed my page title to improve my post. But the message is still there. Only when I change the title in the snippet preview the line will go away. For me `page title` is confusing, it would be better to reference to the snippet title.

manuelaugustin:
## Summary

Change "page title" to "snippet title" in the snippet title feedback string.

Fixes #120

### DIFF
--- a/js/assessments/seo/pageTitleWidthAssessment.js
+++ b/js/assessments/seo/pageTitleWidthAssessment.js
@@ -87,25 +87,25 @@ class PageTitleWidthAssesment extends Assessment {
 		if ( inRange( pageTitleWidth, 1, 400 ) ) {
 			return i18n.dgettext(
 				"js-text-analysis",
-				"The page title is too short. Use the space to add keyword variations or create compelling call-to-action copy."
+				"The snippet title is too short. Use the space to add keyword variations or create compelling call-to-action copy."
 			);
 		}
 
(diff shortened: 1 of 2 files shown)


CarolineGeven:
I'm unsure if this is the solution we want. `snippet title` is more clear than `page title` when the user edited the snippet title. But when a user does not edit the snippet title, but his page title, we copy the page title to the snippet title. Any edits to the title, would be done from the page title field, not from the snippet title. I'm not sure if changing `page title` to `snippet title` solves this. Perhaps we need to check when the snippet title is edited and when we use the page title as snippet title.

@hedgefield and @moorscode could you please look at this from the UX point of view and come with a suggestion?

hedgefield:
@CarolineGeven that makes sense to me, if we change the warning based on where the edit was made/which field currently has priority for SEO. Could we also include a link with an anchor on either field, so users can jump straight to the one they need?

moorscode:
We are referring to it as `SEO title` instead of `snippet` or `page` title. Which gives a more generalized approach.

The SEO title can be adjust by modifying the page title or the snippet title.

Please change accordingly.

CarolineGeven:
CR and acceptance done 👍

PR #1194, opened 2017-07-12
Tooltip multiline compat for Edge and better alignment.

### LINKED ISSUE #1183: Tooltips: update CSS for compatibility with Edge and maybe move them to yoast-components
Noticed while working on https://github.com/Yoast/wordpress-seo/pull/7372: maybe the fix used there for Edge should be included in the main tooltips file.

Looks like Microsoft Edge doesn't support `width: max-content;`
http://caniuse.com/#search=max-content

Current state is: "MS Edge status: Under Consideration" with low priority. On the other hand, the tooltips CSS comes with a hack to cover IE 11 so no problems there.

There are CSS hacks to target Edge, but I wouldn't recommend them. Looks like Edge just needs a computable width, see https://github.com/Yoast/wordpress-seo/pull/7372
Worth noting this applies only to the "multiline" tooltips, where a width is needed to make the text wrap in new lines.

Also to consider: we're using tooltips in a few places now, maybe worth moving them to yoast-components? Not sure why they should still be in YoastSEO.js.

### LINKED ISSUE #1189: Tooltips: improve vertical alignment of text
Noticed during work done for https://github.com/Yoast/wordpress-seo/pull/7372 and https://github.com/Yoast/wordpress-seo/pull/7426

The vertical alignment of text within the tooltips is a bit off: should be moved down by 1 pixel or two. However, it should be checked on all browsers and also on Windows, since the font used across platform is different and the font metrics could change the rendering.

screen shot 2017-07-10 at 12 10 11

screen shot 2017-07-10 at 12 10 42

afercia:
Fixes the lack of support for`width: max-content` in Microsoft Edge. This means in this browser, the actual multiline tooltip width will be the same as `max-width`, while in modern browsers it is allowed to be _less_ than that.

Improves a bit the vertical alignment of the text. See before and after:

ezgif-2-6ec6135fec

Fixes #1183 
Fixes #1189

### DIFF
--- a/css/_tooltips.scss
+++ b/css/_tooltips.scss
@@ -11,13 +11,13 @@ button.yoast-tooltip {
 	display: none;
 	position: absolute;
 	z-index: 1000000;
-	padding: 5px 8px;
+	padding: 6px 8px 5px;
 	border-radius: 3px;
 	opacity: 0;
 	color: #fff;
 	background: rgba(0,0,0,0.8);
 	text-shadow: none;
-	font: normal normal 11px/1.5 Helvetica, arial, nimbussansl, liberationsans, freesans, clean, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";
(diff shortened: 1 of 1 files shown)


afercia:
Looks like node-sass is failing.

atimmer:
Looks good to me

PR #1228, opened 2017-08-25
Added analysis results callbacks

### LINKED ISSUE #1227: Allow extraction of content and keyword analysis from app through callbacks.
## Problem

With an eye on reactifying wordpress-seo, we need a way of extracting a data from the app through callbacks. Currently, the rendering of the analysis results is done by the AssessorRenderer, but the rendering should be moved to a React component. To be able to re-render the react application when new analysis results are available, a callback needs to notify the react or redux store to update.

## Proposed solution

The app has a callbacks-section, which will need 2 new callbacks: `updatedContentAnalysis` and `updatedKeywordsAnalysis`. These callbacks should be called by the app with the new content analysis and keyword analysis results. Also, they should be called in the execution of the `runAnalysis` function.

Also, the responsibility for rendering should be removed from `YoastSEO.js`. So all rendering related functions of this library should be deprecated.

Xyfi:
## Summary

* Added `updatedKeywordsResults` and `updatedContentResults` callbacks to `app.js`.

* 

## Test instructions

1. In `examples/browserified/example.js` add the following callbacks to the `callbacks` section of the app definition:

```js
updatedKeywordsResults: function( results, overallScore ) {
	console.log( results, overallScore );
},
updatedContentResults: function( results, overallScore ) {
	console.log( results, overallScore );
},
```
2. Run `grunt build`.
3. Open `examples/browserified/index.html` in your browser
4. Open your developer tools, and see the updated analysis results as you edit the text in the input fields.

Fixes #1227

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -183,6 +183,18 @@ function verifyArguments( args ) {
  */
 
 /**
+ * @callback YoastSEO.App~updatedContentResults
+ *
+ * @param {object} results The updated page analysis results.
+ */
+
+/**
+ * @callback YoastSEO.App~updatedKeywordsResults
+ *
+ * @param {object} results The updated keyword analysis results.
+ */
+
+/**
  * Loader for the analyzer, loads the eventbinder and the elementdefiner
  *
  * @param {Object} args The arguments passed to the loader.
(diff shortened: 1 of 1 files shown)


js/app.js
```diff
@@ -183,6 +183,18 @@ function verifyArguments( args ) {
  */
 
 /**
+ * @callback YoastSEO.App~updatedContentResults
+ *
+ * @param {object} results The updated page analysis results.
```
atimmer: The overall score also needs to be documented as a param.

As for the order of the arguments, I would put the overal score first. Less specific first.

boblinthorst:
CR and acceptance 👍. Content score didn't seem to vary, which is strage to me, but that shouldn't be as a result of this pull.

PR #1244, opened 2017-09-26
Add filter for exception words between auxiliary and passive participle

### LINKED ISSUE #920: Check for 'have/has' before participle in passive voice assessment
We should check for 'have' and 'has' directly preceding the participle, to prevent sentences like "Here is a list of ten beliefs I have adopted" from being marked as passive. We could add these words to the hasHavingException in the EnglishParticiple object.

Additionally, these auxiliaries should be filtered whenever they occur between the auxiliary and the participle, as in "the reason we're seeing the increase in this disease is because of the excess body weight we've **been** dealing with for the past several decades, but the obesity epidemic _has_ **affected** everyone". Cf. issue #973 for German; possibly add other auxiliaries such as 'can'.

manuelaugustin:
## Summary

This PR can be summarized in the following changelog entry: Adds a filter to avoid incorrectly marking participles as passive when the participle belongs to a non-passive auxiliary. E.g., 'enjoyed' in 'It's something I've always enjoyed doing' will now be correctly marked as non-passive because it belongs to 'I've' and not 'It's'.

## Test instructions

- Use the browserified example. Don't forget to run `grunt build:js`.
- Paste the sentence `It's something I've always enjoyed doing`.
- Make sure it is not marked as passive.

Fixes #920

### DIFF
--- a/spec/researches/english/EnglishParticipleSpec.js
+++ b/spec/researches/english/EnglishParticipleSpec.js
@@ -7,15 +7,17 @@ describe( "A test for checking the English Participle", function() {
 		expect( mockParticiple.isNonVerbEndingEd() ).toBe( false );
 		expect( mockParticiple.hasRidException() ).toBe( false );
 		expect( mockParticiple.directPrecedenceException() ).toBe( false );
+		expect( mockParticiple.precedenceException() ).toBe( false );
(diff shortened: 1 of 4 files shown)


Xyfi:
CR :+1: 
Acceptance should be done by @IreneStr

IreneStr:
Acceptance 👌

PR #1269, opened 2017-10-24
Add excited to exception -ed verbs

IreneStr:
## Summary

This PR can be summarized in the following changelog entry: Add 'excited' to the list of non-verbs ending in -ed. To be + excited is usually not passive.


manuelaugustin:
CR done

PR #1273, opened 2017-10-31
Restore check-coverage in .travis.yml

Xyfi:
I accidentally removed `yarn run test-coverage` from `.travis.yml`, because it wasn't in `package.json` and that caused the build on Travis to fail. This PR fixes that.


Xyfi:
The command was obsolete, and the removal was justified.

PR #1278, opened 2017-11-03
Add sentences research

atimmer:
I have disabled `no-var-requires` because it is not possible to use `import` with a file that uses `module.exports = `. To use `import` that file needs to actually do an `export`. Because `getSentences` doesn't do an `export` we have to use `require`. So for now lets disable `no-var-requires`.


IreneStr:
CR 👍

PR #1281, opened 2017-11-06
Escape keyword in url and add spec

IreneStr:
## Summary

This PR can be summarized in the following changelog entry: Escapes the keyword in the url generated in the PreviouslyUsedKeywords assessment.

* I used `encodeURIComponent` rather than `encodeURI` to make sure that forward slashes and other characters that can break an URL are escaped.

## Test instructions

* On line 84 of previouslyUsedKeywords.js, add `console.log(url);`
* Link yoastSEO.js to wordpress SEO. Don't forget to run grunt build.
* Create two posts with the same focus keyword. Make sure the focus keyword contains a special character, like a forward slash.
* Check whether the special character is escaped in the url in the console.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,7 +1,4 @@
 {
   "presets": [ "es2015" ],
-  "ignore": [
-    "js/templates.js",
-    "src/templates.js"
-  ]
+  "plugins": [ "transform-object-rest-spread" ],
 }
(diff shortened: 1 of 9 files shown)


Xyfi:
CR :+1:

Xyfi:
Acceptance :+1: Merged master into this PR.

PR #1289, opened 2017-11-10
Add Spanish stopwords

### LINKED ISSUE #1280: Add stopwords for Spanish
Luis Benítez has sent us an excel sheet with Spanish stop words and the following links:
https://www.ranks.nl/stopwords/spanish
http://googleseo.marketing/lista-de-stop-words-o-palabras-vacias-en-espanol/

We should compose a Spanish stop word list based on these lists and build a language specific implementation of the stopword researches.

IreneStr:
## Summary

This PR can be summarized in the following changelog entry: Add Spanish support for the stopWordsInUrl and stopWordsInKeyword assessment. Props Luis Benítez.

## Relevant technical choices: Adds support for the urlStopWordsAssessment and keywordStopWordsAssessment for Spanish.

* Add Spanish stopwords list
* I've used the same structure as is used for the syllable count/Flesch Reading: a helper function that provides the research with the correct stopwords list (`src/config/stopwords.js`) and json files with word list in config folder (`src/config/stopwords/en.json` and `es.json`)
* Add Spanish support for the keywordStopWordsAssessment and urlStopWordsAssessment

## Test instructions

* Link this branch to your local WordPress install. Don't forget to run a `grunt build:js`.
* Set the site language to Spanish. 
* Create a post with a focus keyword containing one or more of the words from `es.json`.
* Put also one of the stopwords in the slug of the post.
* Check whether the following two bullets are shown:
schermafbeelding_2017-11-10_om_14_58_11

* Repeat the above steps for English to check I didn't break anything for English.
* Repeat the above steps for another locale. The above bullet should not show up, even when you are using English/Spanish stop words.

Fixes #1280

### DIFF
--- a/src/config/stopwords.js
+++ b/src/config/stopwords.js
@@ -1,5 +1,4 @@
-let getLanguage = require( "../helpers/getLanguage.js" );
-let isUndefined = require( "lodash/isUndefined" );
+import getLanguage from "../helpers/getLanguage.js";
 
 let es = require( "./stopwords/es.json" ).stopwords;
 let en = require( './stopwords/en.json' ).stopwords;
--- a/src/config/stopwords.js
+++ b/src/config/stopwords.js
@@ -1,17 +1,3 @@
-/** @module config/stopwords */
-
-/**
- * Returns an array with stopwords to be used by the analyzer.
- *
- * @returns {Array} stopwords The array filled with stopwords.
- */
-/*module.exports = function() {
-	return [ "a", "about", "above", "after", "again", "against", "all", "am", "an", "and", "any", "are", "as", "at", "be", "because", "been", "before", "being", "below", "between", "both", "but", "by", "could", "did", "do", "does", "doing", "down", "during", "each", "few", "for", "from", "further", "had", "has", "have", "having", "he", "he'd", "he'll", "he's", "her", "here", "here's", "hers", "herself", "him", "himself", "his", "how", "how's", "i", "i'd", "i'll", "i'm", "i've", "if", "in", "into", "is", "it", "it's", "its", "itself", "let's", "me", "more", "most", "my", "myself", "nor", "of", "on", "once", "only", "or", "other", "ought", "our", "ours", "ourselves", "out", "over", "own", "same", "she", "she'd", "she'll", "she's", "should", "so", "some", "such", "than", "that", "that's", "the", "their", "theirs", "them", "themselves", "then", "there", "there's", "these", "they", "they'd", "they'll", "they're", "they've", "this", "those", "through", "to", "too", "under", "until", "up", "very", "was", "we", "we'd", "we'll", "we're", "we've", "were", "what", "what's", "when", "when's", "where", "where's", "which", "while", "who", "who's", "whom", "why", "why's", "with", "would", "you", "you'd", "you'll", "you're", "you've", "your", "yours", "yourself", "yourselves" ];
-};
-
-
-*/
-
 let getLanguage = require( "../helpers/getLanguage.js" );
 let isUndefined = require( "lodash/isUndefined" );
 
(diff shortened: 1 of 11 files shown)


src/config/stopwords.js
```diff
@@ -1,10 +1,17 @@
...
+
+let es = require( "./stopwords/es.json" ).stopwords;
+let en = require( './stopwords/en.json' ).stopwords;
+
+let languages = { es, en };
+
+module.exports = function( locale = "en_US" ) {
+	let language = getLanguage( locale );
+
+	if( languages.hasOwnProperty( language ) ) {
```
afercia: missing space

src/researches/stopWordsInKeyword.js
```diff
@@ -1,15 +1,15 @@
 /** @module researches/stopWordsInKeyword */
 
-var stopWordsInText = require( "./stopWordsInText.js" );
-
-var escapeRegExp = require( "lodash/escapeRegExp" );
+import stopWordsInText from  "./stopWordsInText.js";
```
afercia: double space

afercia:
CR 🚧 

Minor coding standards. Also, in `src/researches/stopWordsInUrl.js` there's still a require for `stopWordsInText` while in other files it has been changed in an import.

afercia:
Testing a bit more, seems there's a problem with accented letters.

In this case, it's returning both `como` and `cómo`

screen shot 2017-11-13 at 10 19 06

an din this case, both `cuando` and `cuándo`

screen shot 2017-11-13 at 10 30 58

afercia:
To my understanding, here's what happens, for example with `como` and `cómo`:

`como`:
the stop words list is processed by `createWordRegex.js` where `replaceDiacritics()` replaces the `ó` in the stop words list with `o`. Then, there are 2 `como` in the list and the entered term matches twice

`cómo`:
actually the processed stop words list doesn't have diacritics (see above) but more importantly regexes in JavaScript take into account what a "word character" is in JavaScript:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions
> Note: JavaScript's regular expression engine defines a specific set of characters to be "word" characters. Any character not in that set is considered a word break. This set of characters is fairly limited: it consists solely of the Roman alphabet in both upper- and lower-case, decimal digits, and the underscore character. Accented characters, such as "é" or "ü" are, unfortunately, treated as word breaks.

IreneStr:
@afercia How I understand it, the fact that regexes in javascript have a limited list of word characters, only matters when using wildcards like `.` and `\b`. When I remove `replaceDiacritics()` it does match (for example) 'habré', which is on the list, but not 'habre', which is not on the list. I'll investigate a little further how to break no other regexes.

IreneStr:
@afercia Mystery explained: como, cómo, cuando and cuándo are all on the Spanish list 🙃

IreneStr:
@afercia It turns out this ^ is not the full explanation. 
- 'mi' in the slug triggers both matches for 'mi' and 'mí'. Result: The slug for this page contains stop words, consider removing them. ❌ 
- 'mí' in the slug only triggers a match for 'mí'. Result: The slug for this page contains a stop word, consider removing it. ✅

afercia:
To my understanding, the main issue here is about how regexes work in JavaScript 🙂 
> Accented characters, such as "é" or "ü" are, unfortunately, treated as word breaks.

`createWordRegex.js` uses a regex and is then used by `stopWordsInText.js`. If the string passed to the regex contains a diacritic, for example `adónde` JavaScript will treat the `ó` as a word break so the actual string used to match stop words will be `ad nde` (I guess actually without the space but treated as 2 separated words).

IreneStr:
Closed, because we decided in the recalibration meeting that the StopwordsInUrl and StopWordsInKeyword assessments will be removed soon.

PR #1335, opened 2017-12-13
Updated copy and added shortlink to the proper article

### LINKED ISSUE #1334: Change copy for previouslyUsedKeywords
Part of the current copy for the `previouslyUsedKeywords` references an article on cornerstone content. This needs to be altered to specifically point to an article on why reusing keywords might not be a great idea. This also means that copy referencing the cornerstone content needs to be replaced with the following, new one:

`It’s probably a good idea to read %4$sthis post%5$s about why you should not use your focus keyword more than once.`

jcomack:
## Summary

* Updates the copy for the `previouslyUsedKeywords` assessment by referring to a new blogpost on why it might be a bad idea to use the same keyword more than once.

## Test instructions

* Link this PR to your local Yoast SEO install by running `yarn link` and `yarn link yoastseo` in your Yoast SEO directory.
* Make sure to build the JS files by running `grunt build:js`.
* Create a post with a particular keyword.
* Create another post and use the same keyword as before. You shouldn't be seeing the new copy, but a different message regarding reusing keywords.
* Repeat the previous step once more and see that the `previouslyUsedKeyword` assessment now displays the new copy.
* Please verify that the link works as well.

Fixes #1334


IreneStr:
CR 👍

manuelaugustin:
Acceptance test done 👍

PR #1345, opened 2017-12-19
Throw error when passive voice percentage is NaN

### LINKED ISSUE #1344: Assessments containing Nan% are not considered errors
Assessments with unintended results should be considered errors.

## What happened instead?
It's considered a consideration.
schermafbeelding 2017-12-13 om 14 57 57

Pass something to the passive voice assessment that is NaN. We should investigate whether this incorrect classification happens for all assessments, or only the passive voice assessment.

## Technical info

* Platform: WordPress
* Platform version: develop
* Yoast SEO version: develop

Xyfi:
## Summary

* When passive voice percentage returns `NaN` (Not a number), the assessment throws an error.

* 

## Test instructions

* Change line 24 in `passiveVoiceAssessment` to `percentage = NaN`.
* Open the example app and enter any sentence (Can be either passive or active).
* The assessment should show up as an error.

Fixes #1344


src/assessments/readability/passiveVoiceAssessment.js
```diff
@@ -22,6 +22,11 @@ var calculatePassiveVoiceResult = function( passiveVoice, i18n ) {
 
 	var percentage = ( passiveVoice.passives.length / passiveVoice.total ) * 100;
 	percentage = formatNumber( percentage );
+
+	if( isNaN( percentage ) ) {
```
afercia: Minor: missing space after if.

afercia:
I think the JS should be built and committed?

afercia:
CR 👍 
Pushed a minor CS and built the JS.

IreneStr:
@Xyfi 
- This is not a problem that is specific to the passive voice assessment. Please fix this for all readability assessment that use percentages or numbers (which most likely means all of them). I think it would be a good idea to fix this in the same central way as the other types of errors (line 183-196 in `assessor.js`)
- Please also add unittests.
- Please uncommit `examples/browserified/example-browserified.map`

afercia:
Seems this is a bit more complicated than expected. One of the reasons it may happen is when `0` gets divided by `0`, which returns `NaN`. TL;DR the tokenization may fail for some reasons (weird content?) and the passive voice assessment doesn't calculate the percentage in a safe way as other assessments do.

afercia:
As discussed, this will avoid the division by zero error but there are also other things to check. They should be addressed in separate issues though, so I think this is good to go, for now. Will open separate issues.

Xyfi:
I've discussed with @afercia to merge this PR as is after acceptance testing. Then we should move forward to implementing custom errors in the tokenizer, so assessments won't be ran if the arguments are not valid.

@IreneStr Do you agree?

Xyfi:
Acceptance :+1:

PR #1347, opened 2017-12-19
Prevent redundant refreshing

moorscode:
## Summary

Please note that this PR does not resolve any duplicate render calls on the `Results.render` method as should have been described in the original issue.

* Reduce the number of refreshing the analysis on page load. 

* Only refresh the analysis when all YoastSEO.js plugins have been loaded.
* Always use the debounced version of the refresh function instead of using the runAnalyzer on plugins loaded.

## Test instructions

* Load a page or post and see that the analysis is being done as expected

Fixes https://github.com/Yoast/wordpress-seo/issues/8399

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -492,6 +492,11 @@ App.prototype.getData = function() {
  * @returns {void}
  */
 App.prototype.refresh = function() {
+	// Until all plugins are loaded, do not trigger a refresh.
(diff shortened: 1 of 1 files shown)


afercia:
CR done 👍

Xyfi:
Acceptance done :+1:

PR #1354, opened 2017-12-28
Added protocol check to determine whether link is relative

Xyfi:
## Summary

Relative URL's now count as internal links in the link assessment.

* Whether a link is relative is determined by extracting the protocol from the URL with a new helper-function in `stringProcessing/url.js` called `getProtocol`.
* Relative fragment URLs (`#header`) are not considered internal links because they don't navigate to another page.
* The `getProtocol` returns the protocol including the colon (`http:`). This is because Node's `url.parse` method returns the protocol like that.

## Test instructions

* Add a relative URL to a new post.
* The assessment that says you should use internal links no longer is present. 

Fixes Yoast/wordpress-seo#7083

### DIFF
--- a/spec/stringProcessing/linkTypeSpec.js
+++ b/spec/stringProcessing/linkTypeSpec.js
@@ -1,14 +1,52 @@
 var linkTypeFunction = require( "../../js/stringProcessing/getLinkType.js" );
 
-describe( "checks type of link", function(){
+describe( "getLinkType", function(){
+
 	it( "should classify relative links as internal", function() {
-		expect( linkTypeFunction( "<a href='http://example.org/'>Internal</a>", "http://example.org" ) ).toBe( "internal" );
-		expect( linkTypeFunction( "<a href='http://example.org/some-path?query_args#hash'>Internal</a>", "http://example.org" ) ).toBe( "internal" );
+		expect( linkTypeFunction( "<a href='another-path'>Internal</a>", "http://example.org" ) ).toBe( "internal" );
+	} );
+
+	it( "should classify relative links with query params as internal", function() {
+		expect( linkTypeFunction( "<a href='another-path?test=123&another=abc'>Internal</a>", "http://example.org" ) ).toBe( "internal" );
+	} );
+
+	it( "should classify relative links with empty query params as internal", function() {
+		expect( linkTypeFunction( "<a href='another-path?test=&another='>Internal</a>", "http://example.org" ) ).toBe( "internal" );
+	} );
+
+	it( "should classify relative links with fragments as internal", function() {
+		expect( linkTypeFunction( "<a href='another-path#subheader'>Internal</a>", "http://example.org" ) ).toBe( "internal" );
+	} );
+
+	it( "should classify absolute links without host as internal", function() {
+		expect( linkTypeFunction( "<a href='/some-path'>Internal</a>", "http://example.org" ) ).toBe( "internal" );
 	});
 
-	it( "returns linktype", function(){
-		expect( linkTypeFunction( "<a href='http://yoast.com'>A link</a>", "http://yoast.com" ) ).toBe( "internal" );
-		expect( linkTypeFunction( "<a href='http://yoast.com'>A link</a>", "http://example.com" ) ).toBe( "external" );
-		expect( linkTypeFunction( "<a href='ftp://yoast.com'>A link</a>", "http://yoast.com" ) ).toBe( "other" );
+	it( "should classify absolute links without host with query params as internal", function() {
+		expect( linkTypeFunction( "<a href='/some-path?test=123'>Internal</a>", "http://example.org" ) ).toBe( "internal" );
+	});
+
+	it( "should classify absolute links without host with a fragment as internal", function() {
+		expect( linkTypeFunction( "<a href='/some-path#subheader'>Internal</a>", "http://example.org" ) ).toBe( "internal" );
+	});
+
+	it( "should classify absolute links without host with query params and fragment as internal", function() {
+		expect( linkTypeFunction( "<a href='/some-path?test=123#subheader'>Internal</a>", "http://example.org" ) ).toBe( "internal" );
+	});
+
+	it( "should classify absolute links as internal", function() {
+		expect( linkTypeFunction( "<a href='http://yoast.com'>Internal</a>", "http://yoast.com" ) ).toBe( "internal" );
+	});
+
+	it( "should classify absolute links with a different host as external", function() {
+		expect( linkTypeFunction( "<a href='http://yoast.com'>Internal</a>", "http://example.com" ) ).toBe( "external" );
+	});
+
+	it( "should classify absolute links with ftp protocol as other", function() {
+		expect( linkTypeFunction( "<a href='ftp://yoast.com'>Internal</a>", "http://yoast.com" ) ).toBe( "other" );
+	});
+
+	it( "should classify absolute relative fragment URLs as other", function() {
+		expect( linkTypeFunction( "<a href='#subheader'>Internal</a>", "http://yoast.com" ) ).toBe( "other" );
 	});
 });
--- a/spec/stringProcessing/linkTypeSpec.js
+++ b/spec/stringProcessing/linkTypeSpec.js
@@ -6,6 +6,22 @@ describe( "getLinkType", function(){
 		expect( linkTypeFunction( "<a href='another-path'>Internal</a>", "http://example.org" ) ).toBe( "internal" );
 	} );
 
+	it( "should classify relative links with double-dot notation as internal", function () {
+		expect( linkTypeFunction( "<a href='../another-path'>Internal</a>", "http://example.org" ) ).toBe( "internal" );
+	} );
+
+	it( "should classify relative links with single-dot notation as internal", function () {
+		expect( linkTypeFunction( "<a href='./another-path'>Internal</a>", "http://example.org" ) ).toBe( "internal" );
+	} );
+
+	it( "should classify relative links with multiple directories as internal", function() {
+		expect( linkTypeFunction( "<a href='another-path/directory'>Internal</a>", "http://example.org" ) ).toBe( "internal" );
+	} );
+
+	it( "should classify relative links with double-dot notation and multiple directories as internal", function() {
+		expect( linkTypeFunction( "<a href='../another-path/directory'>Internal</a>", "http://example.org" ) ).toBe( "internal" );
+	} );
+
 	it( "should classify relative links with query params as internal", function() {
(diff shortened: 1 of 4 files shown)


spec/stringProcessing/urlSpec.js
```diff
@@ -141,4 +141,23 @@ describe( "A URL helper", function() {
 			expect( actual ).toBe( expected );
 		});
 	});
+
+	describe( "getProtocol", function() {
+		it( "should return the protocol of the URL", function() {
+			const urlA = "http://ww.google.nl";
```
afercia: minor: missing w

spec/stringProcessing/urlSpec.js
```diff
@@ -141,4 +141,23 @@ describe( "A URL helper", function() {
...
+			const expected = "http://";
+
+			const actual = url.getProtocol( urlA );
+
+			expect( actual ).toBe( expected );
+		} );
+
+		it( "should return null when passing a relative URL", function() {
+			const urlA = "/relative/url";
```
afercia: minor: terminology: this is not a relative URL, actually it's an absolute URL with implicit domain name

src/stringProcessing/getLinkType.js
```diff
@@ -11,17 +11,20 @@ var urlHelper = require( "./url" );
...
-	var linkType = "other";
+	let linkType = "other";
 
-	var anchorUrl = urlHelper.getFromAnchorTag( text );
+	const anchorUrl = urlHelper.getFromAnchorTag( text );
+	const protocol = urlHelper.getProtocol( anchorUrl );
 
 	// Matches all links that start with http:// and https://, case insensitive and global
-	if ( anchorUrl.match( /https?:\/\//ig ) !== null ) {
+	if ( protocol === "http://" || protocol === "https://" ) {
```
afercia: see terminology note about "protocol"

src/stringProcessing/getLinkType.js
```diff
@@ -11,17 +11,20 @@ var urlHelper = require( "./url" );
...
 		linkType = "external";
 
 		if ( urlHelper.getHostname( anchorUrl ) === urlHelper.getHostname( url ) ) {
 			linkType = "internal";
 		}
+	} else if ( protocol === null && anchorUrl.charAt( 0 ) !== "#" ) {
+		linkType = "internal";
 	}
```
afercia: worth nothing the logic of these conditions fails in case of implicit protocol links like `//www.yoast.com` which are identified as "other" while they actually are internal or external.

src/stringProcessing/url.js
```diff
@@ -81,6 +81,21 @@ function getHostname( url ) {
 	return url.hostname;
 }
 
+/**
+ * Returns the protocol of a URL.
```
afercia: see terminology note about "protocol"

src/stringProcessing/url.js
```diff
@@ -81,6 +81,21 @@ function getHostname( url ) {
...
+ *
+ * @param {string} url The URL to retrieve the protocol of.
+ * @returns {string|null} The protocol of the URL.
+ */
+function getProtocol( url ) {
+	const protocolMatcher = /^(?:[a-z]+:)?\/\//i;
+	const protocolMatch = protocolMatcher.exec( url );
+	if ( ! protocolMatch ) {
+		return null;
+	}
```
afercia: this translates to "if null, then return null": works but maybe coud be a bit more clear?

afercia:
CR 🚧 

Terminology notes:

"relative vs absolute"
URLs without the protocol part or without the domain name are still "absolute" if they don't depend on the context, for example:
`/somepath/somefile`
`//mysite.com/somepath/somefile`
are absolute URLs

Only URLs that depend on context, i.e. the location of the document they're used into, are "relative" for example:
`somepath/somefile`
`../somefile`


"protocol":
technically, the protocol doesn't include the colon and double slash. I'd expect a function named `getProtocol()` to return just, for example, `http` or `https` and not `http://` or `https://`. See also all the other references to "protocol" in the docs and variable names. Maybe the function and all other occurrences should be renamed for better clarity.

src/stringProcessing/url.js
Xyfi: Protocol now refers to the first part of the URL, before the `//`, which includes the colon.

spec/stringProcessing/urlSpec.js
```diff
@@ -141,4 +141,113 @@ describe( "A URL helper", function() {
...
+			const urlA = "test/abc";
+			const host = "www.google.nl";
+			const expected = true;
+
+			const actual = url.isInternalLink( urlA, host );
+
+			expect( actual ).toBe( expected );
+		} );
+
+		it( "should identify a link as external with host", function() {
```
IreneStr: Not really descriptive

src/stringProcessing/url.js
```diff
@@ -81,6 +81,72 @@ function getHostname( url ) {
...
+function protocolIsHttpScheme( protocol ) {
+	if ( ! protocol ) {
+		return false;
+	}
+
+	return ( protocol === "http:" || protocol === "https:" );
+}
+
+/**
+ * Determines whther the link is a relative fragment URL.
```
IreneStr: missing 'e'

src/stringProcessing/url.js
```diff
@@ -81,6 +81,72 @@ function getHostname( url ) {
 	return url.hostname;
 }
 
+/**
+ * Returns the protocol of a URL.
+ *
+ * Note that the colon (http:) is also part of the protocol.
```
IreneStr: Please comment why

src/stringProcessing/getLinkType.js
```diff
@@ -11,18 +11,21 @@ var urlHelper = require( "./url" );
...
 
-	var anchorUrl = urlHelper.getFromAnchorTag( text );
-
-	// Matches all links that start with http:// and https://, case insensitive and global
-	if ( anchorUrl.match( /https?:\/\//ig ) !== null ) {
-		linkType = "external";
+	/**
+	 * A link is "Other" if:
+	 * - The protocol is neither null, nor http, nor https.
+	 * - The link is a relative fragment URL (starts with #).
```
IreneStr: Please comment why those are not internal

src/stringProcessing/url.js
Xyfi: I did this because Node's url.parse method returns the protocol including the colon.

IreneStr:
CR 2 done 👌

afercia:
Pushed some typos fixes.

Acceptance not ok 🚧 

Links like, for example:
`href="path/to"`
`href="/path/to"`
`href="../path/to"`
are identified as "other", while they're valid internal links.

Also, how far we want to go in checking for well formed URLs? Right now, malformed URLs are not checked. For example:
`href="http://#mylink"`
`href="http:// my link"`
Identified as "internal".

`href="http://my link"`
`href="http://mylink"`
Identified as "external".

Xyfi:
I'll write some additional tests to check for the cases you proposed. 

Should malformed URLs be identified as other, or maybe we could implement a new seo-feature that would tell the user they have malformed URLs in their post?

afercia:
> Should malformed URLs be identified as other, or maybe we could implement a new seo-feature that would tell the user they have malformed URLs in their post?

Good question and probably one that the management should make a decision about. Checking all the possibel cases of malformed URLs would be tricky though, so maybe the best option for now is to keep things simple. Worth noting, the current WordPress editor has some _basic_ form of check for malformed URLs but only when link gets added with the visual editor:

screen shot 2018-01-09 at 10 28 23

That doesn't help when users insert a link manually in text mode, and wouldn't help the analysis anyway.

Instead, as far as I know, Gutenberg doesn't have any check. 😞

Xyfi:
@afercia Maybe we can create a separate issue for this and keep moving forward with the PR as is. The internal links should be detected properly now.

spec/stringProcessing/linkTypeSpec.js
```diff
@@ -1,14 +1,52 @@
...
+
+	it( "should classify absolute links with ftp protocol as other", function() {
+		expect( linkTypeFunction( "<a href='ftp://yoast.com'>Internal</a>", "http://yoast.com" ) ).toBe( "other" );
 	});
 
-	it( "returns linktype", function(){
-		expect( linkTypeFunction( "<a href='http://yoast.com'>A link</a>", "http://yoast.com" ) ).toBe( "internal" );
-		expect( linkTypeFunction( "<a href='http://yoast.com'>A link</a>", "http://example.com" ) ).toBe( "external" );
-		expect( linkTypeFunction( "<a href='ftp://yoast.com'>A link</a>", "http://yoast.com" ) ).toBe( "other" );
+	it( "should classify absolute relative fragment URLs as other", function() {
```
IreneStr: absolute relative?

spec/stringProcessing/linkTypeSpec.js
```diff
@@ -1,14 +1,52 @@
...
+
+	it( "should classify relative links with empty query params as internal", function() {
+		expect( linkTypeFunction( "<a href='another-path?test=&another='>Internal</a>", "http://example.org" ) ).toBe( "internal" );
+	} );
+
+	it( "should classify relative links with fragments as internal", function() {
+		expect( linkTypeFunction( "<a href='another-path#subheader'>Internal</a>", "http://example.org" ) ).toBe( "internal" );
+	} );
+
+	it( "should classify absolute links without host as internal", function() {
```
IreneStr: I'm missing tests for `href='some-path'`, `href="/path/to"` and `href="../path/to"`

spec/stringProcessing/urlSpec.js
```diff
@@ -141,4 +141,113 @@ describe( "A URL helper", function() {
...
+			const host = "www.google.nl";
+			const expected = true;
+
+			const actual = url.isInternalLink( urlA, host );
+
+			expect( actual ).toBe( expected );
+		} );
+
+		it( "should identify a relative url as an internal link", function() {
+			const urlA = "test/abc";
```
IreneStr: Missing test for `href="../path/to"`

src/stringProcessing/url.js
```diff
@@ -81,6 +81,72 @@ function getHostname( url ) {
 	return url.hostname;
 }
 
+/**
+ * Returns the protocol of a URL.
+ *
+ * Note that the colon (http:) is also part of the protocol, conform to node's url.parse api.
+ *
+ * @param {string} url The URL to retrieve the protocol of.
+ * @returns {string|null} The protocol of the URL.
```
IreneStr: or null if...

spec/stringProcessing/linkTypeSpec.js
Xyfi: Added

spec/stringProcessing/urlSpec.js
Xyfi: Added

IreneStr:
Acceptance 👍

PR #1359, opened 2018-01-03
Simplify the Sub Heading Keyword assessment message text.

### LINKED ISSUE #1358: Improve the Sub Headings Keyword Assessment message
Currently, this result is listed in the "good" results but sounds more like a suggestion for an improvement:

screen shot 2018-01-03 at 11 17 11

Quoting @IreneStr from Slack:
> We’ve looked at this assessment in the recalibration project. We’ll go for another approach in the future: https://github.com/Yoast/YoastSEO.js/issues/1311. However, this change will only be in a big release somewhere next Spring. For now, @manuelaugustin and I think the way to go is to remove “Try to use it in one more subheading" from the feedback string and keep the green score (score: 9) and the ‘good’ score category.

afercia:
## Summary

* Simplified message for the Sub Headings Keyword assessment.

## Test instructions

Simply removes the part `Try to use it in at least one more subheading.` from the message.

Fixes #1358


IreneStr:
@afercia I think 'only' should be removed as well, because the new feedback string still sounds something should be improved. I'm sorry 😬

afercia:
@IreneStr done!

IreneStr:
CR and acceptance done 👍

PR #1364, opened 2018-01-03
Better "nofollow" detection

moorscode:
Even when other arguments are present in the `rel` tag

## Summary

* Fixes a bug where multiple `rel` arguments prevented correct `nofollow` detection.

* Extended the regex to have a more accepting test.
* As the HTML spec also allows for `rel=nofollow`, this is added as separate check.

## Test instructions

* See issue.

Fixes https://github.com/Yoast/wordpress-seo/issues/7035

### DIFF
--- a/src/stringProcessing/checkNofollow.js
+++ b/src/stringProcessing/checkNofollow.js
@@ -29,7 +29,7 @@ module.exports = function( text ) {
 			if ( nodeValue.rel.toLowerCase().split( /\s/ ).includes( "nofollow" ) ) {
 				linkFollow = "Nofollow";
 			}
-		}
+		},
 	} );
 
 	parser.write( text );
--- a/src/stringProcessing/checkNofollow.js
+++ b/src/stringProcessing/checkNofollow.js
@@ -2,7 +2,6 @@
 
 // We use an external library, which can be found here: https://github.com/fb55/htmlparser2.
 let htmlparser = require( "htmlparser2" );
-let isString = require( "lodash/isString" );
 
 /**
  * Checks if a links has a nofollow attribute value. If it has, returns Nofollow, otherwise Dofollow.
@@ -18,15 +17,15 @@ module.exports = function( anchorHTML ) {
 		 * Detects if there is a `nofollow` argument value in the `rel` argument of a link.
 		 *
 		 * @param {string} tagName The tag name.
-		 * @param {object} nodeValue The attribute with the keys and values of the tag.
+		 * @param {object} attributes  The attribute with the keys and values of the tag.
 		 * @returns {void}
 		 */
-		onopentag: function( tagName, nodeValue ) {
-			if ( tagName !== "a" || ! isString( nodeValue.rel ) ) {
+		onopentag: function( tagName, attributes ) {
+			if ( tagName !== "a" || ! attributes.rel ) {
 				return;
 			}
 
-			if ( nodeValue.rel.toLowerCase().split( /\s/ ).includes( "nofollow" ) ) {
+			if ( attributes.rel.toLowerCase().split( /\s/ ).includes( "nofollow" ) ) {
 				linkFollow = "Nofollow";
 			}
 		},
--- a/src/stringProcessing/checkNofollow.js
+++ b/src/stringProcessing/checkNofollow.js
@@ -5,24 +5,24 @@ let htmlparser = require( "htmlparser2" );
 let isString = require( "lodash/isString" );
 
 /**
- * Checks if a links has a nofollow attribute. If it has, returns Nofollow, otherwise Dofollow.
+ * Checks if a links has a nofollow attribute value. If it has, returns Nofollow, otherwise Dofollow.
  *
- * @param {string} text The text to check against.
+ * @param {string} anchorHTML The anchor HTML to check against.
  * @returns {string} Returns Dofollow or Nofollow.
  */
-module.exports = function( text ) {
+module.exports = function( anchorHTML ) {
 	let linkFollow = "Dofollow";
 
 	let parser = new htmlparser.Parser( {
 		/**
-		 * Detects if there is a `nofollow` argument in the `rel` attribute of a link.
+		 * Detects if there is a `nofollow` argument value in the `rel` argument of a link.
 		 *
 		 * @param {string} tagName The tag name.
-		 * @param {object} nodeValue The attribute with the keys and values of the tag.
+		 * @param {object} node The attribute with the keys and values of the tag.
 		 * @returns {void}
 		 */
 		onopentag: function( tagName, nodeValue ) {
-			if ( tagName.toLowerCase() !== "a" || ! isString( nodeValue.rel ) ) {
+			if ( tagName !== "a" || ! isString( nodeValue.rel ) ) {
 				return;
 			}
 
@@ -32,7 +32,8 @@ module.exports = function( text ) {
 		},
 	} );
 
-	parser.write( text );
+	parser.write( anchorHTML );
+	parser.end();
 
 	return linkFollow;
 };
--- a/src/stringProcessing/checkNofollow.js
+++ b/src/stringProcessing/checkNofollow.js
@@ -10,7 +10,7 @@ module.exports = function( text ) {
 	var linkFollow = "Dofollow";
 
 	// Matches all nofollow links, case insensitive and global
-	if ( text.match( /rel=([\'\"])nofollow\1/ig ) !== null ) {
+	if ( text.match( /rel=(\'|\")([^\1]+\s)?nofollow(\s[^\1]+)?\1|rel=nofollow/ig ) !== null ) {
 		linkFollow = "Nofollow";
 	}
 	return linkFollow;
--- a/src/stringProcessing/checkNofollow.js
+++ b/src/stringProcessing/checkNofollow.js
@@ -10,8 +10,9 @@ module.exports = function( text ) {
 	var linkFollow = "Dofollow";
 
 	// Matches all nofollow links, case insensitive and global
-	if ( text.match( /rel=(\'|\")([^\1]+\s)?nofollow(\s[^\1]+)?\1|rel=nofollow/ig ) !== null ) {
+	if ( text.match( /\srel=(nofollow(\s|\/>|>)|(\'|\")([^\3]+\s)?nofollow(\s[^\3]+)?\3)/ig ) !== null ) {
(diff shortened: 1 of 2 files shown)


afercia:
CR 🚧 
Definitely not my area of expertise so I may be wrong 🙂 but testing the new regex, I have a few questions. 

1
In this part: `(nofollow(\s|\/>|>)` is there any case where the opening tag of a link ends with `/>`? I guess no, that would be incorrect so it could be simplified.

2
`([^\3]*\s)?` and `(\s[^\3]*)?`
If I'm not wrong the first means "anything that is not the thing captured in the 3rd group (which is quote or double quote) followed by a space" and the last one means vice-versa. Which is smart, but if I'm not wrong there's a problem with the quantifiers because they're both "greedy". Testing a bit, the match continues for any character until the last quote (as captured previously that means until the last single quote or double quote):

screen shot 2018-01-05 at 11 52 51

and with double quote:

screen shot 2018-01-05 at 11 52 45

Making the "anything which is not a quote" lazy, could be a start. Something along these lines:
`\srel=(nofollow(\s|\/>|>)|(\'|\")([^\3]*?\s)?nofollow(\s[^\3]*?)?\3)`

But there are other possible variations to make everything "lazy" that should be evaluated. This should also help performance, especially with very long text.

3
Should a case like the following one fail? Currently, this passes: `blabla rel='nofollow'blabla` but technically it's malformed HTML.

4 
Performance: as it is know, testing on regex101 with a text of just 300 words I get the following results:
old regex: 128 steps (~0ms)
new regex: 4305 steps (~5ms)

These numbers largely depends on the text used and I'm not sure if they're a concern, but with very long text there could be some impact on performance.

src/stringProcessing/checkNofollow.js
```diff
@@ -10,12 +11,33 @@ const noFollowExpression = /\srel=(nofollow(\s|\/>|>)|(\'|\")([^\3]*\s)?nofollow
...
-	var linkFollow = "Dofollow";
+	let linkFollow = 'Dofollow';
 
-	// Matches all nofollow links, case insensitive and global
-	if ( text.match( noFollowExpression ) !== null ) {
-		linkFollow = "Nofollow";
-	}
+	let parser = new htmlparser.Parser( {
+		/**
+		 * Handles the opening tag. If the opening tag is included in the inlineTags array, set inScriptBlock to true.
```
atimmer: This is not your code doc ;)

src/stringProcessing/checkNofollow.js
```diff
@@ -1,17 +1,38 @@
 /** @module stringProcessing/checkNofollow */
 
+// We use an external library, which can be found here: https://github.com/fb55/htmlparser2.
+let htmlparser = require( "htmlparser2" );
+let isString = require( "lodash/isString" );
+
 /**
  * Checks if a links has a nofollow attribute. If it has, returns Nofollow, otherwise Dofollow.
```
afercia: typo: link, not links
typo: "nofollow attribute": "nofollow" is not an attribute, is an attribute value

src/stringProcessing/checkNofollow.js
```diff
@@ -1,17 +1,38 @@
...
  *
  * @param {string} text The text to check against.
  * @returns {string} Returns Dofollow or Nofollow.
  */
 module.exports = function( text ) {
-	var linkFollow = "Dofollow";
```
afercia: about the docblock and passed argument name: this is not exactly a generic "text". We're passing a string representing an anchor tag, for clarity I'd consider to rename it.

src/stringProcessing/checkNofollow.js
```diff
@@ -1,17 +1,38 @@
...
+	let linkFollow = "Dofollow";
+
+	let parser = new htmlparser.Parser( {
+		/**
+		 * Detects if there is a `nofollow` argument in the `rel` attribute of a link.
```
afercia: nofollow is not an argument, is an attribute value

src/stringProcessing/checkNofollow.js
```diff
@@ -1,17 +1,38 @@
...
+		 *
+		 * @param {string} tagName The tag name.
+		 * @param {object} nodeValue The attribute with the keys and values of the tag.
```
afercia: I guess this is copied from `src/stringProcessing/htmlParser.js` however, this object actually contains _all_ the attributes found in a tag. I'd consider to rename `nodeValue` to something more meaningful and improve hte description.

src/stringProcessing/checkNofollow.js
```diff
@@ -1,17 +1,38 @@
...
+		 * @returns {void}
+		 */
+		onopentag: function( tagName, nodeValue ) {
```
afercia: see above, consider to rename nodeValue

src/stringProcessing/checkNofollow.js
```diff
@@ -1,17 +1,38 @@
...
+			if ( tagName.toLowerCase() !== "a" || ! isString( nodeValue.rel ) ) {
```
afercia: the passed anchor text already has tag names and attributes names lower cased because the parser is already used for the text passed to the Paper, and it already lower cases tag names, see `src/stringProcessing/htmlParser.js`. Regardless, the parser parses again the link here and by default it returns lower case tag names. See the docs:
https://github.com/fb55/htmlparser2/wiki/Parser-options

Option: xmlMode
Default: false.

Option: lowerCaseTags
If set to true, all tags will be lowercased. **If xmlMode is disabled, this defaults to true**.

src/stringProcessing/checkNofollow.js
```diff
@@ -1,17 +1,38 @@
...
+				return;
+			}
+
+			if ( nodeValue.rel.toLowerCase().split( /\s/ ).includes( "nofollow" ) ) {
+				linkFollow = "Nofollow";
+			}
+		},
+	} );
+
+	parser.write( text );
```
afercia: Not sure if we should also use `parser.end();`, as shown in all the examples in the htmlparser2 docs.

afercia:
CR done 🚧 

Apart from the CR, worth noting there's at least one edge case of malformed HTML where the parser fails:

screen shot 2018-01-10 at 10 31 54

I'd say it's a minor edge case not worth fixing, just wanted to mention it as something we should be aware of.

atimmer:
@afercia I was looking into changing the parser to https://github.com/inikulin/parse5. That parser matches what browsers do. If we change to that we stop having to worry about malformed HTML.

For this specific malformed HTML, I think we can safely ignore it.

src/stringProcessing/checkNofollow.js
moorscode: It actually contains all attributes in this case as well, though we are only interested in the `rel` attribute in processing.

src/stringProcessing/checkNofollow.js
moorscode: If it's not a string (or empty) there is no way `nofollow` would be present in the argument, which is the intended behaviour of this specific implementation.

The alternative is to do the following check: `! ( "rel" in nodeValue )`
Which is much harder to read in my opinion.
And would be spread out over multiple lines per coding standards.

Alternatively `_isUndefined ` could be used from Lodash, but that is also more general and does not remove the need for an import.
I don't think we need to optimize for use of imports as these should be generally available, and if used somewhere else in the code it is justified to use them "everywhere". Could you confirm @atimmer

src/stringProcessing/checkNofollow.js
Xyfi: I would use the same variable name as the documentation of `htmlparser2` does, which is `attribs` ( or `attrs` or `attributes` ).

src/stringProcessing/checkNofollow.js
```diff
@@ -1,17 +1,39 @@
...
+module.exports = function( anchorHTML ) {
+	let linkFollow = "Dofollow";
+
+	let parser = new htmlparser.Parser( {
+		/**
+		 * Detects if there is a `nofollow` argument value in the `rel` argument of a link.
+		 *
+		 * @param {string} tagName The tag name.
+		 * @param {object} nodeValue The attribute with the keys and values of the tag.
+		 * @returns {void}
```
Xyfi: Doesn't return `void`

src/stringProcessing/checkNofollow.js
```diff
@@ -1,17 +1,39 @@
...
+		 */
+		onopentag: function( tagName, nodeValue ) {
+			if ( tagName !== "a" || ! isString( nodeValue.rel ) ) {
```
Xyfi: I know you are trying to be explicit, but I think `tagName !== "a" || ! nodeValue.rel` should suffice to reduce the number of operations needed. HTML Attributes are always strings as far as I know.

src/stringProcessing/checkNofollow.js
moorscode: As far as I can see it does.

afercia:
Aside: I actually don't know what `parser.end();` really does 🙂 Just seen it's always used in their examples. Shold it be addes also in `src/stringProcessing/htmlParser.js`?

Xyfi:
CR :+1:

src/stringProcessing/checkNofollow.js
atimmer: `isString` is fine here, as this ensures that member methods are available to use on the string. It prevents errors such as 'calling method on undefined'.

atimmer:
@afercia It closes the parser stream. It is good practice to use it when you're done.

afercia:
Acceptance not OK 🚧 

Seems to me the htmlparser "fixes" unbalanced quotes in the HTML attributes so there's at least one test that formally passes but actually returns an unexpected result. See the test `should ignore malformed rel tag`: it passes, but if you paste in the browserified example the string:
`string <a href='http://example.com' rel="nofollow'>link</a>`

it returns "Nofollow". Should be ignored instead.

I guess if we want to still check for malformed quotes, we should test it with a regex _before_ the htmlparser kicks in.

moorscode:
I have tested using the package suggested by @atimmer, and these give the same result in unit tests.

As switching packages should be done in a more general sense, instead of having two different tools for the same job, I don't think it's the best choice to implement it in this PR.

We can leave it as it is for now (suggested by @atimmer) as want to follow how the parsers work.

afercia:
Definitely not my area of expertise, just noticed that `parse5` seems way bigger in size.

Xyfi:
Acceptance :+1:

PR #1373, opened 2018-01-09
Add locale to Paper and remove duplicate analyses

manuelaugustin:
Added locale to the Paper in the performance analysis and removed duplicate analyses.


IreneStr:
CR and acceptance 👍

PR #1391, opened 2018-01-24
Changed snippet preview to show mobile preview by default

Xyfi:
## Summary

Snippet preview now shows mobile preview by default.

## Test instructions

* Link YoastSEO.js to `wordpress-seo` using `yarn link` and build the project.
* Check various pages that contain the snippet preview and verify mobile is selected by default.

Fixes Yoast/wordpress-seo#7533


afercia:
Looking better at this, seems this is not enough to set the initial state to "mobile". As implemented in #954 the initial state depends on the width of the snippet preview compared to 640 pixels. If the available space is equal to or greater than 640 pixels, it's always set to "desktop".

afercia:
A few main issues.
The width calculations were previously implemented in a way to override the default config. Removed anything related, since we don't need that any longer. The initial mode is now really based on `defaults.previewMode`

Instead, when clicking on "Desktop" there's now the need to calculate the widths and, if there's not enough space available for the desktop preview, we need to activate the horizontal scrolling and make the visual "scroll hint" appear. Done.

The `yoast-section` is now the one from yoast-components and its styling sets all its content to `box-sizing: border-box`. This changed the behavior of the preview. Refactored a bit trying to simplify but worth noting this UI is still fragile and any change in the CSS layout, paddings, etc. can break things. Would be nice to make it in a more solid way in the future and avoid "magic numbers".

On Desktop, the Google snippet max-width is 600 pixels. In our preview we need a 20 pixels additional space on the left to display the "arrows" when hovering/focusing the preview. We don't need padding on the right. This makes the total width 620 pixels. The CSS computed width and the JS minimum width must always be the same: 620 pixels. 

To recap:

- removes initial widths calculation: the default view is now always "mobile"
- when "Desktop" gets clicked, there's the need to check the widths and set the "scroll hint" accordingly
- checks the width of the preview "view" element instead of its container, as the "view" element is already available where we need it and we don't need to get it again
- 620 pixels instead of 640: we need padding only on the left
- removes the `this.element.container` property, no more used
- updated docs for the `this.element` object
- removes doc reference for `element.measurers.titleWidth` that was removed here https://github.com/Yoast/YoastSEO.js/commit/c9089623e4c732fc9d600caf4cb79446324f9b0b
- updates the standalone example to box-sizing border model

To test:
- verify the mobile view is always the default one
- entry real content in the title, url, and meta description
- test both on the standalone example and on the plugin (yarn  link and build the assets)
- switch to desktop
- shrink your browser window at various sizes
- on WP, try also to collapse/expand the admin menu
- verify the desktop snippet preview becomes "scrollable" and the hint appears when the available space ie less then 620 pixels
- verify the "arrows" on the left are visible when hovering the preview or focusing the snippet editor field (see screenshot below)

screen shot 2018-01-26 at 18 47 31

manuelaugustin:
Acceptance done 👍

PR #1392, opened 2018-01-25
Fix transition words marking by adding marks into the sentence

igorschoester:
## Summary

A highlighted transition word sentence is no longer highlighted completely, but only the transition words themselves in that sentence; As received by the researcher.

### Notable

The researcher also checks for certain two part transition words. If those are found it returns the transition words with the words in between those two parts included.

Example sentence: "The best result does not only mean the best answer, but it also means the best experience.".
The new returned highlight is: " not only mean the best answer, but ". Where before the whole sentence was highlighted.

* The mark in the current state does not keep track of where it is found in the text. This is why the whole sentence is used in a mark. And there can be multiple marks in that sentence.
* The original casing needs to be preserved, this is the reason for the search with leftover in the wrapStringWithRegex file.

## Test instructions

* Link YoastSEO.js to wordpress-seo using yarn link and build the project.
* Edit a few posts and in the Yoast SEO readability checks, highlight the transition words. Check if the transition words are correctly highlighted, preserving the case.

Fixes Yoast/wordpress-seo#6525

### DIFF
--- a/src/markers/addMarksInSentence.js
+++ b/src/markers/addMarksInSentence.js
@@ -1,0 +1,25 @@
+const forEach = require( "lodash/forEach" );
+
+const wrapStringWithRegex = require( "../stringProcessing/wrapStringWithRegex" );
+const Mark = require( "../values/Mark" );
+
+/**
+ * Marks multiple texts within a sentence with HTML tags
+ *
+ * @param {string}        sentence The unmarked text.
+ * @param {Array<string>} marks    Text to search for and mark.
+ *
+ * @returns {Mark} A new mark containing the original sentence and one marked with HTML tags.
+ */
+module.exports = function( sentence, marks ) {
+	let markedSentence = sentence;
+
+	forEach( marks, function( mark ) {
+		markedSentence = wrapStringWithRegex( markedSentence, mark, "<yoastmark class='yoast-text-mark'>", "</yoastmark>" );
+	} );
+
+	return new Mark( {
+		original: sentence,
+		marked: markedSentence,
+	} );
+};
--- a/src/markers/addMarksInSentence.js
+++ b/src/markers/addMarksInSentence.js
@@ -4,7 +4,7 @@ const wrapStringWithRegex = require( "../stringProcessing/wrapStringWithRegex" )
 const Mark = require( "../values/Mark" );
 
 /**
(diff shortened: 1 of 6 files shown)


src/markers/addMarksInSentence.js
```diff
@@ -0,0 +1,25 @@
+const forEach = require( "lodash/forEach" );
+
+const wrapStringWithRegex = require( "../stringProcessing/wrapStringWithRegex" );
+const Mark = require( "../values/Mark" );
+
+/**
+ * Marks multiple texts within a sentence with HTML tags
```
manuelaugustin: "multiple text parts"?

src/stringProcessing/wrapStringWithRegex.js
```diff
@@ -0,0 +1,26 @@
+/** @module stringProcessing/wrapStringWithRegex */
+
+/**
+ * Replaces all occurrences in a string and wrap them.
```
manuelaugustin: wraps

src/stringProcessing/wrapStringWithRegex.js
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * @param {string} text        The text to look through.
+ * @param {string} search      A string to use as regex.
+ * @param {string} wrapPrefix  Insert this before the found text.
+ * @param {string} wrapSuffix  Insert this after the found text.
+ *
+ * @returns {string} The text with all the found string wrapped.
```
manuelaugustin: strings

manuelaugustin:
CR done.

I think in general the new markers work great! I just had some little comments concerning the documentation (see the in-line comments), and one point concerning test coverage: the branch coverage of wrapStringWithRegexSpec.js is low. For this you can add a test where you don’t provide parameters for wrapPrefix and wrapSuffix, so that the default values are also tested.

As discussed, there’s one bigger change that still should be implemented: the way two-part transition words are marked should be changed. Instead of marking the whole string that’s returned by the researcher, only the transition words themselves should be marked. So for the sentence “It’s both big and small.” we should mark “both” and “and” as separate words.

manuelaugustin:
Closed as it's not relevant anymore with the introduction of the tree parser, which is currently in progress.

PR #1427, opened 2018-03-01
Change boundaries of outbound link assessment

### LINKED ISSUE #1301: Change boundaries of outbound links assessment
The only thing changing in this assessment is that an orange bullet should become a red bullet.

manuelaugustin:
## Summary

This PR can be summarized in the following changelog entry: When there are no outbound links in the text, the feedback bullet is red instead of orange.

## Test instructions

* Check out the branch.
* Use the browserified example. Don't forget to run a `grunt build:js`.
* Add a text without an outbound link.
* Make sure that you get a feedback string about the missing outbound link with a red bullet.

Fixes #1301


afercia:
CR 👍

PR #1483, opened 2018-05-04
Fix some no-useless-escapes warnings

### LINKED ISSUE #1481: Fix the ~200 eslint warnings
We should fix the ~200 eslint warnings in spec and src folders (mostly concern superfluous escapes in regexes). These warnings are simply annoying and make it more time-consuming to find actual problems that fail builds.

nataliashitova:
## Summary

* n/a

* A lot of superfluous escapes in regexes were fixed to get rid of warnings at `grunt check` and builds.
* The rest of the warnings do not seem to be legit, because removing escapes changes regex processing.

## Test instructions

* Run `yarn test` in the branch to make sure these escapes were indeed superfluous and that processing works well without them.

Fixes #1481

### DIFF
--- a/spec/assessments/introductionKeywordSpec.js
+++ b/spec/assessments/introductionKeywordSpec.js
@@ -23,6 +23,6 @@ describe( "An assessment for finding the keyword in the first paragraph", functi
 		var assessment = firstParagraphAssessment.getResult( paper, Factory.buildMockResearcher( 0 ), i18n );
 
 		expect( assessment.getScore() ).toBe( 3 );
-		expect( assessment.getText() ).toBe( "The focus keyword doesn\'t appear in the first paragraph of the copy. Make sure the topic is clear immediately." );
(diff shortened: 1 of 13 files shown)


atimmer:
As soon as you fix some warnings you should adjust the value of `maxWarnings` for ESLint found here: https://github.com/Yoast/YoastSEO.js/blob/d7baef0d498dd54d3da8199de9293d9b98c10cec/grunt/config/eslint.js#L6

manuelaugustin:
@atimmer As far as I can see the `maxWarnings` option has to do with the `eslint:target` warnings, not the `eslint:tests` warnings. In this branch, only `eslint:tests` warnings were fixed. So in this case we don't need to change the `maxWarnings`, right?

atimmer:
That's true.

manuelaugustin:
CR and acceptance done 👍
Additionally checked some of the regexes/feedback strings using the beta script.

PR #1338, opened 2017-12-14
Change upper boundary of meta description length to 320

manuelaugustin:
## Summary

This PR can be summarized in the following changelog entry: The upper boundary of the meta description length has been changed to 320 characters.

## Test instructions

This PR can be tested by following these steps:

* Link this PR to your local Yoast SEO install by running `yarn link`.
* Make sure to build the JS files by running grunt `build:js`.
* Create a post with a meta description with a length of 120 - 320 characters. Make sure you get `good` as feedback.
* Create a post with a meta description with more than 320 characters. Make sure you get `improvement` as feedback.

Fixes #1337

PR #1497, opened 2018-05-17
Changed maximum meta description length

### LINKED ISSUE #1496: Change upper boundary of meta description
We should change the upper boundary of the meta description length to 156 characters in order to reflect current changes in how meta descriptions are displayed in the SRP.

Xyfi:
## Summary

* Maximum meta description length has been changed from `320` to `156`.

*

## Test instructions

* Create a string of 156 characters.
Example:
`This is what the page might look like on a Google search result page. This is what the page might look like on a Google search result page. This is what the`
* Test if the length is still perceived as `good`.
* Add some characters. Check if the following SEO assessment is given: `The meta description is over 156 characters. Reducing the length will ensure the entire description will be visible.`

Fixes #1496

### DIFF
--- a/src/assessments/seo/metaDescriptionLengthAssessment.js
+++ b/src/assessments/seo/metaDescriptionLengthAssessment.js
@@ -2,7 +2,9 @@ let AssessmentResult = require( "../../values/AssessmentResult.js" );
 let Assessment = require( "../../assessment.js" );
 let merge = require( "lodash/merge" );
 
-const maximumLength = 320;
+import Config from "../../config/config";
+
+const MAXIMUM_META_DESCRIPTION_LENGTH = Config.maxMeta;
 
 /**
  * Assessment for calculating the length of the meta description.
@@ -20,7 +22,7 @@ class MetaDescriptionLengthAssessment extends Assessment {
 
 		let defaultConfig = {
 			recommendedMaximumLength: 120,
-			maximumLength: maximumLength,
+			maximumLength: MAXIMUM_META_DESCRIPTION_LENGTH,
 			scores: {
 				noMetaDescription: 1,
 				tooLong: 6,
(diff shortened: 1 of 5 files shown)


src/assessments/seo/metaDescriptionLengthAssessment.js
```diff
@@ -2,7 +2,9 @@ let AssessmentResult = require( "../../values/AssessmentResult.js" );
 let Assessment = require( "../../assessment.js" );
 let merge = require( "lodash/merge" );
 
-const maximumLength = 320;
+import Config from "../../config/config";
+
+const MAXIMUM_META_DESCRIPTION_LENGTH = Config.maxMeta;
```
manuelaugustin: Is there a specific reason for this new naming convention for variables? (all caps with underscores instead of camelcase)

manuelaugustin:
CR done 🚧

Small question about naming conventions, see comment.

src/assessments/seo/metaDescriptionLengthAssessment.js
Xyfi: We don't really have a convention for that, and since the value was stored in two differently named variables in `metaDescriptionLengthAssessment.js` and `snippetPreview.js` I used all caps because of personal preference.

manuelaugustin: Ah, but now the value is just imported from the config. So I think it would make sense to stick with the general naming convention in YoastSEO.js and use camel case.

manuelaugustin:
Acceptance 👎

* The bar isn't updating correctly yet, see screenshot. Refer to https://github.com/Yoast/YoastSEO.js/pull/1338/files to see all places wher the meta description value needs to be changed. (To make sure all references to the old value have been changed, you can just do a `cmd + f` through `snippetPreview.js` and search for `320`)

screen_shot_2018-05-18_at_10_58_32

igorschoester:
Acceptance 👍

PR #1504, opened 2018-05-23
Add morphological analysis for English nouns

### LINKED ISSUE #1501: Build a prototype of the morphological regex-based analysis for English nouns
We need to try implementing morphological regex-based analysis for English nouns. 

Larger context is available at #1500.

Some similar projects:
https://github.com/spencermountain/compromise
https://github.com/abdulwasae/English-Word-Inflections-and-POS
https://www.nodebox.net/code/index.php/Linguistics

### LINKED ISSUE #1514: Implement a general module for keyword analysis

nataliashitova:
## Summary

* Implements morphological analysis for keyword nouns in English. 

* For every English word build its possible singular and plural noun form using regexes and a list of exceptions. 
* The keyword filled in the focus keyword field is now checked for presence in the title in all its noun forms.

## Test instructions

* Add singular-plural forms of nouns to `/spec/morphology/englishNounsSpec.js` and check if `yarn test` passes
* Use browserified to check titleKeyword assessment. 
 * Single-word or multiple-word keyword: Try using different forms of nouns in the keyword field and in the title. Check if the assessment returns correct result.
* For multiple-word keyword, try using only function words inside keyword - the assessment should effectively search for them in the title. Try using both function and content words inside keyword - the assessment should only search for content words.
 * Try putting the keyword inside double quotation marks - the assessment should only search for exact matches in the title.

Fixes #1501. Fixes #1514


src/researches/findKeywordInPageTitle.js
```diff
@@ -13,12 +13,28 @@ var escapeRegExp = require( "lodash/escapeRegExp" );
...
-	var title = paper.getTitle();
-	var keyword = escapeRegExp( paper.getKeyword() );
-	var locale = paper.getLocale();
-	var result = { matches: 0, position: -1 };
-	result.matches = wordMatch( title, keyword, locale );
-	result.position = title.toLocaleLowerCase().indexOf( keyword );
+	const title = paper.getTitle();
+	// const locale = paper.getLocale();
+
+	const keywordForms = buildKeywordForms( paper );
```
manuelaugustin: This makes sense for this prototype, but for a release version, we need to make sure we don't build the forms separately each time we use them in an assessment. Ideally, the forms should only be built when the keyword is entered the first time or when it's altered.

src/researches/findKeywordInPageTitle.js
```diff
@@ -13,12 +13,28 @@ var escapeRegExp = require( "lodash/escapeRegExp" );
...
 
 module.exports = function( paper ) {
-	var title = paper.getTitle();
-	var keyword = escapeRegExp( paper.getKeyword() );
-	var locale = paper.getLocale();
-	var result = { matches: 0, position: -1 };
-	result.matches = wordMatch( title, keyword, locale );
-	result.position = title.toLocaleLowerCase().indexOf( keyword );
+	const title = paper.getTitle();
+	// const locale = paper.getLocale();
```
manuelaugustin: Even for the prototype, I suggest you make sure that the morphological forms are only built when the locale is English. Currently it unnecessarily builds forms for other languages as well (according to the English pattern). I think throwing out the basic functionality for the other languages now will make it more difficult to restore it when we actually want to release the morphological analysis for English.

src/researches/buildKeywordForms.js
```diff
@@ -0,0 +1,66 @@
...
+ * @returns {array} Array of all forms to be searched for keyword-based assessments.
+ */
+
+module.exports = function( paper ) {
+	let keyword = paper.getKeyword();
+
+	// If the keyword is embedded in double quotation marks no morphological analysis is required.
+	// Return keyword itself, without outer-most quotation marks.
+	if ( keyword[ 0 ] === "\"" && keyword[ keyword.length - 1 ] === "\"" ) {
+		console.log( "Requested exact match, returning keyword itself.", keyword.substring( 1, keyword.length - 1 ) );
```
manuelaugustin: I'd remove the console.logs for the PR.

src/researches/buildKeywordForms.js
```diff
@@ -0,0 +1,66 @@
...
+		return [].concat( keyword.substring( 1, keyword.length - 1 ) );
+	}
+
+	let forms = [];
+
+	// If the keyword is a single word return all its possible forms.
+	if ( keyword.indexOf( " " ) === -1 ) {
+		console.log( "Keyword is one word, returning  all forms of this keyword. ", getForms( escapeRegExp( keyword ) ) );
```
manuelaugustin: See above.

src/researches/findKeywordInPageTitle.js
```diff
@@ -13,12 +13,28 @@ var escapeRegExp = require( "lodash/escapeRegExp" );
...
+	const titleLowerCase = title.toLocaleLowerCase();
+
+	keywordForms.forEach( function( form ) {
+		const formToLowerCase = form.toLocaleLowerCase();
+		const keywordFormIndex = titleLowerCase.indexOf( formToLowerCase );
+
+		if ( keywordFormIndex > -1 ) {
+			positions = positions.concat( keywordFormIndex );
+		}
+		}
```
manuelaugustin: Should have an indentation of one tab less (see `grunt check`)

spec/researches/findKeywordInPageitleSpec.js
```diff
@@ -29,7 +29,7 @@ describe( "Match keywords in string", function() {
 
 		var mockPaper = new Paper( "", { keyword: "focus keyword", title: "focus-keyword" } );
 		result = pageTitleKeyword( mockPaper );
-		expect( result.matches ).toBe( 0 );
+		expect( result.matches ).toBe( 2 );
```
manuelaugustin: This behavior is interesting. First, there's fact that the hyphen is ignored - we should think about whether that is desirable. Second, there now are two matches, I suppose one for "focus" and one for "keyword". I don't think this is particularly problematic for the `pageInTitle` assessment, since here we're not interested in the number of matches. However, this sort of behavior might cause trouble in assessments where the number of matches is important. So I'd start thinking about how this can be avoided already at this step.

src/morphology/english/irregularNouns.js
```diff
@@ -0,0 +1,296 @@
...
+	[ "palatonaris", "palatonares" ],
+	[ "pelvis", "pelves" ],
+	[ "penis", "penes" ],
+	[ "pileipellis", "pileipelles" ],
+	[ "postnaris", "postnares" ],
+	[ "prepubis", "prepubes" ],
+	[ "pseudopenis", "pseudopenes" ],
+	[ "pubis", "pubes" ],
+	[ "sternochondroscapularis", "sternochondroscapulares" ],
+	[ "tax", "taxi", "taxis", "taxes" ],
```
manuelaugustin: Why is `tax` and `taxi` seen as on lemma? 🤔

src/researches/buildKeywordForms.js
```diff
@@ -0,0 +1,66 @@
...
+ * @param {Object} paper The paper to analyze keyword forms for.
+ *
+ * @returns {array} Array of all forms to be searched for keyword-based assessments.
+ */
+
+module.exports = function( paper ) {
+	let keyword = paper.getKeyword();
+
+	// If the keyword is embedded in double quotation marks no morphological analysis is required.
+	// Return keyword itself, without outer-most quotation marks.
```
manuelaugustin: Format as multi-line comment.

src/morphology/english/getNounForms.js
```diff
@@ -0,0 +1,88 @@
...
+
+
+const getNounForms = function( word ) {
+	let forms = [];
+	let isIrregular = false;
+
+	const irregular = checkIrregulars( word );
+	if ( ! isUndefined( irregular ) ) {
+		forms = forms.concat( irregular );
+		isIrregular = true;
```
manuelaugustin: Instead of setting `isIrregular` to true and then checking whether it's true, I think you can just declare the `return unique( forms );` statement at this position.

src/morphology/english/getNounForms.js
```diff
@@ -0,0 +1,88 @@
+// "use strict";
+const irregularNouns = require( "./irregularNouns.js" );
+const singularizeRegex = require( "./regexNoun.js" ).singularizeRegex;
+const pluralizeRegex = require( "./regexNoun.js" ).pluralizeRegex;
+const hispanicRegex = require( "./regexNoun.js" ).hispanicRegex;
+
+const isUndefined = require( "lodash/isUndefined.js" );
+const unique = require( "lodash/uniqBy" );
+
+const checkIrregulars = function( word ) {
```
manuelaugustin: Please add documentation to this function and the other functions below.

src/researches/buildKeywordForms.js
```diff
@@ -0,0 +1,66 @@
...
+	if ( keyword[ 0 ] === "\"" && keyword[ keyword.length - 1 ] === "\"" ) {
```
manuelaugustin: Concerning the exact match functionality (i.e., using quotation marks): I think it makes sense to use quotation marks for this in the prototype, but when we release this as a feature there should probably be a different interface solution. This could, for example, be a checkbox that users can check to switch on exact matching functionality.

src/researches/buildKeywordForms.js
```diff
@@ -0,0 +1,66 @@
...
+const getWords = require( "../stringProcessing/getWords.js" );
+const getLanguage = require( "../helpers/getLanguage.js" );
+const getFunctionWords = require( "../helpers/getFunctionWords.js" )();
+
+const includes = require( "lodash/includes" );
+const filter = require( "lodash/filter" );
+const escapeRegExp = require( "lodash/escapeRegExp" );
+
+/**
+ * Analyses the focus keyword string. Checks if morphology is requested or if the user wants to match exact string.
```
manuelaugustin: Analyzes

src/researches/buildKeywordForms.js
```diff
@@ -0,0 +1,66 @@
...
+ * If morphology is required the module builds all wordforms for all content words (prepositions, articles, conjunctions).
+ *
+ * @param {Object} paper The paper to analyze keyword forms for.
+ *
+ * @returns {array} Array of all forms to be searched for keyword-based assessments.
```
manuelaugustin: {Array}

src/researches/buildKeywordForms.js
```diff
@@ -0,0 +1,66 @@
...
+const getLanguage = require( "../helpers/getLanguage.js" );
+const getFunctionWords = require( "../helpers/getFunctionWords.js" )();
+
+const includes = require( "lodash/includes" );
+const filter = require( "lodash/filter" );
+const escapeRegExp = require( "lodash/escapeRegExp" );
+
+/**
+ * Analyses the focus keyword string. Checks if morphology is requested or if the user wants to match exact string.
+ * If morphology is required the module builds all wordforms for all content words (prepositions, articles, conjunctions).
```
manuelaugustin: word forms

manuelaugustin:
CR 🏗

First of all, it's really great that we finally have functionality to create singular/plural forms of keywords! 😄

* For specific feedback points see the comments. Some of the points are about changes that should best be implemented at the current stage (i.e., as it usually goes with a code review). Other comments are more about considerations for the long run. These are things that should be addressed before the release, but not necessarily at this stage of the prototype.
* I saw that `getNounForms.js` isn't fully covered by unit tests yet.
* Functionality-wise, I think there still need to be a few tweaks. Creating the forms works great, but matching is still a bit of an issue. It works fine with single keywords, but not with multi-word keyphrases. Neither the word order, nor the completeness of the keyphrase are taken into account. That's because we just check whether any of the forms in the array matches. My keyphrase can be `Donald Duck`, but it will also match on `Donald Trump`, simply because the array contains `Donald`. I think once we have all the forms, we actually need to build a regex for matching. This regex should take word order into account, but provide different options for the forms (and if possible, it should make function words optional).
* I'd be curious to get some idea of the performance of the new functionality. To get a rough idea, could you run a performance analysis on trunk and one on this branch, and compare the difference for the `findKeywordInPageTitle` assessment?

nataliashitova:
Also add possessives

src/researches/findKeywordInPageTitle.js
nataliashitova: Agreed.

src/researches/findKeywordInPageTitle.js
nataliashitova: Fixed.

src/researches/buildKeywordForms.js
nataliashitova: Since the development is in progress, I commented them. I will remove them at a later stage of development.

src/researches/findKeywordInPageTitle.js
nataliashitova: Fixed.

spec/researches/findKeywordInPageitleSpec.js
nataliashitova: First part: I just replicated the behavior of Google.
screen shot 2018-06-08 at 11 15 21
Second part: The problem is that we really do not know how Google treats multi-word matches. I believe that longer (multi-word) matches get higher weight in Google than shorter (e.g., single-word) matches. But without quotation marks both partial and full matches are definitely considered. Maybe we can rather think about making our assessments somewhat smarter in this respect.

src/morphology/english/getForms.js
```diff
@@ -0,0 +1,12 @@
+const getNounForms = require( "./getNounForms.js" ).getNounForms;
+const unique = require( "lodash/uniqBy" );
```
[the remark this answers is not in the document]
nataliashitova: Agreed

src/morphology/english/irregularNouns.js
nataliashitova: Actually, there are three lemmas there: tax - taxes, taxi - taxis, taxis - taxes. Because they overlap in forms it is convenient to see them as one lemma (and Google allows such inaccuracies itself). Technically we can accurately predict word forms for "tax" and for "taxi", but not for the other four cases.

src/researches/buildKeywordForms.js
nataliashitova: Fixed

src/morphology/english/getNounForms.js
nataliashitova: Good point.

src/morphology/english/getNounForms.js
nataliashitova: Done.

src/researches/buildKeywordForms.js
nataliashitova: I think either way is fine. Quotation marks is how you search on Google, so I think it is valid, too.

src/researches/buildKeywordForms.js
nataliashitova: ok

src/researches/buildKeywordForms.js
nataliashitova: ok

nataliashitova:
@manuelaugustin 
> For specific feedback points see the comments. Some of the points are about changes that should best be implemented at the current stage (i.e., as it usually goes with a code review). Other comments are more about considerations for the long run. These are things that should be addressed before the release, but not necessarily at this stage of the prototype.

Agreed. I think for many things it will make sense to wait for verb forms to appear as well.

> I saw that `getNounForms.js` isn't fully covered by unit tests yet.

It is covered now.

> Functionality-wise, I think there still need to be a few tweaks. Creating the forms works great, but matching is still a bit of an issue. It works fine with single keywords, but not with multi-word keyphrases. Neither the word order, nor the completeness of the keyphrase are taken into account. That's because we just check whether any of the forms in the array matches. My keyphrase can be Donald Duck, but it will also match on Donald Trump, simply because the array contains Donald. I think once we have all the forms, we actually need to build a regex for matching. This regex should take word order into account, but provide different options for the forms (and if possible, it should make function words optional).

I think it will be good to get more insight into how Google matches multi-word forms. I would think into direction of differentiating full and partial matches in our reports. As you see, Google does both, so only matching strings like `any-form-of-word1` `any-form-of-word2` `any-form-of-word3` would not be accurate.
screen shot 2018-06-08 at 12 13 28

> I'd be curious to get some idea of the performance of the new functionality. To get a rough idea, could you run a performance analysis on trunk and one on this branch, and compare the difference for the findKeywordInPageTitle assessment?

I ran the performance analysis, but I doubt if the results are representative, because keyword is not specified for the texts we have.

||time|percentage|average|min|max|
|---------|----------|---------------|-------|-------|--------|
|MORPHO||||||
|Text 1|4|0.087|0.4|0|3| 
|Text 2|1|0.011|0.1|0|1|
|Text 3|1|0.006|1|1|1|
|DEVELOP||||||
|Text 1|4|0.086|0.4|0|3|
|Text 2|1|0.011|0.1|0|1|
|Text 3|0|0|0|0|0|

nataliashitova:
I believe that we need to discuss how to adjust our keyword-based assessments to the new possibilities, when it comes to multi-word keyphrases. 

**Current implementation.** Consider a multi-word keyphrase as a collection of independent words, for which we get all possible forms and search for them in the text/title/description, ignoring order and whether they appear alone or together with the other words from the keyphrase. Function words are filtered out from the search.
> Keyphrase: Dogs and cats
> A `dog` and a `cat`. `Dogs` are pretty awesome, `cats` are not so much. (4 matches)

This option tries to mimic Google's behavior, but not fully, since it does not account for partial/full matches. The current way assessment scores are computed will not work this well, because they will overestimate number of matches.

From here on there are three degrees of freedom: differential weighting of full/partial matches, distance estimation and ordering. 

**Differential weighting** is something that Google definitely does - an article containing "Dogs and cats" will rank for "dog cat" higher than an article that only has "dog". This factor is rather easy to implement: For the assessments we can return feedback that would differentiate between full and partial matches and calculate the assessment score giving much higher weight to full matches than to partial matches. Whether or not we are doing differential weighting, keyphrase components that appear together (e.g., only separated by punctuation marks and function words) should be considered as one match.

**Distance estimation** is also something that Google does, but we do not know to what extent it matters for the rank - an article containing "Dogs and cats" will rank for "dog cat" higher than an article that has "Dogs many-many-many-many-words cats". It feels to me that it might be computationally challenging to properly implement this factor for long keyphrases.

Whether **ordering** is considered on top of distance estimation is hard to tell. I could not find a good example. Will an article containing "Dogs and cats" rank for "dog cat" higher than an article that has "Cats and dogs"?

manuelaugustin:
CR & acceptance test done 👍

PR #1526, opened 2018-06-07
Add verb inflection for English

### LINKED ISSUE #1514: Implement a general module for keyword analysis

### LINKED ISSUE #1516: Add verb inflection for English
For proper morphological analysis of keywords we need to add verb inflection.
(1) s-form - already implemented in the pluralize for nouns (#1501). Double-check if this rule generates correct forms for verbs as well.
(2) ing-form
(3) past-tense form (-ed or irregular)
(4) past-participle form (-ed or irregular)

nataliashitova:
## Summary

* Implements morphological analysis for keyword verbs in English.

* For every English verb build its possible verb forms using regexes and a list of exceptions.

## Test instructions

* Add verb forms of nouns to `/spec/morphology/englishVerbSpec.js` and check if `yarn test` passes
* More proper testing will be possible after #1501 is merged.

Fixes #1516 Fixes #1514


src/morphology/english/getVerbForms.js
```diff
@@ -0,0 +1,217 @@
...
+const ingFormToInfinitiveRegex = require( "./regexVerb.js" ).ingFormToInfinitive;
+const edFormToInfinitiveRegex = require( "./regexVerb.js" ).edFormToInfinitive;
+const infinitiveToSFormRegex = require( "./regexVerb.js" ).infinitiveToSForm;
+const infinitiveToIngFormRegex = require( "./regexVerb.js" ).infinitiveToIngForm;
+const infinitiveToEdFormRegex = require( "./regexVerb.js" ).infinitiveToEdForm;
+
+const isUndefined = require( "lodash/isUndefined.js" );
+const unique = require( "lodash/uniqBy" );
+
+const normalizePrefixed = function( word ) {
```
manuelaugustin: Could you add documentation to all functions in this file?

src/morphology/english/getVerbForms.js
```diff
@@ -0,0 +1,217 @@
...
+
+const infinitiveToEdForm = function( word ) {
+	return buildVerbFormFromRegex( word, infinitiveToEdFormRegex );
+};
+
+const getInfinitive = function( word ) {
+	let infinitive = word;
+	let guessedForm = "inf";
+
+	if ( endsWithS( word ) ) {
```
manuelaugustin: Say that `endsWithS` returns true and the infinitive/guessedForm are set accordingly, the function would still continue to check whether `endsWithIng` and `endsWithEd` are also true. But that's not really necessary because only one of these conditions can hold true, right? That's why I think it would be better to do an early return right when an if-condition evaluates to `true`.

manuelaugustin:
CR 🏗

Looks great! There's a lot of complex regex fun going on! Who'd have thought that English with its meager verbal morphology already requires so much work. But it's so great that we can build all of those forms now! 😄

Regarding the code review, I just have some minor remarks on code style and documentation, see comments.

src/morphology/english/getVerbForms.js
nataliashitova: Fixed.

src/morphology/english/getVerbForms.js
nataliashitova: Fixed.

src/morphology/english/getVerbForms.js
```diff
@@ -0,0 +1,217 @@
...
+	};
+};
+
+const getVerbForms = function( word ) {
+	let forms = [];
+
+	const irregular = checkIrregulars( word );
+	if ( ! isUndefined( irregular ) ) {
+		forms = forms.concat( irregular );
+		return unique( forms );
```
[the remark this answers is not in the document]
nataliashitova: Fixed.

nataliashitova:
I added more specs for combined noun+verb form building and improved verb regexes a little.

manuelaugustin:
CR & acceptance done 👍

PR #1527, opened 2018-06-08
Improve image assessment by checking how often the keyword is included in alt-tags

### LINKED ISSUE #1388: Penalize image anchor text over-optimization
Moved from https://github.com/Yoast/wordpress-seo/issues/7924#issuecomment-359184385

Yoast SEO looks for image anchor text and gives us a green bullet when there is 1 image that uses the focus keyword as an anchor text.

The problem is that there is no limit on how many images with that same anchor text you can use. This means that I could over-optimize my post easily by adding 10 images with the same anchor text without losing my green bullet.

Since a lot of images/screenshots are used in today's web, over-optimizing anchor text is something we need to prevent.

manuelaugustin:
## Summary

*  Adds a specific range for how often the keyword should be included in image alt tags.

* When there less than 5 images, including an alt tag with the keyword at least once will always result in a green bullet. There is no upper limit. When there are exactly five images, 2-4 alt tags with keywords will result in a green bullet. This has been hard-coded since it's on the border between the less strict and the stricter condition. When there are 6 or more keywords, the number of alt tags with the keyword should be between 30% & 75% of all images.

## Test instructions

* Make a beta script with `feature/recalibration` as the branch for Yoast SEO Free, and this branch for YoastSEO.js.
* Follow the steps below and make sure you always get appropriate feedback after each step. When everything is alright, you should get a green bullet, whereas you'll get an orange bullet when something needs to be changed. The only condition where you'll get a red bullet is when there are no images at all.
* Add a new post.
* Add one image.
	* Add an alt tag with a description.
	* Remove the alt tag again.
* Add a keyword.
	* Add an alt tag that doesn't contain the keyword.
	* Add an alt tag that does contain the keyword.
* Add more images (up to 4).
	* Vary the number of images with alt tags that contain the keyword. As long as there is at least one image alt tag that contains the keyword, you should get a green bullet.
* Add exactly five images to your post. The bullet should be green as long as 2-4 of them contain the keyword.
* Add more than five images to the post. Make sure you get appropriate feedback when adding/removing alt tags with the keyword: between 30% & 75% should result in a green bullet, going higher or lower than that in an orange bullet.
	

Fixes #1388

### DIFF
--- a/spec/assessments/textImagesSpec.js
+++ b/spec/assessments/textImagesSpec.js
@@ -1,24 +1,24 @@
-var ImageCountAssessment = require( "../../js/assessments/seo/textImagesAssessment" );
-var Paper = require( "../../js/values/Paper.js" );
-var Factory = require( "../helpers/factory.js" );
-var i18n = Factory.buildJed();
+const ImageCountAssessment = require( "../../js/assessments/seo/textImagesAssessment" );
+const Paper = require( "../../js/values/Paper.js" );
+const Factory = require( "../helpers/factory.js" );
+const i18n = Factory.buildJed();
 
 let imageCountAssessment = new ImageCountAssessment();
 
 describe( "An image count assessment", function() {
 	it( "assesses no images", function() {
-		var mockPaper = new Paper( "sample" );
+		const mockPaper = new Paper( "sample" );
 
-		var assessment = imageCountAssessment.getResult( mockPaper, Factory.buildMockResearcher( 0 ), i18n );
+		const assessment = imageCountAssessment.getResult( mockPaper, Factory.buildMockResearcher( 0 ), i18n );
 
 		expect( assessment.getScore() ).toEqual( 3 );
 		expect( assessment.getText() ).toEqual( "No images appear in this page, consider adding some as appropriate." );
 	} );
 
 	it( "assesses a single image, without a keyword and alt-tag set", function() {
-		var mockPaper = new Paper( "These are just five words <img src='image.jpg' />" );
+		const mockPaper = new Paper( "These are just five words <img src='image.jpg' />" );
 
-		var assessment = imageCountAssessment.getResult( mockPaper, Factory.buildMockResearcher( {
+		const assessment = imageCountAssessment.getResult( mockPaper, Factory.buildMockResearcher( {
 			noAlt: 1,
 			withAlt: 0,
 			withAltKeyword: 0,
@@ -30,9 +30,9 @@ describe( "An image count assessment", function() {
 	} );
 
 	it( "assesses a single image, without a keyword, but with an alt-tag set", function() {
-		var mockPaper = new Paper( "These are just five words <img src='image.jpg' alt='image' />" );
+		const mockPaper = new Paper( "These are just five words <img src='image.jpg' alt='image' />" );
 
-		var assessment = imageCountAssessment.getResult( mockPaper, Factory.buildMockResearcher( {
+		const assessment = imageCountAssessment.getResult( mockPaper, Factory.buildMockResearcher( {
 			noAlt: 0,
 			withAlt: 1,
 			withAltKeyword: 0,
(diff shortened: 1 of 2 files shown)


src/assessments/seo/textImagesAssessment.js
```diff
@@ -80,19 +125,20 @@ class TextImagesAssessment extends Assessment {
...
+		// Has alt-tags, but no keyword is set.
+		if ( this.altProperties.withAlt > 0 ) {
 			return {
-				score: this._config.scores.withAltKeyword,
+				score: this._config.scores.withAlt,
 				resultText: i18n.dgettext(
 					"js-text-analysis",
-					"The images on this page contain alt attributes with the focus keyword."
+					"The images on this page contain alt attributes. " +
+					"Once you've set a focus keyword, also include the keyword where appropriate."
```
nataliashitova: The feedback string is not clear about where to include the keyword. Maybe "Once you've set a focus keyword, don't forget to include it in alt attributes, where appropriate".

src/assessments/seo/textImagesAssessment.js
```diff
@@ -102,16 +148,55 @@ class TextImagesAssessment extends Assessment {
...
+		if ( this.hasTooFewMatches() ) {
 			return {
-				score: this._config.scores.withAlt,
-				resultText: i18n.dgettext(
+				score: this._config.scores.withAltTooFewKeywordMatches,
+				resultText: i18n.sprintf( i18n.dgettext(
 					"js-text-analysis",
-					"The images on this page contain alt attributes."
+					"Only in %1$d out of %2$d images on this page contain alt attributes with the focus keyword. " +
+					"Where appropriate, try to include the focus keyword in more alt attributes."
```
nataliashitova: Please add comments for translators

src/assessments/seo/textImagesAssessment.js
```diff
@@ -102,16 +148,55 @@ class TextImagesAssessment extends Assessment {
...
+		 * This check needs to be made before the check for too many matches because of the special rule for
+		 * exactly 5 matches.
+		 */
+		if ( this.hasGoodNumberOfMatches() ) {
+			return {
+				score: this._config.scores.withAltGoodNumberOfKeywordMatches,
+				resultText: i18n.sprintf( i18n.dngettext(
+					"js-text-analysis",
+					"The image on this page contains an alt attribute with the focus keyword.",
+					"%1$d out of %2$d images on this page contain alt attributes with the focus keyword.",
```
nataliashitova: Please add comments for translators

src/assessments/seo/textImagesAssessment.js
```diff
@@ -102,16 +148,55 @@ class TextImagesAssessment extends Assessment {
...
+			};
+		}
+
+		if ( this.hasTooManyMatches() ) {
+			return {
+				score: this._config.scores.withAltTooFewKeywordMatches,
+				resultText: i18n.sprintf( i18n.dgettext(
+					"js-text-analysis",
+					"%1$d out of %2$d images on this page contain alt attributes with the focus keyword. " +
+					"That's a bit much. Only include the focus keyword when it really fits the image."
```
nataliashitova: Please add comments for translators

src/assessments/seo/textImagesAssessment.js
manuelaugustin: Fixed

nataliashitova:
Also the title of this PR is not vague (looks like default). Could you make it a little more straight-forward?

src/assessments/seo/textImagesAssessment.js
manuelaugustin: Oh yes, translator comments. Oops 🤦‍♀️

src/assessments/seo/textImagesAssessment.js
manuelaugustin: Fixed

manuelaugustin:
Ah yes, good point! Done.

src/assessments/seo/textImagesAssessment.js
nataliashitova: As discussed, please change the translator comment for %2$d from "the number of images that don't contain an alt attribute with the keyword" to "the total number of images".

src/assessments/seo/textImagesAssessment.js
nataliashitova: As discussed, please change the translator comment for %2$d from "the number of images that don't contain an alt attribute with the keyword" to "the total number of images".

src/assessments/seo/textImagesAssessment.js
nataliashitova: As discussed, please change the translator comment for %2$d from "the number of images that don't contain an alt attribute with the keyword" to "the total number of images".

src/assessments/seo/textImagesAssessment.js
```diff
@@ -1,9 +1,10 @@
 let AssessmentResult = require( "../../values/AssessmentResult.js" );
 let Assessment = require( "../../assessment.js" );
+const inRangeStartEndInclusive = require( "../../helpers/inRange.js" ).inRangeStartEndInclusive;
 let merge = require( "lodash/merge" );
 
 /**
- * Represents the assessment that will look if the images have alt-tags and checks if the keyword is present in one of them.
+ * Represents the assessment that checks whether images have alt-tags and and whether the alt tags contain the keyword.
```
nataliashitova: and and

src/assessments/seo/textImagesAssessment.js
```diff
@@ -43,6 +50,8 @@ class TextImagesAssessment extends Assessment {
 		let assessmentResult = new AssessmentResult();
 		this.imageCount = researcher.getResearch( "imageCount" );
 		this.altProperties = researcher.getResearch( "altTagCount" );
```
nataliashitova: The test coverage for this researcher (imageAltTags) is 92%. Please check if you can add specs.

src/assessments/seo/textImagesAssessment.js
manuelaugustin: Fixed

src/assessments/seo/textImagesAssessment.js
```diff
@@ -102,16 +148,64 @@ class TextImagesAssessment extends Assessment {
...
+						"%1$d out of %2$d images on this page contain alt attributes with the focus keyword. " +
+						"That's a bit much. Only include the focus keyword when it really fits the image."
+					),
+					this.altProperties.withAltKeyword,
+					this.imageCount
 				),
 			};
 		}
+
+		// Images, but no alt tags.
```
nataliashitova: This condition is triggered if there are 4 or less images with empty alt-tags.
screen shot 2018-06-15 at 09 43 34
However, if there are 5 or more images with empty alt-tags, an earlier condition is triggered by `hasTooFewMatches()`, hence the feedback is different.
screen shot 2018-06-15 at 09 43 05
In my eyes the latter looks really bad.

Please consider adding a check to `hasTooFewMatches` this.altProperties.withAltKeyword>0. Or suggest an alternative solution.

nataliashitova:
I think the PR description (Relevant technical choices) might be misleading. At least to me `less than 4 images` sounds like 4 not included. However, a case of 4 images is treated exactly as 1, 2 and 3. Please consider changing the PR description.

src/assessments/seo/textImagesAssessment.js
nataliashitova: Even when this `hasTooFewMatches()` is triggered correctly (e.g., for cases when there are >4 images and alt-tags are not empty but none of them has a keyword), the feedback string "Only (in - _hope you remove it!_) 0 out of 8 images on this page contain alt attributes with the focus keyword." sounds awkward. Could you please consider adding a separate string for those zero-cases? Something like "None of the 8 images on this page...".

src/assessments/seo/textImagesAssessment.js
```diff
@@ -102,16 +148,64 @@ class TextImagesAssessment extends Assessment {
...
-				resultText: i18n.dgettext(
-					"js-text-analysis",
-					"The images on this page contain alt attributes."
+				score: this._config.scores.withAltTooFewKeywordMatches,
+				resultText: i18n.sprintf(
+					/* Translators: %1$d expands to the number of images containing an alt attribute with the keyword,
+					 * %2$d expands to the total number of images. */
+					i18n.dgettext(
+						"js-text-analysis",
+						"Only in %1$d out of %2$d images on this page contain alt attributes with the focus keyword. " +
```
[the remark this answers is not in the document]
manuelaugustin: Fixed

src/assessments/seo/textImagesAssessment.js
manuelaugustin: Ah yes, that wasn't working the way it was supposed to. I've added `this.altProperties.withAltKeyword>0` to the `hasTooFewMatches()` condition.

When there are >4 images and alt-tags are not empty but none of them has a keyword, the feedback is `The images on this page do not have alt attributes containing the focus keyword.`, so there should not be a problem with zero-cases.

manuelaugustin:
The PR description was indeed incorrent. I changed it to `less than 5`.

spec/assessments/textImagesSpec.js
```diff
@@ -1,106 +1,256 @@
...
-			withAltKeyword: 0,
-			withAltNonKeyword: 0,
-		} ), i18n );
+		const assessment = imageCountAssessment.getResult( mockPaper, Factory.buildMockResearcher( {
+			altTagCount: {
+				noAlt: 1,
+				withAlt: 0,
+				withAltKeyword: 0,
+				withAltNonKeyword: 0,
+		},
```
nataliashitova: Please add a tab.

manuelaugustin:
The keyword behavior you describe is apparently also the same in other keyword assessments. Since it's not specific for the image alt tag assessment I think it's not necessary to address it in this PR. However, I think we should keep this in mind so I've made an issue for this https://github.com/Yoast/YoastSEO.js/issues/1554

nataliashitova:
CR and acceptance 👍

Issue #1399, opened 2018-02-02
Use normalize quotes for keyword in title and keyword in first paragraph assessment


IreneStr:
In https://github.com/Yoast/YoastSEO.js/pull/909 we've introduced the use of normalizeQuotes in the keyword density assessment. However, the quote problem still exists for the keyword in title and keyword in first paragraph assessments

<img width="620" alt="schermafbeelding_2018-02-02_om_09_25_33" src="https://user-images.githubusercontent.com/17744553/35724635-b79ab22e-07fe-11e8-9b47-12a0a688c76e.png">

<img width="854" alt="schermafbeelding 2018-02-02 om 09 25 04" src="https://user-images.githubusercontent.com/17744553/35724700-fbadd374-07fe-11e8-8026-b3859976b0cc.png">

Text: ```The brain’s reward system. So far, you’ve got physical relaxation, which diminishes stress, reduced judgment, allowing you to talk and behave however you want, and stimulation of the
reward system, which makes you feel like something nice is about to happen. But the fourth neurotransmitter tops the bill: opioids. Sometimes called endorphins or internal opiates, they get released by alcohol too. Everyone knows that opiates feel good, but did you know that you can get your opiates legally by downing a stiff drink? The American martini – which consists of three ounces of gin and little else – feels particularly nice for a very simple reason. The faster the alcohol goes in, the more internal opiates get released. Hence the aaaaahhhhh.  ```
Keyword: ```brain's```
SEO title: ```The brain’s rewards system explained```

PR #1533, opened 2018-06-12
Add markings to keywordDensityAssessment

### LINKED ISSUE #1531: Add (temporary) markings to keyphrase occurrences in the text
In order to facilitate development of the morphological analysis module we need to add (temporary) markings to the keyword occurrences in the text. These markings should be used for keywordDensityAssessment and introductionKeywordAssessment.

The decision whether the markings should remain after the development is done can be taken later.

nataliashitova:
## Summary

* Adds markings on keyword forms in the text to the assessment that considers density of keyword in the text.

* Markings are added to all words within the keyphrase separately. More context on it in #1504.

## Test instructions

* Use browserified to check keywordDensityAssessment assessment.
* Single-word or multiple-word keyword: Try using different forms of nouns in the keyword field and in the text. Check if the assessment returns correct result and highlights all relevant forms.

Fixes #1531


spec/researches/buildKeywordFormsSpec.js
```diff
@@ -1,4 +1,4 @@
-const buildKeywordForms = require( "../../js/researches/buildKeywordForms.js" );
+const buildKeywordForms = require( "../../src/researches/buildKeywordForms.js" );
```
manuelaugustin: I think we usually use the build files in the js folder for the specs.

src/researches/keywordCount.js
```diff
@@ -12,8 +11,12 @@ const escapeRegExp = require( "lodash/escapeRegExp" );
  * @returns {number} The keyword count.
  */
 module.exports = function( paper ) {
-	const keyword = escapeRegExp( normalizeQuotes( paper.getKeyword() ) );
+	const keywordForms = buildKeywordForms( paper );
```
manuelaugustin: `normalizeQuotes`  was removed for the keyword, but it is still used for the text. So now when you have a keyword with non-normalized quotes, e.g., `brain’s`,  it won't be found in the text because the normalized version of that would be `brain's`.

I was also looking at this issue: https://github.com/Yoast/YoastSEO.js/issues/1399 The idea is that we should always normalize the quotes for the assessments. I wonder whether there is a more central location where we can take care of it, rather than in each individual assessment. We should have a look at this.

spec/researches/buildKeywordFormsSpec.js
nataliashitova: Fixed.

src/researches/keywordCount.js
nataliashitova: Thanks for pointing it out! I took care of these cases in my other PR #1536 that processes possessive forms.

PR #1536, opened 2018-06-13
Add possessive inflection to the morphological analysis

### LINKED ISSUE #1514: Implement a general module for keyword analysis

### LINKED ISSUE #1535: Add morphological analysis of possessive forms of nouns
Make sure forms like `boy's` and `boys'` are correctly brought back to the base form (`boy` and `boys` respectively) as well as generated from from those. Right now it is not the case.

nataliashitova:
## Summary

* Adds analysis of possessive nouns (e.g., "boy's" in "the boy's car") so that noun forms can be correctly built. 
* Prohibits building of verb forms from possessive forms (to prevent cases like "boy'sing").

* This PR also takes care of normalized/non-normalized apostrophes to be correctly recognized in the text.
* More relevant specs were added

## Test instructions

* Use browserified to check titleKeyword assessment.
* Try using different forms of nouns (singular, plural, singular possessive, plural possessive) in the keyword field and in the title. Check if the assessment returns correct result.
* Check if normalized and non-normalized apostrophes are parsed correctly.

Fixes #1535 Fixes #1514


src/researches/buildKeywordForms.js
```diff
@@ -16,22 +46,23 @@ const escapeRegExp = require( "lodash/escapeRegExp" );
...
  */
-
 module.exports = function( paper ) {
 	let keyword = paper.getKeyword();
 	const language = getLanguage( paper.getLocale() );
 	const getForms = getFormsForLanguage[ language ];
 
-	// If the language is not yet supported with respect to morphological analysis, return keyword itself.
+	// If the language is not yet supported with respect to morphological analysis, return keyword itself,
+	// including variations of the apostrophe in case it is present.
```
manuelaugustin: Format as multi-line comment.

src/researches/buildKeywordForms.js
```diff
@@ -7,6 +7,36 @@ const includes = require( "lodash/includes" );
 const filter = require( "lodash/filter" );
 const isUndefined = require( "lodash/isUndefined" );
 const escapeRegExp = require( "lodash/escapeRegExp" );
+const unique = require( "lodash/uniq" );
+const flatten = require( "lodash/flatten" );
+
+/**
+ * Checks if the input word contains a normalized or a non-normalized apostrophe.
+ * If so generates a complementary form (e.g., "il'y a" > "il`a")
```
manuelaugustin: You're using a backtick (\`) in your example, but you're transforming `'` and `’`.

src/morphology/english/regexNoun.js
```diff
@@ -88,8 +88,31 @@ const hispanic = [
 	};
 } );
 
+const possessiveToBase = [
+	[ /['’]s$/i, "" ],
+	[ /(s)(['’]s?)$/i, "$1" ],
+].map( function( a ) {
```
manuelaugustin: I think it would be good to use a more descriptive variable than `a` (`word` maybe?).

src/morphology/english/getNounForms.js
```diff
@@ -1,11 +1,62 @@
...
 const singularizeRegex = require( "./regexNoun.js" ).singularizeRegex;
 const pluralizeRegex = require( "./regexNoun.js" ).pluralizeRegex;
 const hispanicRegex = require( "./regexNoun.js" ).hispanicRegex;
 
 const isUndefined = require( "lodash/isUndefined.js" );
 const unique = require( "lodash/uniq" );
+const flatten = require( "lodash/flatten" );
+
+/**
+ * Checks if the input word is a possessive form (e.g., "boy's" in "the boy's car") and true if so.
```
manuelaugustin: `and returns true if that is the case.`

src/morphology/english/getForms.js
```diff
@@ -1,11 +1,13 @@
...
+const checkPossessive = require( "./getNounForms.js" ).checkPossessive;
 const getVerbForms = require( "./getVerbForms.js" ).getVerbForms;
 const unique = require( "lodash/uniq" );
 
 const getForms = function( word ) {
-	let forms = [].concat( getNounForms( word ), getVerbForms( word ) );
-
-	return unique( forms );
+	if ( checkPossessive( word ) ) {
+		return unique( getNounFormsWithPossessives( word ) );
```
manuelaugustin: `checkPossessive` and `getNounFormsWithPossessive` are running the same regex (`possessiveToBaseRegex`). So basically they're doing the same job, except that one returns a boolean and the other also carries out replacement. Would it be possible to change it so that `getNounFormsWithPossessive` does the checking and replaces if applicable, and does nothing if it's not applicable?  Now the same regex is running twice, which seems a bit redundant.

manuelaugustin:
CR 🏗

Nice, now we've got the most important parts of English morphology covered! 😄

For feedback see the comments.

We still need to think about normalizing quotation marks. If you use some kind of typographically funky quotation marks (e.g., `〝keyword〞`), you'll get forms such as `keyword〞s`, `keyword〞's` etc. In the old version, this was all handled by `normalizedQuotes`. We need to think about to what extent it's necessary to recognize quotation marks in keywords. For that, we probably need to know whether there are cases in which people would use them in their keywords.

src/researches/buildKeywordForms.js
nataliashitova: Fixed.

src/researches/buildKeywordForms.js
nataliashitova: I added all 5 forms (including the normalized one) to the generator.

src/morphology/english/regexNoun.js
nataliashitova: Well, it's not really a `word`, but a regex-pair, a match and a substitute. Not sure how to call them together...

src/morphology/english/getNounForms.js
nataliashitova: Fixed.

src/morphology/english/getForms.js
nataliashitova: I removed the redundancy in terms of calling the same regex twice. However, I kept the function `checkPossessive` because it is used in another function of this script (`getNounFormsWithPossesives`) and outside the script (in `getForms.js`).

nataliashitova:
@manuelaugustin 
I fixed single quotes (=apostropes) inside the possessives module and I fixed the double quotes variation around the keyphrase in the general module. Unless the keyword is supplied being enclosed in double quotation marks, all punctuation within the keyphrase is ignored by `getWords()`.

manuelaugustin:
CR 🏗

All the apostrophes worked great 👍

However, the keyword still isn't recognized when in the title it's embedded in 'funky' quotation marks, e.g., `〝`

manuelaugustin:
CR & acceptance done 👍

PR #1403, opened 2018-02-07
Implement keyword distribution assessment

manuelaugustin:
## Summary

This PR can be summarized in the following changelog entry:
* Adds an assessment that checks whether there are any parts of the text (30% of the total or more) that don't include the keyword.

## Relevant technical choices:

* The keyword distance is based on the indices of the keywords, i.e. it is character-based (rather than word-based).
* The assessment gets triggered if the text contains 200 words or more and if there are at least 2 keyword occurrences in the text.
## Test instructions

This PR can be tested by following these steps:

### Browserified
* Check out the branch.
* Use the browserified example. Don't forget to run a `grunt build:js`.
* Add a text and a keyword. Make sure the text has ≥200 words and the keyword occurs at least 2 times in the text.
* Make sure there is an appropriate feedback message in the SEO assessment. This will either be `Some parts of your text do not contain the keyword. Try to distribute the keyword more evenly.` or `Your keyword is distributed evenly throughout the text. That's great.`
* When your keyword is not evenly distributed, you can also use the marker function to mark the keyword. Check whether this works.

### WordPress
* Run the beta script, combining this branch with the `feature/recalibration` branch of `wordpress-seo`. Check the same functionality as described above.

Note: This assessment is currently still triggered when using multiple keywords with Premium. However, this won't be the case anymore once a secondary keyword assessor has been implemented (https://github.com/Yoast/YoastSEO.js/issues/1479) in order to prevent keyword stuffing.

Fixes #1314

PR #1549, opened 2018-06-19
Implement keyword distribution assessment

### LINKED ISSUE #1545: Copy the keyword distribution assessment to the feature/synonyms branch
The keyword distribution assessment that has been developed within feature/recalibration now should be made available in YoastSEO.js before the release of feature/recalibration. Therefore, the assessment and the research should be copied to the feature/synonyms branch.

manuelaugustin:
## Summary

* Adds an assessment that checks whether there are any parts of the text (30% of the total or more) that don't include the keyword.

* The keyword distance is based on the indices of the keywords, i.e. it is character-based (rather than word-based).
* The assessment gets triggered if the text contains 200 words or more and if there are at least 2 keyword occurrences in the text.

## Test instructions

See https://github.com/Yoast/YoastSEO.js/pull/1403

Fixes #1545

### DIFF
--- a/src/assessments/seo/largestKeywordDistanceAssessment.js
+++ b/src/assessments/seo/largestKeywordDistanceAssessment.js
@@ -23,8 +23,8 @@ class largestKeywordDistanceAssessment extends Assessment {
 		super();
 
 		const defaultConfig = {
-			overRecommendedMaximumKeywordDistance: 40,
-			recommendedMaximumKeywordDistance: 30,
+			overRecommendedMaximumKeywordDistance: 50,
+			recommendedMaximumKeywordDistance: 40,
 			scores: {
 				good: 9,
 				okay: 6,
@@ -53,7 +53,7 @@ class largestKeywordDistanceAssessment extends Assessment {
 
 		assessmentResult.setScore( calculatedResult.score );
 		assessmentResult.setText( calculatedResult.resultText );
-		assessmentResult.setHasMarks( calculatedResult.score < 2 );
+		assessmentResult.setHasMarks( calculatedResult.score < 9 );
 
 		return assessmentResult;
 	}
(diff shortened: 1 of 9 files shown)


src/assessments/seo/largestKeywordDistanceAssessment.js
```diff
@@ -0,0 +1,133 @@
...
+	 * @param {Object} config The configuration to use.
+	 *
+	 * @returns {void}
+	 */
+	constructor( config = {} ) {
+		super();
+
+		const defaultConfig = {
+			overRecommendedMaximumKeywordDistance: 40,
+			recommendedMaximumKeywordDistance: 30,
```
nataliashitova: I am debating with myself if lowing down these boundaries would be a smart move. On the one hand, all well-written texts I tested (a.k.a. Yoast Blog) pass this test with flying colors. On the other hand, with non-working morphology these boundaries might be a little too low for many amateur writers. Could you please give these boundaries a little more test?

src/assessments/seo/largestKeywordDistanceAssessment.js
```diff
@@ -0,0 +1,133 @@
...
+	 */
+	getResult( paper, researcher, i18n ) {
+		this._largestKeywordDistance = researcher.getResearch( "largestKeywordDistance" );
+		let assessmentResult = new AssessmentResult();
+
+		const calculatedResult = this.calculateResult( i18n );
+
+		assessmentResult.setScore( calculatedResult.score );
+		assessmentResult.setText( calculatedResult.resultText );
+		assessmentResult.setHasMarks( calculatedResult.score < 2 );
```
nataliashitova: Why not to allow markings for the OK case? I am afraid it is an artefact of the previously used scoring schema that only had good and bad.
Please also fix it in feature/recalibration.

manuelaugustin: Fixed for this PR

Concerning `feature/recalibration`: I think once we'll have merged `feature/synonyms` into `feature/development` and `feature/development` into `feature/recalibration` that will take care of it.

src/assessments/seo/largestKeywordDistanceAssessment.js
manuelaugustin: I've raised the boundaries by 10%. So now you only get a bad score if half of your text doesn't contain the keyword, and an okay score if 40% of your text doesn't contain your keyword.

nataliashitova:
CR and acceptance 👍

PR #1550, opened 2018-06-19
Parse synonyms list to array

### LINKED ISSUE #1542: Make a (provisional) keyword parser that transforms the keyword into an array
For development purposes we need to create a provisional keyword parser that would split the keyword by comma and treat the first chunk as focus keyword and the rest as an array of synonyms. This should be written with the idea in mind that in the nearest future this parser would only be used for synonyms, while the keyword will be supplied separately.

nataliashitova:
## Summary

* No user-facing changes

* Adds a function that parses a list of synonyms into an array, trimming out all non-word characters at the beginning and at the end of the element string.
* Adds a research that receives a list of synonyms (for now, within the focus keyword string), parses it and returns an object that has fields `keyword` and `synonyms`.

## Test instructions

* Try enriching the spec for the parser function and the research and check if the new tests pass.

Fixes #1542

### DIFF
--- a/spec/researches/getSynonymsSpec.js
+++ b/spec/researches/getSynonymsSpec.js
@@ -1,0 +1,39 @@
+const getSynonyms = require( "../../js/researches/getSynonyms" );
+const Paper = require( "../../js/values/Paper" );
+
+// todo: Change this spec as soon as the synonym interface is ready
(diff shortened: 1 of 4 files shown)


manuelaugustin:
CR & acceptance done 👍

PR #1553, opened 2018-06-20
Implement topic density assessment

### LINKED ISSUE #1544: Make a topicDensity research
For the new topicDensityAssessment we need a new research that matches array of synonyms with the text. Using results of recent development for feature/morphology will be very helpful here.

nataliashitova:
## Summary

* Implements a topic density assessment that considers density of not only the keyword but also synonyms of the keyword, supplied by the user.

* The keywordDensityAssessment is retained as it was.
* The keywordDensityAssessment is currently triggered all the time, but as soon as the Synonym Supply interface is ready it will only be triggered if no synonyms are supplied.
* The topicDensityAssessment has somewhat elevated boundaries compared to keywordDensityAssessment (by 0.5%), which is meant to allow those users who have already adjusted their writing style to keywordDensityAssessment to still get a green bullet with the new match functionality.

## Test instructions

* No changes have been implemented in WP-seo-premium yet, so please use browserified example.
* Paste a text and add a keyword. The keywordDensityAssessment should return meaningful result. The topicDensityAssessment should not be triggered.
* Add another word to the keyphrase, separated by a comma.
* The keywordDensityAssessment should change its result, most likely to 0% (unless you are extremely lucky and there is an exact combination of your keyword and synonym in your text).
* The topicDensityAssessment should appear and deliver a result similar to the original keywordDensityAssessment, but with somewhat higher %, in case both supplied words in the keyphrase occur in the text.
* Try adding multi-word items to the list, items with non-comma punctuation marks between words, words in different cases.

Fixes #1544

### DIFF
--- a/spec/assessments/keywordDensitySpec.js
+++ b/spec/assessments/keywordDensitySpec.js
@@ -17,11 +17,19 @@ describe( "An assessment for the keywordDensity", function() {
 
 		paper = new Paper( "string with the keyword", { keyword: "keyword" } );
 		result = keywordDensityAssessment.getResult( paper, factory.buildMockResearcher( {
+			getKeywordDensity: 0.1,
+			keywordCount: 1,
+		}, true ), i18n );
+		expect( result.getScore() ).toBe( 4 );
+		expect( result.getText() ).toBe( "The keyword density is 0.1%, which is too low; the focus keyword was found 1 time." );
+
+		paper = new Paper( "string with the keyword", { keyword: "keyword" } );
+		result = keywordDensityAssessment.getResult( paper, factory.buildMockResearcher( {
 			getKeywordDensity: 10,
 			keywordCount: 1,
 		}, true ), i18n );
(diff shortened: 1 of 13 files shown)


manuelaugustin:
CR & acceptance test done 👍

I made some tiny code style changes and as discussed, I implemented these minor fixes:

- When the keywords/synonyms overlap (e.g., `key, key word`) the longer version is now correctly marked.
- Plurals in feedback strings are fixed (i.e., `1 time` instead of `1 times`).

Could you do a quick CR of the last commit? And there are also some merge conflicts that need to be solved. After that's done, I can merge the PR!

nataliashitova:
@manuelaugustin CR and acceptance 👍 
I also fixed feedback strings for keywordDensity for the exact same matter. Could you please CR and acceptance test this?

manuelaugustin:
CR & acceptance test of the last additions: 👍

PR #1580, opened 2018-07-05
Use keyword density instead of topic density when synonyms are set

### LINKED ISSUE #1577: Remove topic density assessment
The topic density needs to be removed. Instead, the keyword density assessment should always be triggered, also if synonyms are set.

manuelaugustin:
## Summary

* Removes the topic density assessment and uses the keyword density assessment also when synonyms are set.

## Test instructions

* Make a beta version using `release/7.8` for Yoast SEO Premium and this branch for YoastSEO.js
* Add a text with > 100 words. Set a focus keyword and a synonym.
* Add the focus keyword and the synonym in the text.
* Make sure that the keyword density assessment is triggered also after you've added the synonym. It should only show keyword occurrences (both in terms of counting as well as in terms of 

Fixes #1577


src/researches/topicCount.js
```diff
@@ -12,17 +12,24 @@ const Mark = require( "../values/Mark.js" );
...
+module.exports = function( paper, onlyKeyword = false ) {
 	const keyword = paper.getKeyword();
 	const synonyms = parseSynonyms( paper.getSynonyms() );
 	const text = normalizeQuotes( paper.getText() );
 	const sentences = getSentences( text );
+	let topicWords = [];
 
-	let topicWords = [].concat( keyword, synonyms ).filter( Boolean );
-	topicWords.sort( ( a, b ) => b.length - a.length );
+	if ( onlyKeyword === false ) {
```
nataliashitova: You can start with the non-default positive condition and then go to the default false. Makes more sense in my head.

src/researches/topicCount.js
```diff
@@ -12,17 +12,24 @@ const Mark = require( "../values/Mark.js" );
...
+		topicWords = topicWords.concat( keyword, synonyms ).filter( Boolean );
+		topicWords.sort( ( a, b ) => b.length - a.length );
+	} else {
+		topicWords = topicWords.concat( keyword ).filter( Boolean );
```
nataliashitova: You do not need to use `.filter( Boolean) ` I thing, since keyword cannot be empty at the moment this research is reached.

spec/researches/topicCountSpec.js
```diff
@@ -77,4 +77,9 @@ describe( "Test for counting the keyword and synonyms in a text", function() {
 			{ keyword: "", synonyms: "" } );
 		expect( topicCount( mockPaper ).count ).toBe( 0 );
 	} );
+
+	it( "only counts keywords when the onlyKeyword option is set to true", function() {
```
nataliashitova: Please change to "counts only keywords (and not synonyms) when the onlyKeyword option is set to true".

nataliashitova:
CR 👍

nataliashitova:
Acceptance 👍

PR #1585, opened 2018-07-09
Change boundaries, add gray bullet, enable markings in good condition

### LINKED ISSUE #1584: Change keyword distribution boundaries
The keyword distribution boundaries should be changed to 40% (green) and 50% (yellow), independently of whether the paper has synonyms or not.

manuelaugustin:
## Summary

* n/a

* Makes boundaries the same regardless of whether or not there are synonyms (50% for orange, 40% for green).
* Enable markings also in the good condition.
* Show a gray bullet when there are not enough keyword occurrences to calculate instead of not triggering the assessment.

## Test instructions

* Add a text (>200 words) and a keyword.
* You should get a gray bullet and feedback about adding the keyword/synoym more often.
* Add the keyword once in the text. You should still get the same feedback.
* Add the keyword in the text again. The assessment should show feedback about the distribution.
* Distribute your keyword so that you just about get a green/orange result. Replace your keyword with a synoynm. The result should still be the same.
* Distribute your keyword so that you get a green result. Make sure you can still switch on markings.

Fixes #1584

### DIFF
--- a/spec/assessments/largestKeywordDistanceAssessmentSpec.js
+++ b/spec/assessments/largestKeywordDistanceAssessmentSpec.js
@@ -6,9 +6,17 @@ const Mark = require( "../../js/values/Mark.js" );
 
 let keywordDistanceAssessment = new largestKeyWordDistanceAssessment();
 
-describe( "An assessment to check the largest percentage of text in which no keyword occurs", function() {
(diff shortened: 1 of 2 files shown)


nataliashitova:
CR and acceptance 👍

PR #1586, opened 2018-07-09
Fix apostrophes in word boundaries

### LINKED ISSUE #1579: Only first mention of the word is highlighted
1. Enter text in the post:

> Synonyms
> We’re so excited to let you know that synonyms are coming to Yoast SEO Premium! As we are still working on this, naming and details are still in the works. In one of the upcoming releases, we’ll allow for synonyms to be used to accompany your focus keywords. So, that means that besides the actual focus keyword, we’ll also let you to insert a number of synonyms, and we will adjust the keyword density calculation accordingly.
> 
> Imagine you are writing about forests. You might also want to use the word ‘woods’ to refer to the same thing. You can set ‘forest’ as a keyword and ‘woods’ as a synonym. In addition, you can also use the synonym field to add the plural ‘forests’. To set multiple synonyms, just separate them with commas. Actually, in the future, you won’t need to fill in ‘forests’, when your keyword is ‘forest’ because all word forms (such as plurals, past tense of verbs, and adverbs from adjectives) will be recognised automatically. We are already on it, so stay tuned!
> 
> Note that this does differ from the multiple keywords option. That option allows you to optimize for totally different words, whereas we will take the synonyms into account for keyword density and other checks in our plugin. For instance, synonyms are also used when we calculate keyword distribution (more on that below).

2. Enter Focus keyword - forests and Keyword synonyms - woods
3. Click to Highlight keyword or its synonyms.
4. Check that only word forests is highlighted:
screen shot 2018-07-05 at 11 21 19

5. Remove quotes from second word 'woods':
screen shot 2018-07-05 at 11 22 46

6. Click to Highlight keyword or its synonyms.
7. Check the word woods is not highlighted, only the word forests is highlighted.
8. Remove quotes from first word 'woods':
screen shot 2018-07-05 at 11 24 11

9. Click to Highlight keyword or its synonyms.
10. Check that only first word woods is highlighted:
screen shot 2018-07-05 at 11 24 40

It is expected that the second word woods also should be highlighted and even in the step 7.

## Technical info
* Yoast SEO version: 7.8 RC3

igorschoester:
## Summary

* Fixes a bug that caused keywords to be incorrectly recognized within possessive forms (e.g. `Natalia` in `Natalia's fix`.

* Adjusted `matchTextWithArray` to use `matchTextToWord`.
* Match single quotes only when it is on the edge of a word boundary. This is to not count possessives like: `Natalia's fix`.

## Not fixed in this PR

The original text is changed. Then the matches are being found. However the index AND the actual match are not lining up with the original text at all. The markings are done by searching for the original text and replacing it. But the original text is not the original text!
In `stringProcessing/matchTextWithWord.js` you can see an place where the original text is changed before the matches are run.

## Test instructions

* Add a text with at least 200 words and set the focus keyword to `Natalia`. Add `Natalia's` in the text. No keyword occurrence should be found.

### DIFF
--- a/spec/researches/getTopicDensitySpec.js
+++ b/spec/researches/getTopicDensitySpec.js
@@ -26,7 +26,7 @@ describe( "Test for counting the keyword and synonyms in a text", function() {
 		mockPaper = new Paper( "<img src='http://image.com/image.png'>", { keyword: "key&word" } );
 		expect( getTopicDensity( mockPaper ) ).toBe( 0 );
 		mockPaper = new Paper( "This is a nice string with a keyword keyword keyword.", { keyword: "keyword" } );
-		expect( getTopicDensity( mockPaper ) ).toBe( 30 );
+		expect( getTopicDensity( mockPaper ) ).toBe( 20 );
 		mockPaper = new Paper( "a string of text with the beautiful Keyword in it.", { keyword: "keyword" } );
 		expect( getTopicDensity( mockPaper ) ).toBe( 10 );
 		mockPaper = new Paper( "a string of text with the Key word in it.", { keyword: "key word" } );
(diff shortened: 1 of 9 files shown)


manuelaugustin:
CR & acceptance test done 👍

PR #1594, opened 2018-07-10
Create example webworker skeleton

### LINKED ISSUE #1590: Create web worker skeleton in examples
Use webpack with the [worker loader](https://github.com/webpack-contrib/worker-loader).

Parent issue: https://github.com/Yoast/YoastSEO.js/issues/1571

igorschoester:
## Summary

* Added an example webpack app for testing purposes.

Took it a bit beyond the original scope since this is the first issue and went experimenting too.

* Used create-react-app for the setup. Ejected and removed some things.
* Set up the page like the browserified example without any function (and without snippet preview).
* Implemented an analysisWorker that just does a handshake.
* Added a redux store with 2 reducers: `paper` and `configuration`.
* Added `yoastseo` as alias to import it the same as in the plugin.
* Added typescript loader for the one ts file.
* Started work on scheduler: this should get continued in another issue to prevent bloating this much more.
* Added ESLint ECMAScript 2017 support until the Yoast ESLint config includes this PR: https://github.com/Yoast/eslint/pull/8

## Issues

~~Importing researchers in the base `YoastSEO.js` from the `src` results in requires not being found. This needs to be fixed in order to develop with it. For now they import from `js`.~~

## Test instructions

* 

Fixes #1590


igorschoester:
> - add SEO assessor to be called by the webworker, not only the content assessor

This will be done in this issue: https://github.com/Yoast/YoastSEO.js/issues/1620

>   - AnalysisWorker.js and analysis.worker.js are just confusing. Please call the first one AnalysisWorkerWrapper or AnalysisWrapper.

I will rename it to AnalysisWrapper.

>   - I am not a fan of the word "analyzation" and (especially) "analyzations". I do not see a reason to not call it simply "analysis" or "analyses".

Luckily this is just in the JSDocs and console log. Will change.

> - We should discuss if we want the built files (like `languages/yoast-seo.pot`) to be committed. 

This was accidental. I think these should only get included in a release process. Especially the `.pot` files since most of the time the only thing changed in there is the creation date. But since it is generated it should not be a problem for merging, just re-generate it if it causes problems.

> - Please add documentation about webworker architecture: information flow, what is already implemented, what is planned to be implemented.

I will add a markdown document with some basic documentation (like we already discussed now). The overview issue is more for what is planned at this point. I don't think that should live in the code.

PR #1598, opened 2018-07-11
Expose more API to use

atimmer:
## Summary

* Exposes all assessments and more as a public API.

* I've tried to use import/export where possible, but it is not easily possible to mix this with `require` statements and I didn't want to search and replace all occurrences of `require`.


igorschoester:
Don't forget to pull in develop again after https://github.com/Yoast/YoastSEO.js/pull/1595 was merged.

igorschoester:
Created an issue for OutboundLinksAssessment: It's not possible to get the best result. https://github.com/Yoast/YoastSEO.js/issues/1621

PR #1611, opened 2018-07-17
Add WP-style transliterations

### LINKED ISSUE #1607: Fix discrepancy in transliteration in the slug and in the YoastSEO.js
We use different ways of transliterating accent characters in [YoastSEO.js](https://github.com/Yoast/YoastSEO.js/blob/develop/src/config/transliterations.js) and in [WP](https://github.com/Yoast/wordpress-develop-mirror/blob/master/src/wp-includes/formatting.php). This is problematic, because the assessment that searches for keyword in the URL hardly ever finds anything for languages that have a lot of accent characters (e.g., Spanish, Norwegian, Turkish). Since importing the original transliteration map from WP is complicated, we should build in a second transliteration map into YoastSEO.js which will only be used for search within URLs.

Examples are available in #1193.

nataliashitova:
## Summary

* Improves the recognition of keywords with special diacritics in the URL.

* A second transliteration map was added in order to match the one that is used by [WP](https://github.com/Yoast/wordpress-develop-mirror/blob/master/src/wp-includes/formatting.php). This transliteration map is much less specific than the one that we have been using so far, but it is used to create automatic permalinks. This way we can match with both language-specific transliterations (that were already available in YoastSEO.js) and with the straightforward  WP way.

## Test instructions

* Take a text in a language that contains a lot of accent characters (i.e., any letters from the latin alphabet that additionally have upper elements such as accents, circles, dots, hooks, or lower elements such as tails).
* Make one of the words that contains accent characters your keyword and use it in the title. 
* Check if in the transliteration that appeared in the permalink (automatically, after publication) all accent characters were substituted by their ASCII analogs.
* Check if the URL-Keyword assessment returns "The focus keyword appears in the URL for this page."

Fixes #1607

### DIFF
--- a/spec/stringProcessing/transliterateWPstyleSpec.js
+++ b/spec/stringProcessing/transliterateWPstyleSpec.js
@@ -1,0 +1,79 @@
+const transliterationObjects = require( "../../js/config/transliterationsWPstyle.js" );
+const transliteration = require( "../../js/stringProcessing/transliterateWPstyle.js" );
+
+describe( "a test for returning correct number of transliteration objects for different languages", function() {
+	it( "returns only generic transliteration objects for Spanish.", function() {
+		expect( transliterationObjects( "es_ES" ).length ).toBe( 309 );
+	} );
+
+	it( "returns only generic transliteration objects for English.", function() {
+		expect( transliterationObjects( "en_EN" ).length ).toBe( 309 );
+	} );
+
+	it( "returns generic transliteration objects and language-specific transliteration objects for German.", function() {
+		expect( transliterationObjects( "de_DE" ).length ).toBe( 317 );
+	} );
+
+	it( "returns generic transliteration objects and language-specific transliteration objects for Danish.", function() {
+		expect( transliterationObjects( "da_DK" ).length ).toBe( 315 );
+	} );
+
+	it( "returns generic transliteration objects and language-specific transliteration objects for Catalan.", function() {
+		expect( transliterationObjects( "ca_CA" ).length ).toBe( 310 );
+	} );
+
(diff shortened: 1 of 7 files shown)


spec/stringProcessing/transliterateWPstyleSpec.js
```diff
@@ -0,0 +1,79 @@
...
+
+	it( "returns a Danish string without special characters.", function() {
+		expect( transliteration( "åÅææÆÆøØéÉ", "da_DK" ) ).toBe( "aaAaaeaeAeAeoeOeeE" );
+	} );
+
+	it( "returns a Dutch string without special characters.", function() {
+		expect( transliteration( "çÇñÑèÈêÊéÉëËôÔöÖïÏüÜäÄ", "nl_NL" ) ).toBe( "cCnNeEeEeEeEoOoOiIuUaA" );
+	} );
+
+	it( "returns an unchanged string for a non-existing locale.", function() {
```
manuelaugustin: The string here will still be changed, e.g. `expect( transliteration( "äbc", "qxz" ) ).toBe( "abc" )` will return true; the same goes for `expect( transliteration( "äbc", "" ) ).toBe( "abc" )`. This is because the `isUndefined( locale )` check in `transliterationsWPstyle.js` will return `false` for `"qxz"` and `""`. It only returns true when there is no locale (so `expect( transliteration( "äbc" ) ).toBe( "äbc" )` works).

If the intended behavior is to not transliterate for a non-existing/empty locale, I think the check has  to be changed.

manuelaugustin:
CR and acceptance done 🚧

* There is one small issue in the specs, see comment.
* I'd change the changelog entry to something that tells uses a bit more clearly what has been changed, e.g. `Improves the recognition of keywords with special diactritics in the URL.`

PR #1660, opened 2018-08-02
Implement webworker scheduler

### LINKED ISSUE #1617: Implement a scheduler in the webworker

Create a scheduler in the webworker so that we can control which tasks will run.
For now it should be that the scheduler should run only the newest added and throw away any tasks that are older.

Already made a start in the example PR: https://github.com/Yoast/YoastSEO.js/pull/1594

requires #1590

igorschoester:
## Summary

* N/A

* Added an index.js to the full text tests to export an array of the texts (used with the spam button).
* Added a polling mechanism in the scheduler to check for the next task.

## Test instructions

* In the webpack example folder run `yarn` and `yarn start`.
* Use the initialize button first.
* Use the analyze spam button and check the console. The spam requests every test text 10 times. At this point we have 2 texts.
* Check that the worker only runs the last analysis. Or when spamming the spam button, only runs the newest after completion of the last and then always the last one too.

Fixes #1617


nataliashitova:
CR and Acceptance 👍 
Some assessments still failing (both in spam and in regular Analyze), which might be due to the `window` problem described in https://github.com/Yoast/YoastSEO.js/issues/1669.

PR #1671, opened 2018-08-06
Refactor AnalysisWorkerWrapper callbacks

### LINKED ISSUE #1618: Refactor callbacks inside the AnalysisWorker

Refactor callbacks inside the AnalysisWorker (Wrapper).
This should be something like a QueuedItems class to have flexibility to include more values like time, resolve, reject.

igorschoester:
## Summary

* N/A

* Added a Request class for the AnalysisWorkerWrapper that holds the promise's resolve & reject along with optional extra data.
* `Request.resolve( payload )` creates a Result from the response payload and passes that back to the original callee (example).

## Test instructions

* Does the example still function like before?

Fixes #1618


src/worker/AnalysisWorkerWrapper.js
```diff
@@ -39,38 +36,41 @@ class AnalysisWorkerWrapper {
...
 	 */
-	handleMessage( { data: { type, payload } } ) {
-		let id;
-		console.log( "wrapper", type, payload );
+	handleMessage( { data: { type, id, payload } } ) {
+		let response, request;
+		console.log( "wrapper", type, id, payload );
+
 		switch( type ) {
 			case "initialize:done":
```
manuelaugustin: `initialize:done` and `analyze:done` seem to have the same content. Any reason these are two different cases? Is is it because we expect more differing functionality to be implemented very soon?

src/worker/scheduler/Task.js
```diff
@@ -8,10 +8,10 @@ class Task {
 	/**
 	 * Initializes a task.
 	 *
-	 * @param {number|string} id      The task identifier.
-	 * @param {function}      execute Executes the job with the data.
-	 * @param {function}      done    Callback for the scheduler.
-	 * @param {Object}        [data]  Optional data for when executing the task.
+	 * @param {number}   id      The task identifier.
```
manuelaugustin: So it's been decided that `id` can't be a number anymore? If so, the check and the error message should also be changed.

manuelaugustin:
CR & acceptance test done 👍

Issue #1426, opened 2018-03-01
Eye marker in the analysis doesn't highlight anything on categories or tags


benvaassen:
## What did you expect to happen?
When I click the eye marker, I'd expect to see the issue highlighted. 

## What happened instead?
Instead, it doesn't highlight anything when I click the eye marker. 

## How can we reproduce this behavior?
1. Create a new category/tag or edit an existing category/tag
2. Write some text that triggers an eye marker. 
3. Click the eye marker and see nothing happens. 

![marker](https://user-images.githubusercontent.com/11910661/36839863-80087e18-1d43-11e8-87a7-41f94ced57ef.png)

Also happens for WooCommerce product categories. 

## Technical info
* WordPress version: 4.9.4
* Yoast SEO version: 6.3.1

benvaassen:
Please inform the customer of conversation # 172794 when this conversation has been closed.

benvaassen:
Please inform the customer of conversation # 173648 when this conversation has been closed.

benvaassen:
Please inform the customer of conversation # 195758 when this conversation has been closed.

monbauza:
Please inform the customer of conversation # 342434 when this conversation has been closed.

CarolineGeven:
Closing in favor of https://github.com/Yoast/wordpress-seo/issues/10564

PR #1675, opened 2018-08-07
Remove window reliance in assessor

### LINKED ISSUE #1669: Remove window-reliance from the worker's markers
We need to fix `window` reliance on markers (hasMarker). The worker should only return a list of Mark objects and these should be handled outside the web worker, because the worker cannot access window.

This issue might also be responsible for the fact that some assessments (that need markings) are failing (e.g., see by AnalyzeSpam button).

nataliashitova:
## Summary

* 

* Removes the check of the window presence and window parameters in assessor.js. Assessor is used by the worker, who has no access to window.
* `yoastHideMarkers` in only used in the term-scrapper to disable the markings in Category pages. @CarolineGeven why don't we show makings on Category pages? Also see #1426.
* In general, we want to implement architecture where YoastSEO.js simply provides a list of markings, while the plugin (or any other plugin that makes use of this library) decides whether or not and how to show the markings.

## Test instructions

* Run the `analyze spam` and check console that all assessments have run successfully.

Fixes #1669

### DIFF
--- a/src/assessor.js
+++ b/src/assessor.js
@@ -75,10 +75,6 @@ Assessor.prototype.isApplicable = function( assessment, paper, researcher ) {
  * @returns {boolean} Whether or not the assessment has a marker.
  */
 Assessor.prototype.hasMarker = function( assessment ) {
-	if ( ! isUndefined( window ) && ! isUndefined( window.yoastHideMarkers ) && window.yoastHideMarkers ) {
-		return false;
-	}
-
 	return isFunction( this._options.marker ) && ( assessment.hasOwnProperty( "getMarks" ) || typeof assessment.getMarks === "function" );
 };
 


igorschoester:
CR & Acceptance 👍 
If there should be another action taken due to a response from Caroline, we should do that in a new issue.

CarolineGeven:
@nataliashitova The markers currently don't work for categories. So they should not be displayed at all. I'm moving #1426 to the wordpress-seo repository.

CarolineGeven:
This currently does not work taxonomies. We need to investigate why this doesn't. The eye markers should not be shown.

https://github.com/Yoast/wordpress-seo/issues/10563
https://github.com/Yoast/wordpress-seo/issues/10564

Issue #1687, opened 2018-08-08
Build via webpack


igorschoester:
## Current situation
The build is done via grunt. Which actually build all the JavaScript with the typescript builder. (and twice because of the templates).

## Wanted situation
Use webpack. This will allow us to specify loaders for each type.

*How we found out:*
With the webworker we found out about this because we kept running into require / import problems. The code generated by TypeScript is not having a helper around the fact that a require could live inside a `.default` space. Which webpack does do properly. Also, the worker is created with the assumption of the worker-loader, which is not currently being done in the built. So the `js` file created does not work on the fly.

igorschoester:
TypeScript removal is done in https://github.com/Yoast/YoastSEO.js/pull/1745

Kingdutch:
Hi igorschoester,

I don't know if this is the right issue to post this in or whether we should open a new issue.

For the Drupal module that uses your library I'm trying to compile YoastSEO using Webpack. The imports are a good first start (I did run into #1878 ). However, building this with Webpack and Babel is currently impossible because the project requires keeping too many files in memory at the same time.

```
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 0x10003907e node::Abort() [/usr/local/Cellar/node/10.11.0/bin/node]
```

I've tried converting all the `require` statements in my own code into proper `import` statements but I suspect that Webpack is choking on all the `require` statements throughout the project.

Are there any plans to convert the internal code for YoastSEO to also use import statements instead of require statements? If you're aware of a different solution without conversion then that would also be great.

You can check out the out of memory error by cloning https://github.com/goalgorilla/RTSEO.js/tree/feature/improved-build-pipeline and running `yarn && yarn build` to install dependencies and start Webpack in production mode.

~ Alexander

Kingdutch:
The `module.exports` to `export default` is fixed in #1919

atimmer:
@Kingdutch #1919 covers your short-term issue. I've put that in our process to review.

#1930 should cover the longer term issue.

PR #1690, opened 2018-08-08
Compatibility for app refresh alternative

### LINKED ISSUE #1677: Create an alternative to the app.refresh in the plugin

## Introduction
The current `App` has a refresh function which triggers the analyses. The `App` collects the data and then does the analyses on them. With the worker, the `App` will no longer be used. Therefore the plugin will need to collect the data.

## Task
Create an alternative function in the plugin that runs the analyses.

- Import and initialize the AnalysisWorkerWrapper. Also ensure the translations are passed here.
- Make it possible to receive configuration parameters so that new assessments can be added. This is necessary for the add-ons, which add assessment via the pluggable. This will get done in https://github.com/Yoast/YoastSEO.js/issues/1680
- Collect the data: only use the existing getData for now. This will get done in https://github.com/Yoast/YoastSEO.js/issues/1676
- Request the analyses from the worker (Call `AnalysisWorkerWrapper.analyze` with the complete `Paper`).

igorschoester:
## Summary

* N/A

* Change imports to requires because of the build process. See: https://github.com/Yoast/YoastSEO.js/issues/1687 for more information.
* Add AnalysisWorker to the `index.js`.
* Move Grunt `build:browserify` task out of the `build:js` task to `build:browserified` to speed up the normal build.
* Add `debouncedRefresh` to the app arguments, which defaults to true. We need an unbound refresh to overwrite it. So we pass this as false in the plugin.
* Add `serialize` and `parse` function to the `AssessmentResult`, `Mark` and `Paper` values.

## Test instructions

*

Fixes #1677


manuelaugustin:
CR & acceptance done 👍

PR #1693, opened 2018-08-08
Update the locale of the configuration of the content assessor to paper locale

### LINKED ISSUE #1657: Combine the analysis configuration with it's global one before running

Combine the analysis configuration with it's global one before running.
This means the global configuration will effectively be the default for the worker after initialization.

Think about which configuration options there are. Which of them are needed in the initialize and which are needed in the analyze. The constructor should probably call initialize.

Assessor options in the worker: useCornerStone, useKeywordDistribution, i18n, marker, locale.

nataliashitova:
## Summary

* 

* Checks if the configuration locale is the same as the paper locale and updates it if not.
* Removes configuration from the `analyze` function, which means `useCornerstone` and `useKeywordDistribution` should be passed to the configuration of `initialize`. It also means that the `initialize` should be rerun if the status cornerstone content / non-cornerstone content changes.

## Test instructions

* The two assessments that require configuration locale are Flesch reading ease and Sentence Length. Check if the results that are returned for English, Russian and Italian are in accordance with the default (en_EN) or with the language-specific configs (ru_RU, it_IT) in `src/config/content`.
* E.g., a 19-word sentence in the text should return a green bullet for English and Italian, but a red bullet for Russian. A 24-word sentence should return a green bullet for Italian, but a red bullet for English and Russian.

Fixes #1657


igorschoester:
CR & acceptance 👍

PR #1699, opened 2018-08-10
Check paper status before analysis

### LINKED ISSUE #1658: Improve Paper instance creation

Check if the Paper actually changed on the analyze request to prevent unnecessary initialization.

nataliashitova:
## Summary

* n/a

* Saves the last instances of paper and result to the worker.
* Checks if the paper has changed since the last analysis and if not returns the previous result.
* Checks if the paper text has changed since the last analysis and if not returns the previous result of the readability checks.
* Add a paper reset to the the initialize function.
* Changes default ID so that it's not a string.

## Test instructions

* Run initialize and analyze on the empty paper. Check that no errors are triggered.
* Add some text and rerun analyze. Check that the result has changed.
* Add some more attributes and rerun analyze. Check that the SEO result was updated, but the readability analysis was not rerun (check console logs).
* Change both text and attributes and rerun analyze. Check that both SEO and readability analyses were updated.
* Rerun analyze without changing anything. The console should return "The paper has not changed since you analyzed it last."

Fixes #1658

### DIFF
--- a/src/worker/analysis.worker.js
+++ b/src/worker/analysis.worker.js
@@ -2,6 +2,7 @@
 import Jed from "jed";
 import omit from "lodash/omit";
 import merge from "lodash/merge";
+import isEqual from "lodash/isEqual";
 
 // YoastSEO.js dependencies.
 import * as Paper from "yoastseo/values/Paper";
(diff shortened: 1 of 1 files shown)


manuelaugustin:
CR & acceptance done 👍

PR #1729, opened 2018-08-23
Fixes issue where readability analysis wasn't loaded on an empty post (or new post)

### LINKED ISSUE #1722: Readability analysis doesn't load until there's content
When you open a new post, the readability analysis will keep showing the spinner icon until you add content to the body of the post. (In Gutenberg, just clicking into the body of the post will also make it load.)

We need to change this so that the readability analysis loads right away with the default results for an empty post.

atimmer:
## Summary

* _not applicable_

* Set `_paper` to `null` so it can be distinguished from a Paper with an empty text.

## Test instructions

* Add a new post (with this branch loaded in `wordpress-seo`) and see that the readability analysis contains an assessment result asking to write more text.

Fixes #1722

### DIFF
--- a/src/worker/AnalysisWebWorker.js
+++ b/src/worker/AnalysisWebWorker.js
@@ -175,6 +175,7 @@ export default class AnalysisWebWorker {
 		switch( type ) {
 			case "initialize":
 				this.initialize( id, payload );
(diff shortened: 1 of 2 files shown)


igorschoester:
CR & Acceptance 👍

Issue #1736, opened 2018-08-27
Convert all our usages of lodash to `lodash-es`


atimmer:
Currently we import like this:

```js
import isEqual from "lodash/isEqual"
// or
const isEqual = require( "lodash/isEqual" );
```

This needs to be changed to

```js
import { isEqual } from "lodash-es"
// or
const isEqual = require( "lodash/es" ).isEqual;
```

This will allow us to use tree-shaking.

PR #1737, opened 2018-08-27
Convert require's and module.exports to ES modules

atimmer:
## Summary

* N/A

* Some places had this pattern: `const thing = require( "thing" )().theActualThing`.
* Which I've replaced with: `const thingFactory = require( "thing" ); const thing = thingFactory().theActualThing;`. It isn't the nicest pattern, but it works well.

## Test instructions

* Make sure everything still works int he example.

Requires https://github.com/Yoast/YoastSEO.js/issues/1736


spec/fullTextTests/testTexts/englishPaper2.js
```diff
@@ -146,7 +146,13 @@ const expectedResults = {
 	},
 };
 
-module.exports = {
+export {
+	name,
```
manuelaugustin: Why do we need an `export` in addition to an `export default`?

spec/fullTextTests/testTexts/englishPaper1.js
```diff
@@ -157,7 +157,13 @@ const expectedResults = {
 	},
 };
 
-module.exports = {
+export {
```
manuelaugustin: Same as below

src/helpers/errors.js
```diff
@@ -19,6 +19,8 @@ function showTrace( errorMessage ) {
 	}
 }
 
-module.exports = {
+export { showTrace };
```
manuelaugustin: Same as above

manuelaugustin:
CR & acceptance done 👍

PR #1745, opened 2018-08-29
Remove TypeScript

atimmer:
## Summary

* _Drop TypeScript support_

It has given us no gains and a bit of annoyance. We can add it later,
but that needs to occur with an effort to rewrite at least a great part
of the codebase to use types. That way we actually get some results from
using TypeScript.

## Test instructions

* Everything should still build.
* `yarn test` should pass.
* `grunt check` should pass.
* No traces of TypeScript should be left.


igorschoester:
Acceptance 🏗 

After `grunt:build` the tests fail on the templates:
appspec

Also the built script (beta & yoast-cli) seem to fail now, probably due to no `js` folder? 😉

igorschoester:
Acceptance 👍

PR #1766, opened 2018-09-03
Expose more public API required in yoast-components

atimmer:
## Summary

* Exposes word boundaries for use in other libraries or applications.

* We don't want to expose `createWordRegex` but yoast-components needs it.


manuelaugustin:
CR & acceptance done 👍

Issue #1650, opened 2018-07-30
Refactor URLKeyword assessment


nataliashitova:
We need to refactor URL keyword assessment to accommodate morphology.

* If the keyword has 1 or 2 content words: GOOD if all content words are in the URL, OK otherwise. 
* If the keyword has >2 content words: GOOD if > half of content words are in the URL, OK otherwise.

Please also see if this change fixes the problem in  #1608.

PR #1778, opened 2018-09-04
Refactor urlkeyword assessment

### LINKED ISSUE #278: Analyses does not recognise &amp; as &
When using a focus keyword with an "&" the analysis is broken on the following elements:
- Title
- Slug
- Paragraphs
- Headings

It works on the alt tags in images.

Tested using the focus keyword `pooch & mutt`.
Confirmed this is an issue still present in RC 3.1

### LINKED ISSUE #1608: Punctuation marks used in the keyword get stripped in slug
We need to improve recognition of the keyword in the URL. The problem is that the punctuation marks are stripped in slug, but not in the keyword.
In line with the upcoming changes to include morphological support and synonyms and smarter matching of keyphrase words in URL (see #1558), we should parse the words out of the keyword and out of the URL and compare the arrays. 
See examples in #1193 and #278.

This issue will be addressed together with the refactoring of the URLKeyword assessment within the #1558.

igorschoester:
## Summary

* Implements a new scoring schema for the assessment that checks if the keyword is used in the slug. If the keyword has 1 or 2 content words: GOOD if all content words are in the URL, OK otherwise. If the keyword has >2 content words: GOOD if > half of content words are in the URL, OK otherwise.
* Fixes the bug when the keyword would not be found in the slug when containing punctuation.

* Also fixed the missing quotation marks in the KeyphraseDistribution assessment.

## Test instructions

* Make a post with a short keyphrase (2 content words) and make sure to not use them in the slug. The assessment should return an OK result.
* Use one of the words from your keyphrase in your slug. The assessment should still return an OK result.
* Use the second word from the keyphrase in your slug. The assessment should return a GOOD result.
* Add 3 more content words to your keyphrase without changing the slug. The assessment should return an OK result.
* Use one of the words from your keyphrase in your slug. The assessment should return a GOOD result.
* Remove the keyphrase. The assessment should not appear at all.
* Since this PR also fixes a couple bugs with keyphrase recognition in slugs, try using punctuation marks in the keyphrase (e.g., `St. Paul's Cathedral - Visit`, `pooch & mutt`).

Fixes https://github.com/Yoast/YoastSEO.js/issues/1650 Fixes #1608 Fixes #278


hansjovis:
Acceptance test: ✅ 
* [x] OK result when key phrase is not contained in the slug.
* [x] OK result when only part of the key phrase is contained.
* [x] GOOD result when the whole key phrase is contained.
* [x] OK when less than half of the words in a long key phrase are contained.
* [x] GOOD when more than half of the words in a long key phrase are contained.
* [x] No score when no key phrase.
* [x] Fixes checking of punctuation marks in slug.

PR #1812, opened 2018-09-19
Refactor Polish participle direct precedence exceptions to not use Re…

igorschoester:
## Summary

* Simplify the lookup of the Polish participle direct precedence exceptions.

* Replace the giant RegExp with a for loop with includes in array check.

## Test instructions

* Run `yarn test`.
* Combine with `wordpress-seo`.
* In Chrome on Windows, create a post with the following: `została założona` (an auxiliary followed by a participle; could be any combination that are included in the lists).

Fixes https://github.com/Yoast/wordpress-seo/issues/11011


manuelaugustin:
CR & acceptance test for the Polish passive voice done 👍

Now we still need to test it on Windows to see whether no more browser crashes occur.

PR #1823, opened 2018-09-26
Skip images and empty divs when searching for keywords in the first paragraph

### LINKED ISSUE #1380: Focus keyword isn't recognized in the first paragraph
Copied from https://github.com/Yoast/wordpress-seo/issues/5803#issuecomment-357237159

Did not expect to get this error message as the focus keyword IS in the first paragraph.

 yoast

Using Version 3.6.1. Now you have mentioned the text view I'm thinking that having an image there is causing the problem. See attached. The keyword is underlined and is in the first paragraph of text but there is also an image, right aligned, which appears before the first paragraph when in text view. 
yoast2

@terw-dan:
>My guess is the image is seen as the first paragraph, so the text with the focus keyword will be seen as the second paragraph. 
We'll have to look into this to see if there is a way to correctly check whether a paragraph only has an image and check for the first paragraph that contains text.

## Another report:
I have a similar issue. To get around storage capacity constraints on my server, I embed photos from SmugMug in my featured photos posts. The embed code is placed above the copy, so I always get the message "The focus keyword doesn't appear in the first paragraph of the copy. Make sure the topic is clear immediately." 

See the attached screenshot for an example from an upcoming post, where my keyword phrase is colored powder for Diwali.
screen shot 2017-01-23 at 3 28 36 pm

nataliashitova:
## Summary

* Improves keyword recognition in the first paragraph on texts which start with images and captions

* Skips images and empty divs in the beginning of the text to provide correct feedback in the keywordInFirstParagraph assessment

## Test instructions

* Enter a text that starts with an image or a caption with an image or a division with an image (and nothing else), while the keyphrase appears in the first _real_ paragraph. Check if the keyphrase is recognized correctly.
* Use the following (or similar) patterns

```
<img class=“size-medium wp-image-801” src=“https://test.com/text.jpg” alt=“Some alttag” width=“300” height=“173” />

The text with the keyword. 
```

```
<div style="text-align: center;"><img src="https://test.com/test.jpg" alt="Some alttag" /></div>
<div>The text with the keyword. </div>
```

```
<div style="text-align: center;"><a><img src="https://test.com/test.jpg" alt="Some alttag" /></a></div><div>The text with the keyword. </div>
```

* Watch out! After copying html from here into the local environment there might appear extra symbols that do not allow proper analysis. Make sure to remove them before testing.

Fixes #1380


src/researches/findKeywordInFirstParagraph.js
```diff
@@ -16,13 +19,44 @@ import { isEmpty } from "lodash-es";
  * @returns {boolean} True if the text consists only of images, false otherwise.
  */
 function paragraphConsistsOfImagesOnly( text ) {
```
hansjovis: Since this function also checks for anchor tags and empty divisions as well, I would update this function's name. Perhaps `paragraphDoesNotContainText` ?

src/researches/findKeywordInFirstParagraph.js
```diff
@@ -16,13 +19,44 @@ import { isEmpty } from "lodash-es";
...
+	if ( anchors.length > 0 ) {
+		anchors.forEach( function( anchor ) {
+			text = text.replace( anchor, "" );
+		} );
+
+		if ( text === "" ) {
+			return true;
+		}
+	}
+
```
hansjovis: The block of code above may be a candidate for a little private function (in this file) 😉 .
Just to remove the anchor tags from the text. Then afterwards check if the text is empty after the function was applied.

src/researches/findKeywordInFirstParagraph.js
```diff
@@ -16,13 +19,44 @@ import { isEmpty } from "lodash-es";
...
+		imageTags.forEach( function( imageTag ) {
+			text = text.replace( imageTag, "" );
+		} );
+
+		if ( text === "" ) {
+			return true;
+		}
+	}
+
```
hansjovis: May also be a candidate for a private function.

src/researches/findKeywordInFirstParagraph.js
```diff
@@ -16,13 +19,44 @@ import { isEmpty } from "lodash-es";
...
+	// Remove empty divisions from the text
+	const emptyDivisions = findEmptyDivisions( text );
+	if ( emptyDivisions.length < 1 ) {
 		return false;
 	}
 
-	images.forEach( function( image ) {
-		text = text.replace( image, "" );
+	emptyDivisions.forEach( function( emptyDivision ) {
+		text = text.replace( emptyDivision, "" );
```
hansjovis: This block as well.

hansjovis:
Acceptance test:  ✅ .

PR #1829, opened 2018-10-01
Make sure that common OS files (like .DS_Store) are not tracked.

hansjovis:
## Summary

Adds some pesky little OS files to the `.gitignore` so that they are not tracked by the repository.
These are files that are automatically created by Mac and/or Windows for e.g. showing thumbnails in the finder or explorer.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,10 @@ coverage/
 dist/
 .tscache/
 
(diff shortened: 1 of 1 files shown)


diedexx:
Transferred to https://github.com/Yoast/javascript/pull/22

Issue #1350, opened 2017-12-22
Competing links assessment marker doesn't mark anything.


Xyfi:
## What did you expect to happen?
When I click the competing link assessment marker, I expect the link that links to the competing post to be marked.

## What happened instead?
Nothing gets marked.

## How can we reproduce this behavior?
* Create a post with a keyword and save it. Give it a unique and easy to remember title.
* Create a second post with the same keyword and create a link to the first post.
* Make sure the competing link assessment is shown. It has the following text in English: `You're linking to another page with the focus keyword you want this page to rank for. Consider changing that if you truly want this page to rank.`
* Click the marker and notice that the link is not highlighted as it should be.

## Technical info

* Platform: WordPress
* Platform version: 4.9.1
* Yoast SEO version: 6.0

igorschoester:
Closing this because it is a duplicate of https://github.com/Yoast/YoastSEO.js/issues/1155

PR #1830, opened 2018-10-01
Refactor conflicting links assessment

### LINKED ISSUE #1825: Refactor TextCompetingLinks assessment to include morphology
We need to refactor textCompetingLinks assessment so that it accommodates morphology and the distinction between content words and function words.

Currently, the assessment returns a BAD score if the copy contains a link with the text that is an exact match of the focus keyword.

After refactoring the assessment should return a BAD score if **all content words** from the link text are in the keyphrase and **visa versa**. This will prevent bad scores when linking to overview articles (e.g., from an article about `mountain biking in France` to an article about `mountain biking` in general) and to more specific articles (e.g., from an article about `mountain biking` in general to an article about `mountain biking in France`). 

Both keyphrase and synonyms should be taken into account here.

nataliashitova:
## Summary

* Applies morphological researcher to the text conflicting links assessment.

* 

## Test instructions

* For English Premium
  * Take a text with a link that does not point at itself and make its text an exact match of your keyphrase. The assessment should return a bad score.
  * Take a text with a link that does not point at itself and make its text a form of your keyphrase or a synonym. The assessment should return a bad score.
  * Check if only full matches trigger a red bullet (i.e., all content words from the link text should be in the keyphrase or a synonym and vice versa)
* For English Free and for all other languages with function-word support
  * Check if function words get trimmed when the assessment is applied and that the word order does not matter.
* For languages without function-word support (Free and Premium)
  * Check if word order does not matter.
  * For Premium: check if synonyms are taken into account.

Note that highlights for this assessment don't work at the moment; this is not a regression: https://github.com/Yoast/YoastSEO.js/issues/1350

Fixes #1825

### DIFF
--- a/spec/assessments/TextCompetingLinksAssessmentSpec.js
+++ b/spec/assessments/TextCompetingLinksAssessmentSpec.js
@@ -122,15 +122,15 @@ describe( "A test for marking multiple competing links", function() {
 		let expected = [
 			new Mark( {
 				original: "<a href='http://example.com/keyword'>keys test wording phrased</a>",
-				marked: "<yoastmark class='yoast-text-mark'><a href='http://example.com/keyword'>keys test wording phrased</a></yoastmark>"
+				marked: "<yoastmark class='yoast-text-mark'><a href='http://example.com/keyword'>keys test wording phrased</a></yoastmark>",
 			} ),
 			new Mark( {
 				original: "<a href='http://example.com/keyword'>articles which are interesting</a>",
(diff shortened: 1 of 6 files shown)


manuelaugustin:
CR & acceptance done 👍 

I'd suggest moving `keywordInAnchor` + its associated helper functions out of `getLinkStatistics.js` and into a separate file. I'll add this point to the post-RC clean-up issue.

PR #1834, opened 2018-10-04
Fix cornerstone-specific values in KeywordSubheadings assessment

nataliashitova:
## Summary

* Fixes an erroneously deprecated cornerstone-specific score for an assessment that checks for keyphrase in subheadings

* The score for a case where the only subheading in the article contains keyphrase is put back to 6 (orange bullet) from 9 (green bullet).

## Test instructions

* Check that the KeywordSubheadings assessment [returns correct values](https://github.com/Yoast/YoastSEO.js/wiki/Scoring-SEO-analysis#5-keyword-in-subheadings) for the non-cornerstone content.
* Check that the KeywordSubheadings assessment returns correct values for the cornerstone content.
* Specifically check the case where there is one subheading in the text and it has keyphrase in it.

### DIFF
--- a/spec/cornerstone/SEOAssessorSpec.js
+++ b/spec/cornerstone/SEOAssessorSpec.js
@@ -184,9 +184,10 @@ describe( "running assessments in the assessor", function() {
 			expect( assessment ).toBeDefined();
 			expect( assessment._config ).toBeDefined();
 			expect( assessment._config.scores ).toBeDefined();
-			expect( assessment._config.scores.noMatches ).toBe( 3 );
+			expect( assessment._config.scores.tooFewMatches ).toBe( 3 );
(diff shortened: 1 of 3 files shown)


hansjovis:
Acceptance test: ✅ 

The copy might need to be changed for the cornerstone check.
The orange score notifies the user that is needs improvements, but the text:
`` Keyphrase in subheading: Your subheading reflects the topic of your copy. Good job! ``
does not tell the user why it needs improvement and how it can be improved.

nataliashitova:
This PR was a result of comparing the old scoring schema (before Recalibration) with the Morphology-related scoring schema. Currently, there is not reason to return lower (OKAY) scores to the cornerstone articles with the only subheading reflecting the topic, comparing to the non-cornerstone articles that receive a high (GOOD) score under the same conditions.

PR #1844, opened 2018-10-09
Fix faulty link statistics research

### LINKED ISSUE #1843: Fix TextCompetingLinks which changes the researcher
In the current implementation the assessments that are executed following the TextCompetingLinks assessment receive a wrong instance of the paper (without the text). We need to fix that: either through making this assessment the last one in the list of assessments to be executed by SEO assessor or through making sure a new Researcher is created for this particular assessment, which does not interfere with the researcher needed for the other assessments.

nataliashitova:
## Summary

* 

* Fixes the bug where the paper would be reset to empty upon execution of the links statistics research.

## Test instructions

* In Free: link this branch to the `release/9.0` branch on `wordpress-seo` and check if
  * The Competing Links assessment works as expected (morphological forms are **not** matched, function words are not considered
  * The Internal Links and Outbound Links assessments work as expected (they use the same research).
  * The following assessments are correctly called and executed: TextImages, TextLength,  TitleKeywordAssessment, TitleWidth, UrlKeywordAssessment, UrlLength, urlStopWords.
* For Premium: link this branch to `trunk` on `wordpress-seo-premium` and check the same as for Free, but check also that morphological forms of words are recognized (i.e., you can use different forms of words in the keyphrase and in the link text and the Competing Links Assessment will still return a warning).

Fixes #1843

### DIFF
--- a/src/researches/getLinkStatistics.js
+++ b/src/researches/getLinkStatistics.js
@@ -69,18 +69,16 @@ const filterAnchorsContainingTopic = function( anchors, topicForms, locale ) {
  * @param {Array} anchors An array with all anchors from the paper.
  * @param {Array} keyphraseAndSynonyms An array with keyphrase and its synonyms.
  * @param {string} locale The locale of the paper.
- * @param {Researcher} researcher The morphological researcher.
+ * @param {Object} morphologyData The morphology data (regexes and exception lists) available for the language.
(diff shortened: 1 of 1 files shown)


manuelaugustin:
CR & acceptance done 👍

PR #1856, opened 2018-10-11
Fix "Uncaught TypeError: Cannot read property 'dgettext' of undefined"

### LINKED ISSUE #1855: Uncaught TypeError: Cannot read property 'dgettext' of undefined
For YoastSEO to run without errors. When intializing snippet first and afterwards the app (as in the readme example)

## What happened instead?
Got "Uncaught TypeError: Cannot read property 'dgettext' of undefined"

```
var snippetPreview = new SnippetPreview({
    targetElement: document.getElementById( "snippet" )
});

var app = new App({
    snippetPreview: snippetPreview,
    targets: {
        output: "output"
    },
    callbacks: {
        getData: function() {
            return {
                keyword: focusKeywordField.value,
                text: contentField.value
            };
        }
    }
});
```

## Technical info

* Platform: Standalone
* Platform version: /
* Yoast SEO version: 1.40.0

The problem seems to be that i18n in SnippetPreview doesn't get initialized. This happens because in app.js _i18n gets filled in instead of i18n. itself.

I'll add a PR with a fix. I'm not sure if this fix is completely correct since i'm not really home in the YoastSEO.js code.

kristoffeys:
Fixes #1855

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -288,7 +288,7 @@ var App = function( args ) {
 		app.*/
 		if ( this.snippetPreview.refObj !== this ) {
 			this.snippetPreview.refObj = this;
-			this.snippetPreview._i18n = this.i18n;
+			this.snippetPreview.i18n = this.i18n;
 		}
 	} else if ( args.hasSnippetPreview ) {
 		this.snippetPreview = createDefaultSnippetPreview.call( this );


igorschoester:
The `app` will get deprecated in the future. The `snippetPreview` got replaced by a React version in `yoast-components`. We are trying to move non-analysis functionality out of this repo.

Sorry that the readme is not up-to-date. We will update it after we've properly deprecated the app.

PR #1865, opened 2018-10-12
Iverse keyword distribution highlights

manuelaugustin:
## Summary

This PR can be summarized in the following changelog entry:

* n/a

## Relevant technical choices:

* Changes marking for the keyword distribution assessment. We now mark keyphrase occurrences (all matches that were counted in the keyword distribution assessment, so also partial matches) instead of marking sentences without keyphrase occurrences.

## Test instructions

This PR can be tested by following these steps:

* Im Premium, add a text and make sure that the keyphrase distribution assessment is triggered (see https://github.com/Yoast/YoastSEO.js/wiki/Scoring-SEO-analysis#11-keyphrase-distribution). Use keyphrases of multiple lengths (either <4 or >4 words). Also use synonyms. In the text, use these keyphrases/synonyms in various forms (e.g., plurals).
* Click on the marker button and make sure that all the keyphrase occurrences (also partial ones) get marked.

Fixes #1863

PR #1867, opened 2018-10-12
Relax keyphrase distribution assessment

### LINKED ISSUE #1863: Keyword distribution highlight should be inversed
Basically it should highlight exactly the same as the keyphrase density highlight.

### LINKED ISSUE #1868: Make the keyword distribution feedback less strict
At the moment, the requirements for the keyword distribution assessment are very strict if you want to get a green bullet. We should find a way to make the keyword distribution assessment less strict.

Proposition by @nataliashitova:

`The assessment score is calculated based on the Gini coefficient without additional punishment for not using all keyphrase terms often enough. This way we loose a few use cases, but at least we solve the problem with user frustration. If this relaxation of the rule does not help (enough), we can further adjust the actual boundaries.`

nataliashitova:
## Summary

* Adjusts the scoring schema of the keyphrase distribution to return overall higher scores.

* Deprecates previously introduced punishment for not using all keyphrase terms within a large window in the text
* Introduces a new punishment for having overall low scores per-sentence
* The PR also fixes the markings of the keyphrase.
* Partly duplicates PR #1865.

## Update:
* Deprecates previously introduced gini-coefficient-based estimation of the keyphrase distribution uniformity
* Instead, relies on the distance between topic usages (in sentences).
* For English and other languages with function words:
  * Short topic (<4 words): all content words should be matched for the sentence to reflect the topic.
  * Long topic ( >=4 words): more than half content words should be matched for the sentence to reflect the topic.
* For languages without function words:
  * All words should be matched for the sentence to reflect the topic.

## Test instructions

* Assessment is only in Premium.
* Take a few articles that rank well on Google and check if the assessment returns a Good score.
* Check that the markings work to highlight keyphrase and synonym words, not sentences without those.
* Change the keyphrase on these articles to some nonsense. The score should drop.

Fixes #1868 Fixes #1863

### DIFF
--- a/src/assessments/seo/KeyphraseDistributionAssessment.js
+++ b/src/assessments/seo/KeyphraseDistributionAssessment.js
@@ -3,8 +3,6 @@ import { merge } from "lodash-es";
 import Assessment from "../../assessment";
 import AssessmentResult from "../../values/AssessmentResult";
 import countWords from "../../stringProcessing/countWords";
-import Mark from "../../values/Mark";
-import addMark from "../../markers/addMark";
 
 /**
  * Returns a score based on the largest percentage of text in
@@ -67,7 +65,7 @@ class KeyphraseDistributionAssessment extends Assessment {
 
 		assessmentResult.setScore( calculatedResult.score );
 		assessmentResult.setText( calculatedResult.resultText );
-		assessmentResult.setHasMarks( calculatedResult.score > 0 && calculatedResult.score < 9 );
+		assessmentResult.setHasMarks( this._keyphraseDistribution.sentencesToHighlight.length > 0 );
 
 		return assessmentResult;
 	}
@@ -158,9 +156,7 @@ class KeyphraseDistributionAssessment extends Assessment {
 	 * @returns {Array} All markers for the current text.
 	 */
 	getMarks() {
-		return this._keyphraseDistribution.sentencesToHighlight.map( function( sentence ) {
-			return new Mark( { original: sentence, marked: addMark( sentence ) } );
-		} );
+		return this._keyphraseDistribution.sentencesToHighlight;
 	}
 
 	/**
--- a/src/assessments/seo/KeyphraseDistributionAssessment.js
+++ b/src/assessments/seo/KeyphraseDistributionAssessment.js
@@ -2,7 +2,7 @@ import { merge } from "lodash-es";
 
 import Assessment from "../../assessment";
 import AssessmentResult from "../../values/AssessmentResult";
-import countWords from "../../stringProcessing/countWords";
+import getSentences from "../../stringProcessing/getSentences";
 
 /**
  * Returns a score based on the largest percentage of text in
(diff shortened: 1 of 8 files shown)


src/researches/keyphraseDistribution.js
```diff
@@ -147,33 +148,30 @@ const step = function( maximizedSentenceScores, stepSize ) {
...
  *
- * @param {Array} sentences The sentences to analyze.
- * @param {Object} topicForms The topicForms of the paper.
- * @param {string} locale The locale of the paper.
- * @param {number} stepSize The number of sentences that should be in every step to average over.
+ * @param {Array} sentenceScores The array of sentenceScores to calculate penalty based on.
  *
- * @returns {Array} The scores per portion of text.
+ * @returns {number} The additional penalty for not using all keyphrase terms.
  */
```
hansjovis: Function does not use a window anymore, so the function description can be adapted to reflect that.

igorschoester:
First of all, it is not very clear to me what it needs to be. Just more relaxed is hard to test. Therefore, I tested more on the lines of can I see all the scores instead of it changed.

Acceptance 🚧 

- On a large text without any keyphrase in the text, I get the `Good job!` result. This should be the result: `Include your keyphrase or its synonyms in the text so that we can check keyphrase distribution.`.
- With 1 keyphrase at the very start I get the `Uneven. Some parts of your text do not contain the keyphrase or its synonyms. Distribute them more evenly.` result. This should probably be `Very uneven. Large parts of your text do not contain the keyphrase or its synonyms. Distribute them more evenly.`.

src/assessments/seo/KeyphraseDistributionAssessment.js
```diff
@@ -32,8 +30,8 @@ class KeyphraseDistributionAssessment extends Assessment {
 
 		const defaultConfig = {
 			parameters: {
-				goodDistributionScore: 0.4,
-				acceptableDistributionScore: 0.6,
+				goodDistributionScore: 30,
```
manuelaugustin: In the current version this is set to `40`. I think it might be a good idea to keep this 40 in order to keep continuity in the boundaries.

src/assessments/seo/KeyphraseDistributionAssessment.js
```diff
@@ -158,20 +156,18 @@ class KeyphraseDistributionAssessment extends Assessment {
...
+	 * Checks whether the paper has a text with at least 10 words and a keyword.
 	 *
 	 * @param {Paper} paper The paper to use for the assessment.
 	 *
-	 * @returns {boolean} True when there is a keyword and a text with 200 words or more.
+	 * @returns {boolean} True when there is a keyword and a text with 10 words or more.
 	 */
 	isApplicable( paper ) {
-		return paper.hasText() && paper.hasKeyword() && countWords( paper.getText() ) >= 200;
+		return paper.hasText() && paper.hasKeyword() && getSentences( paper.getText() ).length >= 10;
```
manuelaugustin: Unless all of your sentences are very long, this will make the assessment also trigger on short texts. Currently we trigger when there are 200 + words. Maybe increase to 15 sentences?

manuelaugustin:
CR & acceptance done 👍 

Also changed `is applicable` to at least 15 sentences (as discussed).

PR #1881, opened 2018-10-16
Fix bug with exact matches in URL Keyphrase assessment

### LINKED ISSUE #1874: Keyword in url doesn't work well with exact match keyphrases
When my keyphrase is `"cats and dogs"` (exact match) and my slug is `cats-and-dogs` I get the feedback `Keyphrase in slug: (Part of) your keyphrase does not appear in the slug. Change that!` I expect to get a green bullet in that case. When I change the keyphrase to `cats and dogs` (normal matching), I get a good result for the assessment.

nataliashitova:
## Summary

* Fixes the bug where the exact matching would not work for slug.

*

## Test instructions

* Create a multi-word focus keyphrase, e.g. `dogs and cats`.
* Check if the url keyword assessment works [as expected](https://github.com/Yoast/YoastSEO.js/wiki/Scoring-SEO-analysis#9-url-keyword-assessment).
* Add quotation marks around the focus keyphrase, e.g., `"dogs and cats"`.
* Check that 
  * If the slug contains an exact match, e.g. `dogs-and-cats`, the assessment returns a good score
  * If the slug does not contain an exact match, e.g., `dog-and-cat`, the assessment returns an okay score

Fixes #1874

### DIFF
--- a/spec/assessments/UrlKeywordAssessmentSpec.js
+++ b/spec/assessments/UrlKeywordAssessmentSpec.js
@@ -65,4 +65,26 @@ describe( "A keyword in url count assessment", function() {
 		expect( assessment.getScore() ).toEqual( 9 );
 		expect( assessment.getText() ).toEqual( "<a href='https://yoa.st/33o' target='_blank'>Keyphrase in slug</a>: More than half of your keyphrase appears in the slug. That's great!" );
 	} );
+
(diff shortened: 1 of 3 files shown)


manuelaugustin:
CR & acceptance done 👍

PR #1882, opened 2018-10-16
Fix bug with exact matches in Title Keyword assessment

### LINKED ISSUE #1873: Keyword in title doesn't work well with exact match keyphrases
When my keyphrase is `"cats and dogs"` (exact match) it's impossible to get a green score.

When I use `cats and dogs` as my title, I get the feedback that it's not an exact match. In that case, it should actually return a green feedback (When I use `"cats and dogs"` as my title, it says that my keyphrase isn't at the beginning).

nataliashitova:
## Summary

* Fixes the bug where a keyphrase with suppressed morphology would not be matched at the beginning of the title.

*

## Test instructions

* Define a keyphrase, e.g., `dog` or `dogs and cats` and make sure that it works [as expected](https://github.com/Yoast/YoastSEO.js/wiki/Scoring-SEO-analysis#8-page-title-keyword-assessment)
* Add quotes around the keyphrase, e.g., `"dog"` or `"dogs and cats"` and make sure that
  * Only exact matches are found. E.g., keyphrase `"dog"` should not be matched in title `Dogs and cats`
  * The exact matches are found correctly. E.g., keyphrase `"dog"` is correctly found at the beginning of the title (and the good feedback is returned) in title `Dog toys` and is correctly found **not** at the beginning of the title (and the orange feedback is returned) in title `Cool dog toys`

Fixes #1873

### DIFF
--- a/spec/assessments/TitleKeywordAssessmentSpec.js
+++ b/spec/assessments/TitleKeywordAssessmentSpec.js
@@ -12,7 +12,7 @@ describe( "an assessment to check if the keyword is in the pageTitle", function(
 		} );
 		const assessment = new TitleKeywordAssessment().getResult(
 			paper,
-			Factory.buildMockResearcher( { exactMatch: false, allWordsFound: false, position: -1 } ),
+			Factory.buildMockResearcher( { exactMatch: false, allWordsFound: false, position: -1, exactMatchKeyphrase: false } ),
 			i18n );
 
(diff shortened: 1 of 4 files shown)


manuelaugustin:
CR & acceptance done 👍 

I've made some small changes to the feedback strings to make the feedback more accurate when you don't have an exact match.

PR #1890, opened 2018-10-18
Meta description keyword assessment now only looks at keyphrase counts in entire description.

hansjovis:
## Summary
**Note:** tests need to be adapted to the new scoring scheme.

* Changes the meta description keyphrase assessment to only look at the amount of matches in the full description, instead of a combination of per sentence and full description.

* N/A

## Test instructions

* Make sure that Premium is active and running.
* Make a new post.
* Enter a keyphrase and one or more synonyms.
* Set the meta description to:
     * A text containing **none** of the words in the keyphrase (or its synonyms):
       * This should give a BAD result:
`The meta description has been specified, but it does not contain the keyphrase. Fix that!`
    * A text containing all of the words in the keyphrase (or its synonyms) **once**:
       * This should give a GOOD result: 
`Keyphrase or synonym appear in the meta description. Well done!`
     * A text containing all of the words in the keyphrase (or its synonyms) **twice**:
        * This should give the same GOOD result as the one above.
   * A text containing all of the words in the keyphrase (or its synonyms) **thrice or more**:
       * This should give a BAD result: 
`The meta description contains the keyphrase 3 times, which is over the advised maximum of 2 times. Limit that!`

Fixes https://github.com/Yoast/wordpress-seo/issues/11303

### DIFF
--- a/src/researches/metaDescriptionKeyword.js
+++ b/src/researches/metaDescriptionKeyword.js
@@ -1,5 +1,16 @@
-import getSentences from "../stringProcessing/getSentences";
-import { findTopicFormsInString } from "./findKeywordFormsInString";
+import matchWords from "../stringProcessing/matchTextWithArray";
+
+/**
+ * Counts the number of keyphrase matches,
+ * where a match is defined as a set containing all keywords.
+ *
+ * (E.g. for keyphrase "key word match",
+ * "key word match. key word. match. match." is counted as two matches: {"key","word","match"} x2 + one incomplete set {"match"}).
+ */
+const keyphraseMatchCount = function( description, keyphraseForms, locale ) {
+	const keywordMatchCounts = keyphraseForms.map( keywordForms => matchWords( description, keywordForms, locale ).count );
+	return Math.min( ...keywordMatchCounts );
+};
 
 /**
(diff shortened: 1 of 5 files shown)


src/researches/metaDescriptionKeyword.js
```diff
@@ -1,35 +1,87 @@
...
 /**
- * Matches the keyword in the description if a description and keyword are available.
- * Returns null if no description is specified in the given paper.
+ * Replaces found keyword forms in the given description.
+ *
+ * @param {string} description the description to remove the matched keyword forms from.
+ * @param {Object[]} matchedKeywordForms the matched keyword forms to remove from the description.
+ * @returns {string} the description with the keywords removed.
+ */
+const replaceFoundKeywordForms = function( description, matchedKeywordForms ) {
```
nataliashitova: From what I understood, this function replaces all found matches of all found words, regardless of whether a full match was found or not. This means that in case:
```
keyphrase: "vacation in Spain",
synonyms: "vacation in Iberia",
description: "An awesome meta description about vacations in Iberia."
```
"vacation" will already be stripped after keyphrase search and in the end neither synonyms nor keywords would be matched.
Is that right?

manuelaugustin:
Acceptance done 👍

PR #1898, opened 2018-10-22
Fix tokenizer errors and tokenize < in getSentences.

### LINKED ISSUE #840: Unable to tokenize '<'

The assessor not to break when encountering '<'.
### What happened instead?

schermafbeelding 2016-07-27 om 15 26 29
'>' seems to work fine (but is counted as a word, see #835 ).

Type in '<', look at the console.
### Technical info
- Platform: Standalone
- Yoast SEO version: 3.4

herregroen:
## Summary

* Fixes assessments failing when using the less-than sign in content.

* Always catch tokenizer.end() calls as errors there will severely disrupt the worker.
* Add the same regex as used in the `htmlBlockTokenizer` to the `sentenceTokenizer`.

Changes from @hansjovis :
* Refactors the `createTokenizer` and `tokenizeSentences` functions to make them reusable. These are used to tokenize the less-than sign match in a second pass. 
* Renames some local variables to make sure that they do not overshadow file-level declarations (e.g. `tokens` and `sentenceTokenizer`).
* Refactors some variables from `var` to `const` or `let`.

## Test instructions

* Write a beautiful post with great assessments.
* Add a less-than sign ( `<` ) to your post.
* You should still have great assessments.

For checking whether sentences are correctly found and returned, and you have access to the code, you can add these lines to `src/stringProcessing/getSentences.js` (on line 378, at the bottom of the file, before returning the sentences):
```js
console.log( sentences );
```

Fixes #840.


manuelaugustin:
I think there are some issues with this solution.

The regex `/^<[^><]*$/` will match everything from a `<` to the next `<` or `>`.  This means that `A sentence. A < sentence. Another sentence.` will be parsed into the following sentences:
- `"A sentence."`
- `"A < sentence. Another sentence."` 🚫 This behavior is problematic, because it will lead to long sentences that will distort our analysis results.

I tried to change that regex to a regex that will split up sentences with `<` in a more faithful way. The regex I used was `^<[^><"+.+?!;+<\\(\\)\\[\\]*$`. This regex will split up sentences containing `<` correctly, but it leads to problems with html tags. For an example, see this screenshot:
screen shot 2018-10-24 at 13 24 39

The correct way of splitting up this image would look like this:
screen shot 2018-10-24 at 13 26 20

The problem is that in `core.js` will recognize `greater-than-sign-content` rule is applicable when encountering the first `<`. It will then look for the longest possible string it can build by applying that rule. This will continue until it encounters the `.`.

So the problem is that for sentence segmentation, we want a rule that produces units that are as short as possible. However, this is problematic when dealing with html tags.

src/stringProcessing/getSentences.js
```diff
@@ -22,6 +22,7 @@ var newLines = "\n\r|\n|\r";
 var fullStopRegex = new RegExp( "^[" + fullStop + "]$" );
 var sentenceDelimiterRegex = new RegExp( "^[" + sentenceDelimiters + "]$" );
 var sentenceRegex = new RegExp( "^[^" + fullStop + sentenceDelimiters + "<\\(\\)\\[\\]]+$" );
+var greaterThanContentRegex = /^<[^><]*$/;
```
herregroen:
```suggestion
var greaterThanContentRegex = new RegExp( "^<[^" + fullStop + sentenceDelimiters + "><\\(\\)\\[\\]]*$" );
```

src/stringProcessing/getSentences.js
```diff
@@ -14,283 +13,368 @@ import { normalize as normalizeQuotes } from "../stringProcessing/quotes.js";
...
+		} catch ( e ) {
+			console.error( "Tokenizer end error:", e, e.tokenizer2 );
+		}
+	}
 
-	var sliced;
+	/**
+	 * Returns an array of sentences for a given array of tokens, assumes that the text has already been split into blocks.
+	 *
+	 * @param {Array} tokenArray The tokens from the sentence tokenizer.
```
igorschoester:
```suggestion
	 * @param {Object[]} tokenArray The tokens from the sentence tokenizer.
```

src/stringProcessing/getSentences.js
```diff
@@ -14,283 +13,368 @@ import { normalize as normalizeQuotes } from "../stringProcessing/quotes.js";
...
+	 * @param {Boolean} [trimSentences=true] Whether to trim the sentences at the end or not.
```
igorschoester:
```suggestion
	 * @param {boolean} [trimSentences=true] Whether to trim the sentences at the end or not.
```

src/stringProcessing/getSentences.js
```diff
@@ -14,283 +13,368 @@ import { normalize as normalizeQuotes } from "../stringProcessing/quotes.js";
...
+	 *
+	 * @returns {Array<string>} A list of sentences.
```
igorschoester:
```suggestion
	 * @returns {string[]} A list of sentences.
```

igorschoester:
CR 👍

manuelaugustin:
I've been acceptance testing with an eye on a log of the output of `getSentences`.

I've observed somewhat strange behavior in some edge cases:
When I type in a string ending with a `<` and nothing else, that string gets parsed as `<undefined`.

It's probably an edge case, but this slightly strange behavior should be investigated a bit more.

Note: this was an observation from a quick acceptance test; this PR still needs a bit more extensive testing before it can be merged.

hansjovis:
> I've observed somewhat strange behavior in some edge cases:
When I type in a string ending with a < and nothing else, that string gets parsed as <undefined.
This is fixed.

However, I noticed another, more serious bug:
`"Hey 40 < 50. However, 40 > 50."` should be parsed as `[ "Hey 40 < 50.",  "However, 40 > 50." ]` but is parsed as `["Hey 40 < 50.", "However, 40", "> 50."]`.
This is because the `smaller-than-sign-content` regex matches this text (until the '>'):
`< 50. However, 40`.

I added a test for this case and am now fixing this.

hansjovis:
It should be fixed now 🙂 .

igorschoester:
Changes look good to me 👍

manuelaugustin:
Acceptance 👍

PR #1902, opened 2018-10-22
Optimize dynamically created regex, possibly fixing Chrome70 crash

CarloCannas:
## Summary

* Optimize dynamically created regex

The current code produces horribly long regular expression because the word boundary pattern is repeated for each single word. The list of words [may contain hundreds of words](https://github.com/Yoast/YoastSEO.js/blob/v1.41.0/src/stringProcessing/directPrecedenceException.js#L45), leading to regular expressions tens of thousands characters long.
There should be no need to repeat the word boundary pattern: applying it globally once should be enough.

I have however a few questions:

- Should both createRegexFromArray.js and createRegexFromDoubleArray.js refactored in a single file to avoid code duplication? Is it worth it?
- Why don't I get a failure for https://github.com/Yoast/YoastSEO.js/blob/5799f87432fda39e08f9dd2f71ab0d2bf5f3cebe/spec/stringProcessing/createRegexFromArray.js when I run `npm test`? Is it because I don't have the `premium-configuration` module?
- I had to fix `syllableCountSpec.js` because it had an hard coded check of the actual regexp syntax used, is that really needed? I think it might be more useful to test some configurations a little bit more complex, for example with more fragments than one.

It looks like this optimization fixes https://github.com/Yoast/wordpress-seo/issues/11356

### DIFF
--- a/spec/helpers/syllableCountSpec.js
+++ b/spec/helpers/syllableCountSpec.js
@@ -18,7 +18,7 @@ describe( "A test for creating a language syllable regex", function() {
 		expect( languageSyllableRegex.getRegex() ).toEqual( /(a)/gi );
 	} );
 
-	it( "creates an language syllable regex with a +1 multiplier", function() {
+	it( "creates an language syllable regex with a -2 multiplier", function() {
 		var mockSyllables = {
 			fragments: [ "ee" ],
 			countModifier: -2,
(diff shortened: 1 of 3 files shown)


manuelaugustin:
Hey Carlo, thanks a lot for your PR! Looks great at a first glance! We'll look into it a bit further and see if everything works as expected.

Regarding your questions:
1) Regarding these two files: it would be a bit nicer to combine this functionality in one file, but given the effort it would take it's probably not worth investing too much time.
2) The reason you don't get a failure here is probably because `toMatch` only compares the results of the regexes, not the exact regexes themselves. So it's a positive sign that this test isn't failing because it means that your improved regex produces the same result as the original regex.
3) For the moment we think that these specs cover the things we want to test, but if you have concrete ideas for improvements you can always let us know and we'll be happy to review them!

Again, thanks a lot for your help!

igorschoester:
Thanks for the input @CarloCannas. Closing this PR in favor of https://github.com/Yoast/YoastSEO.js/pull/1910.

PR #1918, opened 2018-10-26
Fix reference to empty self

### LINKED ISSUE #1917: Make a more robust check instead of just using `self`
Because the `Delightful Downloads` plugin sets `self = this` without having "strict" mode enabled, it will overwrite the global `window.self` to the object this is called in.

See: https://github.com/Yoast/wordpress-seo/issues/11403

This is being done at the following location:
https://github.com/Yoast/YoastSEO.js/blob/develop/src/helpers/shortlinker/singleton.js#L13

@atimmer has the approach for a solution ready.

nataliashitova:
## Summary

* Fixes the bug where the Yoast SEO analysis would error if used together with the DelightfulDownloads plugin.

* 

## Test instructions

* Check if regular operations are performed without error in Classic and Gutenberg:
  * Making a new post / page, saving
  * Opening / editing an existing post
  * Check if Yoast SEO analysis makes sense
* Install the [DelightfulDownloads plugin](https://wordpress.org/plugins/delightful-downloads/) and activate it
* Check if the regular operations are still performed without error in Classic and Gutenberg.

Fixes #1917 Fixes https://github.com/Yoast/wordpress-seo/issues/11403

### DIFF
--- a/src/helpers/shortlinker/singleton.js
+++ b/src/helpers/shortlinker/singleton.js
@@ -1,8 +1,16 @@
 import Shortlinker from "./Shortlinker";
 
-// Expose global variable.
-self.yoast = self.yoast || {};
-self.yoast.shortlinker = null;
+// Set global scope.
+let globalScope;
+
+if ( typeof window === "undefined" ) {
+	globalScope = self;
+} else {
+	globalScope = window;
+}
+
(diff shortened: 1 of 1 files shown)


manuelaugustin:
Closed in favor of https://github.com/Yoast/YoastSEO.js/pull/1923

PR #1406, opened 2018-02-09
Implement single h1 assessment

manuelaugustin:
## Summary

This PR can be summarized in the following changelog entry:
* Adds an assessment that checks whether the body of the text contains an H1 at any position other than the very beginning.

## Relevant technical choices:

* We don't know whether a certain WordPress theme or custom post type renders the title as an H1.  For this reason, having an H1 in the body might still be a valid choice. However, in that case the H1 would only be valid at the very beginning of the body. This assessment, therefore, only produces a warning if there is an H1 anywhere in the body except at the very beginning. 

## Test instructions

This PR can be tested by following these steps:

**Browserified**
* Check out the branch.
* Use the browserified example. Don't forget to run a `grunt build:js`.
* Add a text that contains one or multiple H1 headings in the body. Make sure that you get an appropriate feedback string. No feedback is shown when there is no H1 or the H1 is at the very beginning of the body. In that position, the H1 is regarded as acceptable.
* Make sure you can mark incorrect H1s by using the marking button. Again, H1s at the very beginning of the body won't be marked as these are regarded as acceptable.
* Make sure this functionality also works when cornerstone is switched on.

**WordPress**
* Check out the branch `feature/recalibration` in `wordpress-seo`.
* Run yarn install in the plugin's directory.
* Run `yarn add "git+https://github.com/Yoast/yoastseo.js#stories/1298-implement-single-H1-assessment"` in the plugin's directory.
* Navigate to `node_modules/yoastseo` and run `yarn install` and `grunt build:js`.
* Navigate back to the plugin's directory and run `grunt build:js`.
* Test the same functionality as described in the test instructions for the browserified example.

Fixes #1298

Issue #1913, opened 2018-10-24
Recalibration beta phase: overview


manuelaugustin:
## Context
When rolling out the code from recalibration, we want initially do that in a beta phase. In this beta phase, users should be able to switch back and forth between the recalibrated version and the old version. We want to be able to make changes to the new recalibration code and make this code available to users immediately. 

## Specs
**Team Lingo**
- New `recalibratedSEOAssessor` + all related assessors (`relatedKeyphrasesAssessor` etc.)
- Duplicate all assessments and researches that need to diverge from existing ones and that cannot be reused in their current shape.
- Have a way to conditionally load `recalibratedSEOAssessor` instead of `SEOAssessor`.

**Team Plugin**
- Plugin team should provide the condition, which will need to be set based on toggle, that switches an option on and off.
- Add a notification to point to the feature toggle 

**Team Platform**
- Platform team should provide an endpoint where the JavaScript can be downloaded.
- Together with platform team figure out a way to automatically deploy changes to the JavaScript.
- The JavaScript download will not be authenticated (unlike the premium config). Users don’t need a Yoast account. Downloads should be cached in Cloudflare.

nataliashitova:
Question about code reduplication to @omarreiss 

If we only need to be able to adjust assessment boundaries or add|remove assessments from assessor, this can simply be done within one configuration file. Duplicating all assessments and assessors will be so difficult to maintain!
If we also want to have access to two versions of researches at the same time, we could pass a version value (`regular` or `recalibration`) to the `researcher` and make sure all researches return the same type of data no matter which version of the research code is used.

Are there other reasons why reduplication is necessary?

omarreiss:
If the behavior of two assessments diverges from the current implementation to the recalibration, then it makes sense to duplicate. If only the config differs, then you can solve it in the config layer. I don't mind the implementation so much. I'm just saying don't be afraid to maintain two versions of something as long as we are in beta.

manuelaugustin:
This project has been completed.

PR #1936, opened 2018-10-31
Add singleH1Assessment, remove urlLength and urlStopwords in Recalibration

### LINKED ISSUE #1926: Add single H1 assessment
- Add Single H1 assessment to the assessors used for recalibration. The functionality can be found on the old recalibration branch.
- Adjust the feedback strings to follow the new format: "Assessment Title: explanation. Call to action!"
- Also add the assessment to the taxonomy assessor (not to related keywords assessor).

### LINKED ISSUE #1927: Remove URL length assessment
- The URL length assessment needs to be removed from all assessors used for recalibration.

### LINKED ISSUE #1928: Remove stopwords in slug assessment
The URL stopwords in slug assessment needs to be removed from all assessors used for recalibration.

nataliashitova:
## Summary

* Deprecates two SEO assessments that check the length of the URL and whether stop words are used in the URL
* Adds an assessment that checks if the article has spurious uses of `<h1>`.

* The SingleH1Assessment was transferred from the PR where it was introduced #1406. It makes use of the current HTML parser, making it run another time to pick up `<h1>` tags. @atimmer do you find it acceptable to run HTML parser in this way? Or would you suggest another solution?

## Test instructions

* Use the new example to compare regular and recalibration analysis.
* In the regular analysis, check if everything works as before:
  * Add a text with two H1s, no additional assessments should appear
  * Add a slug with a stop word, a urlStopWords assessment should trigger
  * Add a long slug, a urlLength assessment should trigger
* In the recalibration analysis:
  * Add a text with two H1s, a `Single title` assessment should trigger.
  * Add a text with a H1 at the beginning of the text. The `Single title` assessment shouldn't trigger.
  * Add a text with an H1 anywhere but the beginning of the title. The `Single title` should trigger.
  * Add a slug with a stop word, a urlStopWords assessment should **not** trigger
  * Add a long slug, a urlLength assessment should **not** trigger

Fixes #1926 Fixes #1927 Fixes #1928

### DIFF
--- a/spec/researches/h1sSpec.js
+++ b/spec/researches/h1sSpec.js
@@ -7,9 +7,22 @@ describe( "Gets all H1s in the text", function() {
 		expect( h1s( mockPaper ) ).toEqual( [] );
 	} );
 
+	it( "should return one object when there is one H1", function() {
+		const mockPaper = new Paper( "<h1>first h1</h1>some content<h2>content h2</h2>" );
+		expect( h1s( mockPaper ) ).toEqual( [ { tag: "h1", content: "first h1", position: 0 } ] );
+	} );
+
 	it( "should return all H1s in the text", function() {
 		const mockPaper = new Paper( "<h1>first h1</h1><p>not an h1</p><h1>second h1</h1><h2>not an h1</h2>" );
 
+		expect( h1s( mockPaper ) ).toEqual( [
+			{ tag: "h1", content: "first h1", position: 0 },
+			{ tag: "h1", content: "second h1", position: 2 },
+		] );
+	} );
+
+	it( "should find H1 within division tags", function() {
+		const mockPaper = new Paper( "<div><h1>first h1</h1></div><div><p>blah blah</p></div><div><h1>second h1</h1></div>" );
 		expect( h1s( mockPaper ) ).toEqual( [
 			{ tag: "h1", content: "first h1", position: 0 },
 			{ tag: "h1", content: "second h1", position: 2 },
--- a/spec/researches/h1sSpec.js
+++ b/spec/researches/h1sSpec.js
@@ -1,0 +1,18 @@
+import h1s from "../../src/researches/h1s.js";
+import Paper from "../../src/values/Paper.js";
+
+describe( "Gets all H1s in the text", function() {
+	it( "should return empty when there is no H1", function() {
+		const mockPaper = new Paper( "some content<h2>content h2</h2>" );
+		expect( h1s( mockPaper ) ).toEqual( [] );
+	} );
+
+	it( "should return all H1s in the text", function() {
+		const mockPaper = new Paper( "<h1>first h1</h1><p>not an h1</p><h1>second h1</h1><h2>not an h1</h2>" );
+
+		expect( h1s( mockPaper ) ).toEqual( [
+			{ tag: "h1", content: "first h1", position: 0 },
+			{ tag: "h1", content: "second h1", position: 2 },
+		] );
+	} );
+} );
--- a/src/assessments/seo/SingleH1Assessment.js
+++ b/src/assessments/seo/SingleH1Assessment.js
@@ -1,5 +1,6 @@
-import map from "lodash/map";
-import merge from "lodash/merge";
+import { map } from "lodash-es";
+import { merge } from "lodash-es";
+import { isUndefined } from "lodash-es";
 
 import Assessment from "../../assessment.js";
 import { createAnchorOpeningTag } from "../../helpers/shortlinker";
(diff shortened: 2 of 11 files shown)


src/assessments/seo/SingleH1Assessment.js
```diff
@@ -0,0 +1,152 @@
...
+		return ( this._h1s[ 0 ].position === 0 );
+	}
+
+	/**
+	 * Checks whether there are superfluous (i.e., non-title) H1s in the text.
+	 *
+	 * @returns {boolean} Returns true if there are superfluous H1s in the text.
+	 */
+	bodyContainsSuperfluousH1s() {
+		return ( this._h1s.length >= 1 );
```
hansjovis: This function only returns the correct result when the `h1` at the first position has been discarded (as is done in `getResult`).
It may be better to rewrite this function, and remove the `this._h1s.shift();` all together.
Like this:
```
 // There can be only one H1, at the beginning of the text. 
return !( firstH1AtBeginning() && this._h1s.length === 1) 
```
This would move all the logic to this function and makes getting an overview of the logic a bit better 😉 .

src/cornerstone/seoAssessor.js
```diff
@@ -33,75 +34,142 @@ const CornerstoneSEOAssessor = function( i18n, options ) {
...
-		new MetaDescriptionKeywordAssessment(),
-		new MetaDescriptionLength( {
-			scores:	{
-				tooLong: 3,
-				tooShort: 3,
-			},
-		} ),
-		new SubheadingsKeyword(
-			{
+	if ( process.env.YOAST_RECALIBRATION === "enabled" ) {
```
hansjovis: There is some repeated code in this function. Not every assessment gets replaced for recalibration (yet 😛 ).
Perhaps we can have a list with default assessments, and then an if-else to add only the assessments that are different (f you catch my drift). Would save some lines and make it more clear what assessments have changed 🙂 .

src/seoAssessor.js
```diff
@@ -31,25 +32,46 @@ const SEOAssessor = function( i18n, options ) {
...
-		new OutboundLinks(),
-		new TitleKeywordAssessment(),
-		new InternalLinksAssessment(),
-		new TitleWidth(),
-		new UrlKeywordAssessment(),
-		new UrlLength(),
-		urlStopWords,
-		new FunctionWordsInKeyphrase(),
-	];
+	if ( process.env.YOAST_RECALIBRATION === "enabled" ) {
```
hansjovis: See my other comment above.

src/taxonomyAssessor.js
```diff
@@ -24,20 +25,36 @@ const TaxonomyAssessor = function( i18n ) {
...
-		taxonomyTextLengthAssessment,
-		new TitleKeywordAssessment(),
-		new PageTitleWidthAssessment(),
-		new UrlKeywordAssessment(),
-		new UrlLengthAssessment(),
-		urlStopWordsAssessment,
-		new FunctionWordsInKeyphrase(),
-	];
+	if ( process.env.YOAST_RECALIBRATION === "enabled" ) {
+		this._assessments = [
```
hansjovis: See my other comment 😉 .

spec/researches/h1sSpec.js
```diff
@@ -0,0 +1,18 @@
...
+	} );
+
+	it( "should return all H1s in the text", function() {
+		const mockPaper = new Paper( "<h1>first h1</h1><p>not an h1</p><h1>second h1</h1><h2>not an h1</h2>" );
+
+		expect( h1s( mockPaper ) ).toEqual( [
+			{ tag: "h1", content: "first h1", position: 0 },
+			{ tag: "h1", content: "second h1", position: 2 },
+		] );
+	} );
```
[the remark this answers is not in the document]
nataliashitova: Fixed

src/cornerstone/seoAssessor.js
nataliashitova: Upon further investigation together with @hansjovis and @manuelaugustin it was decided to stick with the original version as it allows fast transition between beta-recalibration and forever-recalibration. Also, as per https://github.com/Yoast/YoastSEO.js/issues/1913#issuecomment-432659703 we should not be afraid of duplicate code.

src/seoAssessor.js
nataliashitova: See my answer above.

src/taxonomyAssessor.js
nataliashitova: Same.

manuelaugustin:
Acceptance done 👍 

I just made some small changes to the feedback string.

As discussed, we should refactor the single title assessment to make it less resource-intensive in a future iteration, see https://github.com/Yoast/YoastSEO.js/issues/1964

PR #1953, opened 2018-11-04
Examples webpack work

igorschoester:
## Summary

This PR can be summarized in the following changelog entry:

* N/A

## Relevant technical choices:

- Fixed some spacing irregularities with the paper UI, columns and inputs.
- Fixed Jed error from `yoast-components`.
- Fixed `locale` not being saved to the configuration.
- Added webpack alias to the `YoastSEO.js` `src` folder.
- Added `ButtonContainer`, `HeadingContainer` and `HorizontalContainer`.
- Added `fullTestTexts`'s papers as input with the help of `react-select`.
- Added overall scores.
- Changed raw markings: moved inside a collapsible to save screen space.
- Changed control buttons to `yoast-components`'s `IconButton`.
- Changed `StatusBox` to use flexbox.
- Removed TypeScript loader.
- Removed the H1, this is not an actual web page; it took up valuable screen space.

## Test instructions

This PR can be tested by following these steps:

* Go to `examples/webpack`.
* Install dependencies: `yarn`.
* Run it: `yarn start`.
* Check out the changes (see the list in the choices section).
* Ensure other things are still functioning properly.

PR #1954, opened 2018-11-04
1942 relevant words example

### LINKED ISSUE #1942: Clean up examples: relevant-words-example
Please integrate the `relevant-words-example` example into the `webpack` example. Remove the `relevant-words-example` afterward.
Please do a double check on the `wordRelevance` example. It should contain the same information as you now included in the `webpack` example. If this is the case, also remove the `wordRelevance` example.

igorschoester:
## Summary

* N/A

* ~~Branched from #1953. This PR contains 2 commits:~~
  - ~~Adding: 0fc1fe51a062dd40f0d282c806d57b6fbf8fa53e~~
  - ~~Cleanup: 6f599dd33c8a02cdc787b00417cf4d2b53e6dd3c~~
* Added relevant words from the browserified example to the webpack example.
* Added `react-table` for the UI.

## Test instructions

* Go to `examples/webpack`.
* Install dependencies: `yarn`.
* Run it: `yarn start`.
* Check out the relevant words section (possibly compare it to the browserified).

Requires #1953 
Fixes #1942


examples/webpack/src/utils/calculateRelevantWords.js
```diff
@@ -0,0 +1,51 @@
...
+		const output = {
+			word: word.getCombination(),
+			relevance: formatNumber( word.getRelevance() ),
+			length: word._length,
+			occurrences: word.getOccurrences(),
+			multiplier: formatNumber( word.getMultiplier( word.getRelevantWordPercentage() ) ),
+			relevantWordPercentage: formatNumber( word.getRelevantWordPercentage() ),
+		};
+
+		if ( word._length === 1 ) {
```
hansjovis: This could be made into a ternary statement.
E.g.
```js
output.lengthBonus = word._length === 1 ? "" : WordCombination.lengthBonus[ word._length ];
```

manuelaugustin:
Acceptance 🚧

* There's a problem with triggering the analysis: when I add a text for the first time (e.g., after first load or after pressing `clear inputs`) the relevant words section doesn't automatically show results. Only after hiding/showing the content of the collapsible once the results will show. After triggering the analysis for the first time that way, the results will automatically update as expected.
* There's a problem with the rows: the number of rows doesn't seem to update after the initial analysis. So if your initial analysis only has one row (e.g. if you enter the text `word word`), the results will keep showing just one row. Even if you set a text that would have more prominent word combinations (e.g. by selecting one of the sample papers) only the first row will be displayed. Conversely, if you do it the other way around (first setting the text with many word combinations), you get a lot of empty rows when you then change the text to something like `word word`.

Apart from that, all the basic functionality seems to work as expected. Great that we can finally use the relevant words example again!

igorschoester:
Found that `react-table` does not update when you alter the default page size. This was my way of disabling the pagination and removing the empty rows. Therefore, there now is pagination.
Before finding the cause I tried updating dependencies to see if that would help. Nothing broke, so I included it 😉 

- Reverted table pagination. Now with a default of 100 rows.
- Added minimum rows of 3 to make room for the `No rows found` text.
- Moved the compare outside of the react component into the redux connector.
- Updated dependencies.

manuelaugustin:
Acceptance 👍

Looks perfect now! ✨

PR #1972, opened 2018-11-11
Use morphology in devtool

### LINKED ISSUE #1961: Use morphology in YoastSEO.js development tool
Currently we can't use morphological forms in the YoastSEO.js development tool. We should implement a way to use these. There should be a toggle that lets you switch between using/not using morphology data.

igorschoester:
## Summary

* N/A

~~Branched from `examples-webpack-work`. Only commit of this PR is 0746cd9d54bb92cfcd81e5bd531db3973747e8ff~~

* Added use morphology toggle. Which uses local data.
* Added options reducer and moved `isRelatedKeyphrase` there too. Meant for the data that should not be sent to the worker.
* Fixed the analyze control not respecting the related keyphrase option.

## Test instructions

* Go to `examples/webpack`.
* Install packages: `yarn`.
* Run the dev server: `yarn start`.
* Ensure the morphology toggle works. For example check out the difference in the keyphrase density when using different word forms of the keyphrase in the text.
* Ensure related keyphrase toggle still functions.
* Ensure the analyze control runs analyze or an analyze related keywords depending on the is related keyphrase state.

Requires #1953 
Fixes #1961


examples/webpack/src/utils/getMorphologyData.js
```diff
@@ -0,0 +1,21 @@
+let morphologyData = null;
+
+function loadLocalMorphologyData() {
```
hansjovis: Although it is pretty straightforward what these functions do, it would be nice to add some JSDoc just in case 😉.
I will add this myself, and move it to code review done afterwards.

examples/webpack/src/utils/getMorphologyData.js
```diff
@@ -0,0 +1,21 @@
...
+		// eslint-disable-next-line global-require
+		data = require( "../../../../premium-configuration/data/morphologyData.json" );
+	} catch ( error ) {
+		// Falling back to empty data.
+	}
+	return data;
+}
+
+
+export default function getMorphologyData() {
```
hansjovis: Same for this function.

manuelaugustin:
Acceptance done 👍

Works like a charm! It's great that we can now use the morphology data in the development tool, too!

PR #1979, opened 2018-11-13
Replace underscore with whitespace in URL keyphrase assessment

### LINKED ISSUE #1978: Strip underscore from URL in the URL keyword assessment
Right now, when a dash appears in the slug, it is replaced with a space in order to compare the slug to the keyphrase.

However, in some slugs (e.g. Wikipedia) the underscore is used as a word boundary instead of the dash. We should add the underscore to the characters that are stripped from the URL.

agnieszkaszuba:
## Summary

* Adds underscore to characters that are replaced with whitespace in URL in the keywordCountInURL function

## Test instructions

* In your local Wordpress environment, create a text with a URL that uses underscores as word boundaries (rather than dashes). 
* Enter a keyphrase that uses the same words as those in the slug
* Make sure that a green bullet is returned for the URL keyphrase assessment

Fixes #1978

### DIFF
--- a/spec/researches/keywordCountInUrlSpec.js
+++ b/spec/researches/keywordCountInUrlSpec.js
@@ -177,4 +177,11 @@ describe( "test to check url for keyword", function() {
 		researcher.addResearchData( "morphology", morphologyData );
 		expect( urlKeyword( paper, researcher ) ).toEqual( { keyphraseLength: 1, percentWordMatches: 0 } );
 	} );
+
+	it( "works with dash as word boundary un url", function() {
+		const paper = new Paper( "", { url: "cats_and_dogs", keyword: "cats and dogs" } );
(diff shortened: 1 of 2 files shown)


spec/researches/keywordCountInUrlSpec.js
```diff
@@ -177,4 +177,11 @@ describe( "test to check url for keyword", function() {
 		researcher.addResearchData( "morphology", morphologyData );
 		expect( urlKeyword( paper, researcher ) ).toEqual( { keyphraseLength: 1, percentWordMatches: 0 } );
 	} );
+
+	it( "works with dash as word boundary un url", function() {
```
hansjovis:
```suggestion
	it( "works with dash as word boundary in url", function() {
```

igorschoester:
Acceptance 👍

PR #1981, opened 2018-11-13
Remove unneeded functionWords argument from getRelevance call

IreneStr:
## Summary

* N/A

* As can be seen in https://github.com/Yoast/YoastSEO.js/blob/develop/src/values/WordCombination.js#L145, `getRelevance` takes no arguments. However, we passed `functionWords` to it in `getRelevantWords`.

## Test instructions

* Make sure no unit tests fail.

It's very clear from the code that the passed functionWords were not used, however if you want to test it very thourougly: 
* Make sure there are no unexpected results in the relevant words. Most importantly: no one-word combinations that are function words should show up in the relevant words list.


igorschoester:
CR & Acceptance 👍

PR #1991, opened 2018-11-19
Fix url-permalink and add another full-text spec

### LINKED ISSUE #1957: Fix url and permalink in specs full text tests
In `specs/fullTextTests/testTexts`. The paper data does not include `permalink` and the `url` contains the permalink data.

This should be:
- `url`: should contain only the slug.
- `permalink`: should contain only the base url.

Let's take the `englishPaper1.js` as an example:
- `url`: `future-of-voice-search`
- `permlink`: `https://yoast.com/future-of-voice-search`

This breaks the specs, so those will need to get looked at too!

nataliashitova:
## Summary

* n/a

* Fixes the confused arguments `url` and `permalink` in the full-text specs.
* Fixes specs for internal and external link counts (because of the confusion those were not counted correctly).
* Adds another English text to the full-text specs.

## Test instructions

* Check if all tests pass

Fixes #1957


igorschoester:
Acceptance ✅

PR #1992, opened 2018-11-19
1773 worker runresearch failsafe

### LINKED ISSUE #1773: Worker runResearch failsafe
The worker `runResearch` action does not have any failsafe for:

- when the researcher throws an error.
- when the researcher does not have the requested research.

Please add a `runResearch:failed` that handles this. See `customMessage` for inspiration.
And adapt the tests to match.

Requires #1774

hansjovis:
## Summary

* The browser console now shows descriptive error messages when triggering single researches via the webworker.

* The webworker returns an `runResearch:failed` message with a descriptive error message, but this error message is not handled yet in the plugin. In the mean time we log the error to the console.

## Test instructions

* Open a research that is called using the `runResearch`-method of the webworker in PHPStorm or another editor.
    * Like the `relevantWords` or `getLinks` researches.
* Edit the code in such away that it throws an error.
   * E.g. add
      ```js
      throw new Error( "This is an error" );
      ```
      In `relevantWords` or `getLinks` before the `return` statement.
* Open a post in WordPress with the premium plugin activated.
* Open the browser console.
* Notice that it shows a descriptive error in the console, with the given error message as defined above.
    * Error should be: 
       ```
        An error occurred after running the 'relevantWords' research.
	      Error: This is an error
       ```
    * Note that is shows an extra error (`Uncaught (in promise)`), since the error is returned from the webworker as well, but not handled yet by the plugin.

Follow the same steps as above, but:
* Edit the code such that it throws a random object.
    ```js
     throw { someData: 42 }
    ```
* The error in the console should read:
   ```
   An error occurred after running the 'relevantWords' research.
   ```
 
Fixes #1773


spec/helpers/formatStringSpec.js
```diff
@@ -0,0 +1,46 @@
...
+
+	it( "escapes regular expressions in parameter names", () => {
+		const string = "%[a-z]% %.*?%";
+		const map = {
+			"[a-z]": "World",
+			".*?": 123,
+		};
+		expect( formatString( string, map ) ).toEqual( "World 123" );
+	} );
+} );
```
igorschoester: Here are a couple more tests, a few more unhappy paths. 3 of theses do not pass.
The `ignores matches without format` I think should be fixed.
The `normal string with percent-signs` is a mix between the two.
Granted the `%%%` one is mean and should probably not be an issue. Although, one might argue you'd need to be able to set the token yourself to prevent it.

```suggestion
} );

	it( "ignores matches without format", () => {
		const string = "%ignore% %this%";
		const map = {
			ignore: "ignore the following:",
		};
		expect( formatString( string, map ) ).toEqual( "ignore the following: %this%" );
	} );

	it( "handles a normal string", () => {
		const string = "This is a normal string.";
		const map = {
			normal: "unused",
		};
		expect( formatString( string, map ) ).toEqual( "This is a normal string." );
	} );

	it( "handles a normal string with percent-signs in it", () => {
		const string = "Is this 1% of all the %-ages?";
		const map = {
			"%": "percent",
		};
		expect( formatString( string, map ) ).toEqual( "Is this 1% of all the %-ages?" );
	} );

	it( "handles multiple percent-signs in a row", () => {
		const string = "Can you handle %%% that?";
		const map = {
			"%": "percent",
		};
		expect( formatString( string, map ) ).toEqual( "Can you handle %percent% that?" );
	} );
```

src/helpers/formatString.js
```diff
@@ -0,0 +1,33 @@
+import { escapeRegExp } from "lodash-es";
+
+/**
+ * Formats a string with named parameters as defined in the given parameter mapping.
+ *
+ * E.g. `"Hello %par_1% and %par_2%"` plus the map `{ par_1: "world", par_2: "you!" }`
+ * gives: `"Hello world and you!"`.
+ *
+ * @param {string} string the string to be formatted.
```
igorschoester:
```suggestion
 * @param {string} string           The string to be formatted.
```

src/helpers/formatString.js
```diff
@@ -0,0 +1,33 @@
...
+ * @param {Object} formatMap the mapping in the form of parameter - value pairs.
```
igorschoester:
```suggestion
 * @param {Object} formatMap        The mapping in the form of parameter - value pairs.
```

src/helpers/formatString.js
```diff
@@ -0,0 +1,33 @@
...
+ * @param {string} [delimiter="%%"] the string used to delimit parameters in the to be formatted string.
```
igorschoester:
```suggestion
 * @param {string} [delimiter="%%"] The string used to delimit parameters in the to be formatted string.
```

src/helpers/formatString.js
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @returns {string} the formatted string.
```
igorschoester:
```suggestion
 * @returns {string} The formatted string.
```

src/worker/wrapTryCatchAroundAction.js
```diff
@@ -1,3 +1,37 @@
...
+ * when an analysis web worker action fails.
+ *
+ * @param {Logger}	logger					The logger instance to log with.
+ * @param {Error} 	error					The error to log.
+ * @param {Object}	payload					The action payload.
+ * @param {string} 	[errorMessagePrefix=""]	The prefix of the error message.
+ *
+ * @returns {string} the error message to send back.
+ */
+const handleError = function( logger, error, payload, errorMessagePrefix = "" ) {
```
igorschoester: I presume you moved this because of the complexity. However, CodeClimate now says the following:
```
Function `handleError` has a Cognitive Complexity of 6 (exceeds 5 allowed). Consider refactoring.
```

Do you see any other improvements to reduce complexity?

manuelaugustin:
CR of final commits & acceptance done 👍

PR #1976, opened 2018-11-12
Add Swedish full text tests

agnieszkaszuba:
## Summary

This PR can be summarized in the following changelog entry:

* Adds Swedish papers to the full text tests

## Relevant technical choices:

* 

## Test instructions

This PR can be tested by following these steps:

* Run the tests.
* Verify the tests pass.
* Run the example webpack: `cd examples/webpack&&yarn&&yarn start`.
* Verify the analyze spam in the example works.

Fixes #1973

PR #2013, opened 2018-11-26
Account for isApplicable in fullTextTests

### LINKED ISSUE #2010: Take isApplicable into account for full-text specs
* The full-text specs should take `isApplicable` into account when running the assessments.

## What happened instead?
* All assessments are run regardless of whether they are applicable or not. This leads to incorrect results, e.g. when assessments return a given default value.

nataliashitova:
## Summary

* Improves full text tests by taking `isApplicable` of the assessments into account

* Removes the `assessmentsToRun` functionality (introduced in https://github.com/Yoast/YoastSEO.js/pull/1976), which served the same purpose as `isApplicable`, but was local to the full text tests.

## Test instructions

* Run the test sequence with `yarn test` and observe that nothing fails.

Fixes #2010


hansjovis:
Acceptance test: ✅ 

- [x] All tests pass when running `yarn test`.

PR #2016, opened 2018-11-26
Don't strip numbers from text when marking keyphrases containing numbers

### LINKED ISSUE #2007: Numeric part of keyphrase gets removed from text
1. Go to create post in default classic editor.
2. Add the following text:

> When will WordPress 5.0 arrive?
> WordPress 5.0 was scheduled to arrive on Tuesday, November 27, 2018, but, as of November 22, the release has been put on hold with no new date set yet. There’s still a possibility that it will be delayed until after the holidays, which means we’ll see it arrive sometime in January. Time is short, which means you need to get busy getting ready.
> 
> Will my site break when I update to WordPress 5.0?
> There’s always a possibility of that happening, but the WordPress core team does its utmost best to prevent this. WordPress 5.0 is a major release, and you shouldn’t forget that an average site is made up of many moving parts — themes, plugins, meta boxes, custom code, etc. There are thousands and thousands of themes and plugins, and not everything works well together — a crooked combination of those could very well cause your site to crash. So, before you upgrade to WordPress 5.0 and activate the new editor, you should have thoroughly tested your configuration for faults.

3. Add Focus keyphrase "WordPress 5.0".
4. Switch to Visual editor.
5. Check analysis:
screenshot 2018-11-23 at 15 00 35

6. Toggle eye marker, see that "5.0" get removed from text in the most cases:
screenshot 2018-11-23 at 15 00 59

## Technical info

* Platform: WordPress 
* Platform version: 4.9.8
* Yoast SEO version: 9.2.1

manuelaugustin:
## Summary

* Fixes a bug that caused numbers to be stripped when marking a keyphrase containing a number, e.g. `Yoast SEO 9.3`.

* Adds an extra escape character to make sure `-` in the `stripWordBoundaries` regex gets properly escaped. Previously this wasn't the case, resulting in `+-;` to be interpreted as `all characters between index 43 (+) and 59 (;)`. These indexes happen to include the numerals `0-9` (and also `+,-./_;`.

## Test instructions

* In the plugin, add a text with >150 words.
* Set a keyphrase including a numeral, e.g. `Yoast SEO 9.3`. Include the keyphrase at least once in your text.
* Mark the keyphrase using the eye marker next to the keyphrase density assessment or, if you're using premium, the keyphrase distribution assessment.
* Confirm that the whole keyphrase gets marked and the text isn't changed.

Fixes #2007

### DIFF
--- a/src/stringProcessing/markWordsInSentences.js
+++ b/src/stringProcessing/markWordsInSentences.js
@@ -23,24 +23,11 @@ const collectMarkingsInSentence = function( sentence, topicFoundInSentence ) {
 export function markWordsInSentences( wordsToMark, sentences, locale ) {
 	let topicFoundInSentence = [];
 	let markings = [];
-	const indexOfSentence = 0;
-	let indexRunningThroughSentence = 0;
-	const matchesIndices = [];
 
 	sentences.forEach( function( sentence ) {
 		topicFoundInSentence = matchWords( sentence, wordsToMark, locale ).matches;
 
 		if ( topicFoundInSentence.length > 0 ) {
-			topicFoundInSentence.forEach( function( occurrence ) {
-				const indexOfOccurrenceInSentence = sentence.indexOf( occurrence, indexRunningThroughSentence );
-				matchesIndices.push(
-					{
-						index: indexOfOccurrenceInSentence + indexOfSentence,
-						match: occurrence,
-					}
-				);
-				indexRunningThroughSentence += indexOfOccurrenceInSentence + occurrence.length;
-			} );
 			markings = markings.concat( new Mark( {
 				original: sentence,
 				marked: collectMarkingsInSentence( sentence, topicFoundInSentence ),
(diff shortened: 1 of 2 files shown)


src/stringProcessing/markWordsInSentences.js
```diff
@@ -23,24 +23,11 @@ const collectMarkingsInSentence = function( sentence, topicFoundInSentence ) {
...
-	const indexOfSentence = 0;
-	let indexRunningThroughSentence = 0;
-	const matchesIndices = [];
 
 	sentences.forEach( function( sentence ) {
 		topicFoundInSentence = matchWords( sentence, wordsToMark, locale ).matches;
 
 		if ( topicFoundInSentence.length > 0 ) {
-			topicFoundInSentence.forEach( function( occurrence ) {
-				const indexOfOccurrenceInSentence = sentence.indexOf( occurrence, indexRunningThroughSentence );
```
hansjovis: According to a commit comment, this code is removed because it is not used anymore.
I think you forgot it to add to the PR as a relevant technical choice 😉.

PR #2019, opened 2018-11-27
1763 add worker transporter tests

### LINKED ISSUE #1763: Add worker transporter tests
Please add tests for the worker transporter: `serialize` and `parse`.

hansjovis:
## Summary

* Adds missing tests for the `parse` and `serialize` functions of the transporter layer between the analysis web worker and the plugin.

* I noticed that the test for parsing serialized `Mark`s I added was failing. Since it was an easy fix, I decided to solve the problem in this PR, instead of making another issue. 

## Test instructions

* Review the added test files, check if everything that you expect to be tested is tested.
* Run `yarn test`.
   * Check if every test passes.
   * Check if the coverage of `serialize` and `parse` (at the bottom of the coverage report) are 100%).

Fixes #1763


manuelaugustin:
CR & acceptance done 👍

PR #2037, opened 2018-12-03
Added Text node class for Structured Tree.

### LINKED ISSUE #2018: Build structured `Text` tree node
The `Text` object represents a piece of text that can be read by a reader. [The HTML spec has a definition of a paragraph that is very close what we want to represent with a `Text`][html-paragraph]. [The information about phrasing content is also useful][html-phrasing-content].

This object will save both a piece of text and the HTML tags inside that text, but in separate properties. The reason for this is that the `Text` object is the thing that all the linguistic analysis needs to deal with. The linguistic analysis can simply retrieve the text from the `Text` node and knows all the HTML is dealt with.

So the constructor accepts both `text` and `formatting` properties. An example:

Given we have the following HTML:

```html
This <strong>sentence</strong> <strong class="className" data-extra="pieceOfData">needs to <em>be</em> read</strong> to have value as a sentence.
```

This will correspond to the following `Text` properties:

```js
let properties = {
	"text": "This sentence needs to be read to have value as a sentence.",
	"formatting": [
		{
			tag: "strong",
			attributes: null,
			start: 5, // "This ".length
			end: 13, // "This sentence".length
		},
		{
			tag: "strong",
			attributes: {
				class: "className",
				"data-extra": "pieceOfData"
			},
			start: 14, // "This sentence ".length
			end: 30, // "This sentence needs to be read".length
		},
		{
			tag: "em",
			attributes: null,
			start: 23, // "This sentence needs to ".length
			end: 25, // "This sentence needs to be".length
		}
	]
};
```

This conversion is out of scope for this issue. This issue is simply creating the value object that contains this property.

Parent #2015.

[html-paragraph]: https://www.w3.org/TR/html52/dom.html#paragraphs
[html-phrasing-content]: https://www.w3.org/TR/html52/dom.html#phrasing-content

hansjovis:
## Summary

* N/A

* Adds a text node class for the new Text Parser.
* I decided to add a `PhrasingContent` class for holding phrasing content / formatting data. IMO this is better for self-documentation of the code. It had the added advantage of making the `toHtml` method easier to be implemented.
* Added a `toHtml` method to serialize the `Text`-node back to HTML.

## Test instructions

* run `yarn test`, check if all the tests pass and all the added classes (`Text`, `PhrasingContent`) have a coverage of 100%.

Fixes #2018 .

### DIFF
--- a/spec/tree/values/PhrasingContentSpec.js
+++ b/spec/tree/values/PhrasingContentSpec.js
@@ -1,0 +1,32 @@
+import PhrasingContent from "../../../src/tree/values/PhrasingContent";
+
+describe( "PhrasingContent", () => {
+	it( "can make a new PhrasingContent object", () => {
+		const phrasingElement = new PhrasingContent( "strong", 5, 30, { id: "some-id" } );
+
+		expect( phrasingElement.tag ).toEqual( "strong" );
+		expect( phrasingElement.start ).toEqual( 5 );
+		expect( phrasingElement.end ).toEqual( 30 );
+		expect( phrasingElement.attributes ).toEqual( { id: "some-id" } );
+	} );
+
+	describe( "to HTML-string", () => {
+		it( "can generate an HTML-string", () => {
+			const phrasingElement = new PhrasingContent( "a", 25, 29, {
+				href: "https://example.com",
+			} );
+
+			expect( phrasingElement.toHtml( "some link" ) ).toEqual(
+				"<a href='https://example.com'>some link</a>"
+			);
+		} );
+
+		it( "can generate an HTML-string with no attributes", () => {
+			const phrasingElement = new PhrasingContent( "strong", 25, 29 );
+
+			expect( phrasingElement.toHtml( "some link" ) ).toEqual(
+				"<strong>some link</strong>"
+			);
+		} );
+	} );
+} );
--- a/spec/tree/values/nodes/TextSpec.js
+++ b/spec/tree/values/nodes/TextSpec.js
@@ -1,0 +1,30 @@
+import Text from "../../../../src/tree/values/nodes/Text";
+import PhrasingContent from "../../../../src/tree/values/PhrasingContent";
+
+describe( "Text tree node", () => {
+	it( "can make a new Text tree node", () => {
+		const phrasingElements = [
+			new PhrasingContent( "a", 25, 29 ),
+		];
+		const text = "This is some text with a link.";
+		const textElement = new Text( text, phrasingElements );
+
+		expect( textElement.phrasingElements ).toEqual( textElement.phrasingElements );
+		expect( textElement.text ).toEqual( text );
+	} );
+
+	it( "can generate an HTML-string from the text and accompanying phrasing content.", () => {
+		const phrasingElements = [
+			new PhrasingContent( "a", 25, 29, {
+				href: "https://example.com",
+			} ),
+			new PhrasingContent( "strong", 0, 4 ),
+		];
+		const text = "This is some text with a link.";
+		const textElement = new Text( text, phrasingElements );
+
+		expect( textElement.toHtml() ).toEqual(
+			"<strong>This</strong> is some text with a <a href='https://example.com'>link</a>."
+		);
+	} );
+} );
--- a/spec/tree/values/nodes/TextSpec.js
+++ b/spec/tree/values/nodes/TextSpec.js
@@ -24,7 +24,7 @@ describe( "Text tree node", () => {
 		const textElement = new Text( text, phrasingElements );
 
 		expect( textElement.toHtml() ).toEqual(
-			"<strong>This</strong> is some text with a <a href='https://example.com'>link</a>."
+			"<strong>This</strong> is some text with a <a href=\"https://example.com\">link</a>."
 		);
 	} );
 } );
--- a/src/tree/values/PhrasingContent.js
+++ b/src/tree/values/PhrasingContent.js
@@ -1,0 +1,46 @@
+/**
+ * Represents phrasing content (e.g. <strong>, <a>, <em>) within an HTML-text.
+ */
+class PhrasingContent {
+	/**
+	 * Represents phrasing content (e.g. <strong>, <a>, <em>) within an HTML-text.
+	 *
+	 * @param {string} tag the HTML-tag of this element (e.g. "strong", "a", etc.)
+	 * @param {number} start the start position of the tag within the text.
+	 * @param {number} end the end position of the tag within the text.
+	 * @param {Object} [attributes] the attributes (as key-value pairs, e.g. "href='...'" => { href: '...' } ).
+	 */
+	constructor( tag, start, end, attributes ) {
+		this.tag = tag;
+		this.attributes = attributes;
+		this.start = start;
+		this.end = end;
+	}
+
+	/**
+	 * Stringifies the given HTML-attributes to a string of " key=value" pairs.
+	 *
+	 * @param {Object} attributes the attributes to serialize.
+	 * @returns {string} the stringified attributes.
+	 */
+	getAttributeString( attributes ) {
+		if ( attributes ) {
+			return Object.keys( attributes ).reduce( ( string, key ) => {
+				return string + ` ${key}='${attributes[ key ]}'`;
+			}, "" );
+		}
+		return "";
+	}
+
+	/**
+	 * Stringifies this phrasing content to an HTML-string.
+	 *
+	 * @param {string} content the content to insert between the content tags.
+	 * @returns {string} the HTML-string.
+	 */
+	toHtml( content ) {
+		return `<${this.tag}${this.getAttributeString( this.attributes )}>${content}</${this.tag}>`;
+	}
+}
+
+export default PhrasingContent;
--- a/src/tree/values/PhrasingContent.js
+++ b/src/tree/values/PhrasingContent.js
@@ -26,7 +26,7 @@ class PhrasingContent {
 	getAttributeString( attributes ) {
 		if ( attributes ) {
 			return Object.keys( attributes ).reduce( ( string, key ) => {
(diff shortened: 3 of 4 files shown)


src/tree/values/FormattingElement.js
```diff
@@ -0,0 +1,46 @@
...
+		this.end = end;
+	}
+
+	/**
+	 * Stringifies the given HTML-attributes to a string of " key=value" pairs.
+	 *
+	 * @param {Object} attributes the attributes to serialize.
+	 * @returns {string} the stringified attributes.
+	 */
+	getAttributeString( attributes ) {
```
igorschoester: This always returns a space in front, which works because it is only used in `toHTML`, but is not very expected if one would use this function separately (not sure why). If that should not happen, maybe add an underscore in front of the name to mark it as private?

src/tree/values/FormattingElement.js
```diff
@@ -0,0 +1,46 @@
...
+		return "";
+	}
+
+	/**
+	 * Stringifies this phrasing content to an HTML-string.
+	 *
+	 * @param {string} content the content to insert between the content tags.
+	 * @returns {string} the HTML-string.
+	 */
+	toHtml( content ) {
```
igorschoester: Should we support self-closing variants?

src/tree/values/nodes/TextContainer.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * Represents a text (with optional phrasing content) within an HTML-document that can be read by a reader.
+ */
```
igorschoester: The example of the issue could be nice to include here?

src/tree/values/FormattingElement.js
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	/**
+	 * Stringifies the given HTML-attributes to a string of " key=value" pairs.
+	 *
+	 * @param {Object} attributes the attributes to serialize.
+	 * @returns {string} the stringified attributes.
+	 */
+	getAttributeString( attributes ) {
+		if ( attributes ) {
```
igorschoester: This should maybe check if attributes is an actual object? `Object.keys` on an array or string gives unexpected results.

src/tree/values/FormattingElement.js
```diff
@@ -0,0 +1,46 @@
...
+	/**
+	 * Represents phrasing content (e.g. <strong>, <a>, <em>), except text, within an HTML-text.
+	 *
+	 * @param {string} tag the HTML-tag of this element (e.g. "strong", "a", etc.)
+	 * @param {number} start the start position of the tag within the text.
+	 * @param {number} end the end position of the tag within the text.
+	 * @param {Object} [attributes] the attributes (as key-value pairs, e.g. "href='...'" => { href: '...' } ).
+	 */
+	constructor( tag, start, end, attributes ) {
+		this.tag = tag;
```
igorschoester: Should we type check these variables?

spec/tree/values/PhrasingContentSpec.js
```diff
@@ -0,0 +1,32 @@
...
+describe( "PhrasingContent", () => {
+	it( "can make a new PhrasingContent object", () => {
+		const formattingElement = new FormattingElement( "strong", 5, 30, { id: "some-id" } );
+
+		expect( formattingElement.tag ).toEqual( "strong" );
+		expect( formattingElement.start ).toEqual( 5 );
+		expect( formattingElement.end ).toEqual( 30 );
+		expect( formattingElement.attributes ).toEqual( { id: "some-id" } );
+	} );
+
```
igorschoester: Missing the `getAttributeString` specs. Mostly to test some unhappy paths.

spec/tree/values/PhrasingContentSpec.js
```diff
@@ -0,0 +1,32 @@
...
+	describe( "to HTML-string", () => {
+		it( "can generate an HTML-string", () => {
```
igorschoester: I'm missing some unhappy path tests. And a multiple elements test.

src/tree/values/FormattingElement.js
hansjovis: Yes we should! 🙂 
E.g. ``.

src/tree/values/FormattingElement.js
hansjovis: In my opinion having the JSDoc types is enough of a barrier to prevent this function being called with the wrong parameters.

src/tree/values/FormattingElement.js
hansjovis: You're right! I made this into a private function.

src/tree/values/FormattingElement.js
hansjovis: I've rewritten the function to use `this.attributes` instead, since it is now a private function.
Since the JSDoc in the constructor indicates (and checks in the IDE) if `attributes` is an object, I think we are relatively safe.

spec/tree/values/PhrasingContentSpec.js
hansjovis: Added a test.
Most unhappy paths (ill-formed attributes) will be caught by `htmlparser5` is my guess, so I am a bit hesitant to add more.
We can always adds more tests later on.

spec/tree/values/nodes/TextSpec.js
```diff
@@ -0,0 +1,68 @@
...
+
+			expect( textContainer.toHtml() ).toEqual(
+				"This is some text with a link."
+			);
+		} );
+
+		/**
+		 * Nested content does not work yet...
+		 */
+		it.skip( "can generate an HTML-string from multiple nested formatting elements", () => {
```
nataliashitova: Please make an issue to fix the nested content problem 😄

hansjovis: OK 🙂. Issue: #2047.

src/tree/values/FormattingElement.js
```diff
@@ -0,0 +1,67 @@
...
+		this.start = start;
+		this.end = end;
+		this.selfClosing = selfClosing;
+		this._validate();
+	}
+
+	/**
+	 * Checks if this is a valid FormattingElement. If it is not, throws an error.
+	 *
+	 * @private
```
igorschoester: This private reminds me, there is also a `@throws` you can use for the `RangeError` http://usejsdoc.org/tags-throws.html 

Also, the upper border of the end could be checked for too right?
Or we could just reset `start < 0` to 0 and `> end` to end. And/or swap the start/end with a warning instead.
Possibilities 💥

src/tree/values/FormattingElement.js
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	/**
+	 * Stringifies this phrasing content to an HTML-string.
+	 *
+	 * @param {string} [content=""] The optional content to insert between the content tags.
+	 *
+	 * @returns {string} The HTML-string.
+	 */
+	toHtml( content = "" ) {
```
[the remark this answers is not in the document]
igorschoester: That seems to be more in the scope of an assessment for example, not the data layer.

src/tree/values/FormattingElement.js
```diff
@@ -0,0 +1,67 @@
+/**
+ * Represents formatting elements (e.g. <strong>, <a>, <em>), except text, within an HTML-text.
```
igorschoester: As discussed, this is supposed to represent source independent data. Please look for the other places too.

```suggestion
 * Represents formatting elements (e.g. <strong>, <a>, <em>), except text, within an text.
```

src/tree/values/nodes/TextContainer.js
```diff
@@ -0,0 +1,75 @@
...
+	constructor( text, formatting ) {
+		this.text = text;
+		this.formatting = formatting;
+		this._validate();
+	}
+
+	/**
+	 * Checks if this is a valid TextContainer. If it is not, throws an error.
+	 *
+	 * @private
```
igorschoester: Same as with FormattingElement.

JSDoc has an @throws too 😄 
Plus the position could be auto-adjusted, warning instead, possibilities!

src/tree/values/FormattingElement.js
hansjovis: The code now tries to fix the mistakes and gives a warning in the console when it does.

This does have the possibility to mess the formatting up. For example formatting elements can overlap after "fixing" it. However, trying to solve this would lead to into a rabbit hole that I do not want to get into... Therefor, I would suggest leaving it at this.

src/tree/values/FormattingElement.js
hansjovis: Changed the docs so they are source agnostic instead of focused on HTML.

src/tree/values/nodes/TextContainer.js
hansjovis: Changed the code such that it tries to fix possible mistakes and give a warning in the console.

src/tree/values/FormattingElement.js
```diff
@@ -0,0 +1,45 @@
...
+		this.end = end;
+
+		// Swap end and start positions when end is smaller.
+		if ( this.end < this.start ) {
+			this._swapStartAndEnd();
+			console.warn( `End position smaller than start of '${type}' element. They have been swapped.` );
+		}
+
+		// Set start position to zero when smaller than zero.
+		if ( this.start < 0 ) {
```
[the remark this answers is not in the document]
hansjovis: You are right. I thought that, when `endIndex` < 0, it would always be swapped with `startIndex` by the code before. In this way, I thought, `endIndex` would never be smaller than zero. 
However, both `startIndex` and `endIndex` could be smaller than zero and not be swapped (e.g. when `startIndex = -8` and `endIndex = -6`).

I did change the code to check `startIndex` and `endIndex` separately, since we do not want the `endIndex` to be zero when only the `startIndex` is smaller than zero (I think?).

nataliashitova:
Acceptance 🎉

PR #2039, opened 2018-12-04
Add paragraph structure

### LINKED ISSUE #2017: Build structured `Paragraph` component
In https://github.com/Yoast/YoastSEO.js/issues/2018 I included a link to how the HTML spec defines a paragraph. Our definition is a little different, according to the HTML spec the text inside a list item is also a paragraph. For our analysis it is more useful to keep these separate. So we have both a `Paragraph` node and a `Text` node.

The `Paragraph` object represents a readable paragraph in the text. It will have the following properties:

```js
properties = {
	"text": new Text(), // The text inside the paragraph.
	"startHtml": "<p>", // Optional, used when the paragraph is enclosed by <p> tags.
	"endHtml": "</p>", // Optional, used when the paragraph is enclosed by <p> tags.
};
```

Parent #2015

nataliashitova:
## Summary

* [not user-facing] Adds a tree-structure element for Paragraph.

* Contrary to the original suggestion by Anton, I added opening and closing tags as parameters to the constructor, for flexibility.

## Test instructions

* Run the `yarn test` and see that nothing fails.

Fixes #2017


spec/tree/values/nodes/ParagraphSpec.js
```diff
@@ -0,0 +1,36 @@
...
+		const closingTag = "</p>";
+
+		const paragraphElement = new Paragraph( textElement, openingTag, closingTag );
+
+		expect( paragraphElement.textContainer ).toEqual( textElement );
+		expect( paragraphElement.startHtml ).toEqual( "<p>" );
+		expect( paragraphElement.endHtml ).toEqual( "</p>" );
+	} );
+
+	it( "can make a new Text tree node", () => {
```
igorschoester:
```suggestion
	it( "can make a new Paragraph tree node", () => {
```

src/tree/values/nodes/Paragraph.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * Represents a paragraph with text within an HTML-document that can be read by a reader.
+ */
+class Paragraph {
+	/**
+	 * A paragraph within an HTML-document that can be read by a reader.
+	 *
+	 * @param {TextContainer} textContainer That includes the text without any formatting and phrasing elements if any.
+	 * @param {string} openingTag The opening tag of the paragraph.
```
igorschoester: Please use the JSDoc optional & default indicators, also for the closingTag.

```js
	 * @param {string} [openingTag="<p>"] The opening tag of the paragraph.
```

spec/tree/values/nodes/ParagraphSpec.js
```diff
@@ -0,0 +1,36 @@
...
+
+describe( "Text tree node", () => {
+	it( "can make a new Paragraph tree node with paragraph opening tags", () => {
+		const phrasingElements = [
+			new FormattingElement( "a", 25, 29 ),
+		];
+		const text = "This is some text with a link.";
+		const textElement = new TextContainer( text, phrasingElements );
+
+		const openingTag = "<p>";
```
igorschoester: The use of `<p>` again here does not seem optimal to me, since that is also the default.

src/tree/values/nodes/Paragraph.js
```diff
@@ -0,0 +1,22 @@
...
+	 * @param {string} closingTag The closing tag of the paragraph.
+	 */
+	constructor( textContainer, openingTag = "<p>", closingTag = "</p>" ) {
```
igorschoester: Is `<p>` a wanted default? Or is an empty string more appropriate?

src/tree/values/nodes/Paragraph.js
```diff
@@ -0,0 +1,22 @@
...
+		// The text with markup inside the paragraph.
+		this.textContainer = textContainer;
+		// The opening tag of the paragraph.
+		this.startHtml = openingTag;
```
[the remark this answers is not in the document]
nataliashitova: I like `start` and `end` 😄

src/tree/values/nodes/Paragraph.js
nataliashitova: I think `<p>` makes a lot of sense as a default, I do not think an empty string is more appropriate. We can keep `<p>` for now and change it later if we notice that it is not good.

igorschoester: It is now `tag` with the default of an empty string. What made you change your mind?

src/tree/values/nodes/Paragraph.js
```diff
@@ -0,0 +1,22 @@
...
+class Paragraph extends Node {
+	/**
+	 * A paragraph within a document.
+	 *
+	 * @param {TextContainer} textContainer That includes the text without any formatting and phrasing elements if any.
+	 * @param {number} startIndex 			Start position of this paragraph in the raw text-string.
+	 * @param {number} endIndex			End position of this paragraph in the raw text-string.
+	 * @param {string} [tag=""]			Optional tag to use for opening / closing this paragraph.
+	 */
+	constructor( textContainer, startIndex, endIndex, tag = "" ) {
```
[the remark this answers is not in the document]
nataliashitova: We do not know yet if we will need to rebuild the HTML from the tree or not. Right now it looks like with our decision to save `startIndex` and `endIndex` for every Node, saving tag names is not necessary. As agreed with you, we should keep the `tag` attribute for now, but we will see later if it is needed. 
I agree that our decisions (including naming decisions) until now were massively HTML-oriented, but as we progress with the Tree Builder we will see which of our choices will remain HTML-specific, and which will need to be changed to accommodate a higher level of abstractness.

src/tree/values/nodes/Paragraph.js
```diff
@@ -0,0 +1,22 @@
...
+		super( "paragraph", startIndex, endIndex );
+		// The text with markup inside the paragraph.
+		this.textContainer = textContainer;
```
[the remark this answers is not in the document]
nataliashitova: I suggest that we see later if we need empty paragraphs.

PR #2068, opened 2018-12-12
Fix locale-specific config import in SentenceLengthAssessment

### LINKED ISSUE #2063: Language-specific boundaries for the SentenceLength assessment are not applied in the full text specs
In the SentenceLength assessment, some languages (Polish, Russian, Italian) have different boundaries (either for the recommended sentence length, or for the recommended percentage of sentences over the word limit) than the default boundaries. In the full text specs, these language-specific boundaries are not applied, though. In the assessment-specific specs, though, they are applied correctly, so we need to investigate why that does not work in the full text specs specifically.

nataliashitova:
## Summary

* [not-user-facing] Fixes the bug when the locale-specific configuration was not exported in SentenceLengthAssessment in full-text tests.

*

## Test instructions

* Run `yarn test` and see that no tests fail.

Fixes #2063


igorschoester:
CR & Acceptance 👍

PR #2093, opened 2019-01-09
Remove slightly from the feedback string for cornerstone content

### LINKED ISSUE #2032: Cornerstone TextLength assessment returns suboptimal feedback

The plugin to return a good feedback string for a 400-word cornerstone text.

## What happened instead?

screen shot 2018-11-28 at 10 50 39

Enter a text which contains 401 words and observe the result.

## Technical info

* Platform: WordPress
* Platform version: 4.9.8
* Yoast SEO version: 9.3.1

nataliashitova:
## Summary

* Changes feedback for the assessment that checks the length of the text in cornerstone articles. 

* Changes one feedback string for the assessment that checks the length of the text for cornerstone content. If a cornerstone text has a length between 400 and 900 words, the assessment now returns an orange bullet and a feedback string _"Text length: The text contains __ words. This is below the recommended minimum of __ words. Add more content."_, instead of an orange bullet and a feedback string _"Text length: The text contains __ words. This is slightly below the recommended minimum of __ words. Add a bit more copy."_

## Test instructions

* Run a developer environment or link the branch to the trunk of wordpress-seo
* Check if the assessment works as described [here](https://github.com/Yoast/YoastSEO.js/wiki/Scoring-SEO-analysis#11-text-length) for both regular / cornerstone content and under current analysis and recalibrated analysis.

Fixes #2032

### DIFF
--- a/src/assessments/seo/TextLengthAssessment.js
+++ b/src/assessments/seo/TextLengthAssessment.js
@@ -33,6 +33,8 @@ export default class TextLengthAssessment extends Assessment {
 			},
 			urlTitle: createAnchorOpeningTag( "https://yoa.st/34n" ),
 			urlCallToAction: createAnchorOpeningTag( "https://yoa.st/34o" ),
+
+			cornerstoneContent: false,
 		};
 
 		this.identifier = "textLength";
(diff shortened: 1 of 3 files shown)


src/assessments/seo/TextLengthAssessment.js
```diff
@@ -90,6 +92,20 @@ export default class TextLengthAssessment extends Assessment {
...
 	}
 
+	/**
+	 * Checks if the feedback should include "slightly".
+	 *
+	 * @param {number} score The score the text received for the assessment.
+	 *
+	 * @returns {boolean} Whether the feedback should include "slightly".
+	 */
+	useSlightly( score ) {
```
igorschoester: This function combines two things:
- Checking if the score is slightly below the minimum.
- If this is cornerstone content.

I think it is better to have purer functions to avoid future confusion.
This will also result in a clearer name of the function. I don't think the current name covers it. Also, if the score is not slightly below minimum, it should return false, but it does not return anything specific (thus undefined).

I would suggest `isCornerstoneContent` and `isSlightlyBelowMinimum` as two separate functions.

manuelaugustin:
CR & acceptance done 👍

PR #2101, opened 2019-01-17
Upgrade webpack-dev-server by re-ejecting a create-react-app

atimmer:
## Summary

* _Not applicable_ 

* Because just normally upgrading `webpack` & `webpack-dev-server` turned out to be too complicated we instead ejected a brand new `create-react-app`. Then we copied all those files and reapplied our custom changes.

## Test instructions

* Run the example inside `examples/webpack`, make sure everything you expect to work, works. And there are no errors. Also, `webpack-dev-server` should be at version 3.1.14, which is why we started this in the first place.


hansjovis:
Acceptance test: ✅

Tested the example and everything works as expected!

PR #2102, opened 2019-01-18
2080 build tree researcher

### LINKED ISSUE #2080: Build TreeResearcher
The `TreeResearcher` should be build to fulfill the same role as the `Researcher` is currently fulfilling. With two main differences:

The `TreeResearcher` works with a tree while the the `Researcher` works with a `Paper`.
The `TreeResearcher` has async methods when applicable to facilitate more functionality in the future.
It has the following requirements:

* It should be build using the `class` syntax.
* It should have `100%` test coverage.

It should have the following methods:

* `constructor`, same as `Researcher` except that it doesn't accept a `Paper` object.
* `setTree( treeRootNode )`, should set the tree reference for the `Tree`Researcher``.
* `addResearch( name, research )`, same as `Researcher`.
* `hasResearch( name )`, same as `Researcher`.
* `getAvailableResearches()`, same as `Researcher`.
* `getResearch( name )`, this method should be asynchronous. So that means it returns a `Promise` that resolves with the results of the research.
* `getResearchForNode( name, node )`, this method should asynchronous. So that means it returns a `Promise` that resolves with the results of the research.
* `addResearchData( researchName, data )`, same as `Researcher`.
* `getData( researchName )`, returns the data for a research. This method should be asynchronous. So it returns a `Promise` that resolves with the data that the research requires.

## Caching

All results of research should be cached inside the relevant nodes. I think for the first iteration we should simply cache in the root node and in every leaf node. We can add other nodes later.

This means that a node has the following object called `researchResults`:

```js
{
    transitionWords: [ /* transitionWords */ ],
    passiveVoice: [ /* passive voice sentences */ ],
}
```

These results can then be combined with siblings to calculate the results for the root node. Read along to see how this should work.

## Recursiveness

`getResearchForNode` is a recursive method that call itself to get the research for its children. For a leaf node it should just return the results. The method itself can not decide what a leaf node is because that depends on the research. For example:

* For `transitionWords` a node with a `TextContainer` is a leaf node.
* For `fleschReadingEase` the root node is a leaf node. This is because that score is only defined for the whole text. (It is a more accurate score when taking more of the text into consideration). It's also probably hard to merge the results for child nodes for the flesch reading ease score.

To facilitate the `getResearchForNode` method all Research should be class instances with the following interface (using duck typing):

* A `calculateFor( node )` method that calculates a researchResult for the given node.
* A `isLeaf( node )` method that returns whether the given node is considered a leaf node for the research.
* A `mergeChildrenResult( resultsForChildren )` method that merges an array of results for all the children of this node.

The recursion step for `getResearchForNode` can use these methods to combine the results for a given research.

Combined with the previous point about caching this would look like this:

```js
async function getResearchForNode( name, node ) {
	const research = this.getResearchInstance( name );
	let researchResult = Promise.resolve();

	if ( research.isLeafNode( node ) ) {
		if ( ! node.hasResearchResult( name ) ) {
			node.setResearchResult( name, await research.calculateFor( node ) );
		}

		researchResult = node.getResearchResult( name );
	} else {
		const children = node.children;

		const resultsForChildren = await Promise.all( children.map( ( child ) => {
			return this.getResearchForNode( name, child );
		} );

		researchResult = research.mergeChildrenResults( resultsForChildren );
	}

	return researchResult;
}
```

Parent #1903
Requires #2028

hansjovis:
## Summary

* _[not-user-facing]_ Adds a TreeResearcher (Treesearcher?) that can apply researches to trees as parsed by the new tree builder functionality.

* Also adds a simple test research (`spec/specHelpers/tree/TestResearch.js`) that's used in the tests. It just splits each text on whitespace and counts the resulting tokens.

## Test instructions

* Checkout this branch locally.
* Run `yarn test`, check if the coverage of all the files in the `tree` subfolder is 100%.
* Run `grunt check`, check if the code added by this PR (`TreeResearcher`, `Research`, `TreeResearchSpec`, `TestResearch`, `ResearchSpec` and to `Node`) does not lead to any eslint warnings.
* Take a look at the code and tests. Check if any (edge) cases are not tested yet and/or you see any issues in the code that may lead to problems in the future.

Fixes #2080


spec/tree/research/researches/Research.js
```diff
@@ -0,0 +1,41 @@
+import Research from "../../../../src/tree/research/researches/Research";
```
igorschoester: Nothing for this line of code.
However, the filename of this file should be `ResearchSpec.js`, right?

hansjovis: You are right! 🙂

nataliashitova:
Acceptance 👍

PR #2118, opened 2019-01-30
Add FunctionWordsInKeyphraseAssessment as seo assessment export

Kingdutch:
The `FunctionWordsInKeyphraseAssessment` class was not available as an seo assessment export. This changes makes it available with the others :)

## Summary

* `FunctionWordsInKeyphraseAssessment` is now available as an assessment in the `seo` group. It was previously forgotten.

* It appears to be the only SeoAssessor assessment that was not exported.

## Test instructions

The following should work:
```js
import { assessments } from 'yoastseo';

const FunctionWordsInKeyphraseAssessment = new assessments.seo.FunctionWordsInKeyphraseAssessment();

console.log(FunctionWordsInKeyphraseAssessment);
```

Fixes #

### DIFF
--- a/src/assessments/index.js
+++ b/src/assessments/index.js
@@ -9,6 +9,7 @@ import TextPresenceAssessment from "./readability/textPresenceAssessment";
 import TransitionWordsAssessment from "./readability/transitionWordsAssessment";
 import WordComplexityAssessment from "./readability/wordComplexityAssessment";
 
+import FunctionWordsInKeyphraseAssessment from './FunctionWordsInKeyphraseAssessment';
 import InternalLinksAssessment from "./seo/InternalLinksAssessment";
 import IntroductionKeywordAssessment from "./seo/IntroductionKeywordAssessment";
 import KeyphraseLengthAssessment from "./seo/KeyphraseLengthAssessment";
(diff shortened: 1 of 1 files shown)


igorschoester:
Thank you for your contribution again @Kingdutch. This will get merged via https://github.com/Yoast/YoastSEO.js/pull/2137.

Please note that in the long run, this way of accessing assessments will get deprecated in favor of using the analysis worker.

Kingdutch:
Hi @igorschoester,

Is the Yoast team already committed to that decision? From what I can see from the documentation, the Webworker is less flexible from an integrators perspective (looking at the documentation in the repository). 

The separately exported components make it possible to modify them in extending projects and offers quite a bit of configurability. One thing I've had to do for an integration with Drupal is to change the way strings are translated. When looking at the web worker it seems that the trend of passing the translator top down is enforced even more than in the current code base.

I'd be interested in having a discussion on how we could keep supporting other use-cases 🙂 

If you want to see how I've currently used the assessments to support your assessment logic while changing the translator then you can check out https://github.com/goalgorilla/RTSEO.js/tree/feature/react-presenter

~ Alexander

igorschoester:
We are committed to the WebWorker functionality @Kingdutch.
You make a good point with the configuration of the feedback strings. We should include those in the configuration so that you can still override them. I will create an issue for this.

PR #2133, opened 2019-02-06
Add functionality to recognize German adjective forms

### LINKED ISSUE #2127: Add word form recognition for German adjectives
We should be able to recognize adjectives in their different morphological forms in German.

This will require:
- [x] Adding the premium configuration data for the exceptions and regular suffixes.
- [x] Refactor `getForms` by adding a `getNounForms` and a `getAdjectiveForms` file for the category-specific functionality.
- [x] Adding functionality to recognize two possible stem versions for adjectives ending in -er/-el and adjectives with umlaut stem changes.
- [x] Checking exceptions. This can be done after we check for noun exceptions.
- [ ] Checking whether the list of regular suffixes needs to be changed based on the ending of the stem.
- [x] Add functionality to return the right suffixes for superlatives/comparatives depending on the ending of the word.
- [ ] Adding all the regular adjective suffixes in addition to the noun suffixes for words that are no exceptions. 

Parent #2095

manuelaugustin:
## Summary

* [not-user-facing] Adds functionality to create German adjective forms.

* For nouns, we check exceptions by matching the exception stems at the end of the stemmed word. This is because nouns are commonly are used in compounds (e.g. `stadt` in `Hauptstadt`). Since this is far less common for adjectives, I decided to match exceptions with the full stemmed word.
* I added some function words we came across while researching adjectives.

## Test instructions

* Make sure you have the premium configuration set to the right branch (see https://github.com/Yoast/YoastSEO.js-premium-configuration/pull/7).
* Change the language of your site to German.
* Add a German text and a keyphrase including German adjectives.
* Add different grammatical forms of the adjective in your text; make sure these forms are recognized.

Fixes #2127

### DIFF
--- a/src/morphology/german/addAdjectiveSuffixes.js
+++ b/src/morphology/german/addAdjectiveSuffixes.js
@@ -1,0 +1,66 @@
+import { uniq as unique } from "lodash-es";
+
+
+const removeSuffixesIfWordEndsInConsonant = function( stemmedWord, suffixList, suffixesToRemove ) {
+	const consonantAtEndOfWord = /[b-df-hj-kmnpqstv-xzß]$/;
+
+	for ( let i = 0; i < suffixesToRemove.length; i++ ) {
+		if ( consonantAtEndOfWord.test( stemmedWord ) ) {
+			suffixList = suffixList.filter( suffix => suffix !== "n" );
+		}
+	}
+
+	return suffixList;
+};
+
+export function getSuffixesComparative( morphologyDataAdjectives, stemmedWord ) {
+	const takesREnding = morphologyDataAdjectives.takesComparativeREnding.slice();
+
+	if ( takesREnding.some( ending => stemmedWord.endsWith( ending ) ) ) {
+		return morphologyDataAdjectives.comparativeSuffixesR;
+	}
+
+	return morphologyDataAdjectives.comparativeSuffixesEr;
+}
+
+export function getSuffixesSuperlative( morphologyDataAdjectives, stemmedWord ) {
+	const takesEstEnding = morphologyDataAdjectives.takesSuperlativeEstEnding.slice();
+
+	if ( takesEstEnding.some( ending => stemmedWord.endsWith( ending ) ) ) {
+		return morphologyDataAdjectives.superlativeSuffixesEst;
+	}
+
+	return morphologyDataAdjectives.superlativeSuffixesSt;
+}
+
+export function addRegularSuffixes( morphologyDataAdjectives, stemmedWord, suffixesToRemove = [] ) {
+	const regularSuffixes = morphologyDataAdjectives.regularSuffixes.slice();
+	let suffixesToAdd = regularSuffixes.filter( suffix => suffixesToRemove.indexOf( suffix ) === -1 );
+
+	// Delete -n from regular suffixes if the word ends in a consonant (except l and r).
+	suffixesToAdd = removeSuffixesIfWordEndsInConsonant( stemmedWord, suffixesToAdd, [ "n" ] );
+
+	return unique( suffixesToAdd.map( suffix => stemmedWord.concat( suffix ) ) );
+}
+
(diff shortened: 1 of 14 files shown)


src/morphology/german/checkNounExceptions.js
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * Checks whether a stemmed word is on any of the exception lists.
+ *
+ * @param {Object}  morphologyDataNouns The German morphology data for nouns.
+ * @param {string}  stemmedWordToCheck  The stem to check.
+ *
+ * @returns {string[]} The created word forms.
+ */
+export function checkNounExceptions( morphologyDataNouns, stemmedWordToCheck ) {
```
hansjovis: I would change the name and description of this function.
`Checking` indicates to me that it only checks whether the stemmed word is an exception. It does not indicate to me that it generates the exceptional forms as well.
Perhaps `generateNounExceptionForms`?

src/morphology/german/getForms.js
```diff
@@ -205,29 +105,41 @@ const addFormsWithRemovedLetters = function( morphologyDataNouns, stemmedWordToC
 export function getForms( word, morphologyData ) {
 	const stemmedWord = stem( word );
 	const forms = new Array( word );
-	const exceptions = checkExceptions( morphologyData.nouns, stemmedWord );
 
-	// Check exceptions.
+	/*
+	 * Check whether the word is on an exception list. Since a given stem might sometimes be on an exception list in
+	 * different word categories (e.g., "sau-" from the noun "Sau" or the adjective "sauer") we need to do this cumulatively.
+	 */
```
hansjovis: I would change this comment to reflect that the function calls below also generate the forms.

src/morphology/german/checkAdjectiveExceptions.js
```diff
@@ -0,0 +1,234 @@
...
+
+/**
+ * Checks whether a give stem stem falls into any of the adjective exception categories.
+ *
+ * @param {Object}  morphologyDataAdjectives The German morphology data for nouns.
+ * @param {string}  stemmedWordToCheck       The stem to check.
+ *
+ * @returns {string[]} The created adjective forms.
+ */
+export function checkAdjectiveExceptions( morphologyDataAdjectives, stemmedWordToCheck ) {
```
hansjovis: Same for this function:
Change this to, for example, `generateAdjectiveExceptionForms`. To indicate that it not only checks, but also generates the exception forms.

src/morphology/german/checkAdjectiveExceptions.js
```diff
@@ -0,0 +1,234 @@
...
+		 * of adjectives with regards to what endings they get on which stem.
+		 */
+		erStemChangeClass1,
+		erStemChangeClass2,
+		erStemChangeClass3,
+		secondStemCompSup,
+		bothStemsComSup,
+	];
+
+	for ( let i = 0; i < exceptionChecks.length; i++ ) {
```
hansjovis: Nice and concise way of applying the functions in turn! 👍

manuelaugustin: 🙇‍♂️

manuelaugustin:
Good point about the function names/documentation! I've changed the naming as suggested.

src/morphology/german/addAdjectiveSuffixes.js
```diff
@@ -3,7 +3,7 @@ import { uniq as unique } from "lodash-es";
 /**
  * Returns a set of comparative suffixes depending on the ending of the stem.
  *
- * @param {Object}  morphologyDataAdjectives    The German morphology data for nouns.
+ * @param {Object}  morphologyDataAdjectives    The German morphology data for adjectives.
```
hansjovis: Did not even notice this error, nice catch!

igorschoester:
Standalone works 👍 
In combination with the plugin, the `worker.initialize` executing in the console didn't work for some reason. Since there will be a better way to test this in the future and this did work in the standalone I don't think this is a blocker at this point.

Acceptance 👍

PR #2135, opened 2019-02-06
Add tree visualization to the devtool

### LINKED ISSUE #2122: Add visualization of the tree to the example
To be able to properly develop the tree and debug it if necessary, the tree needs to be visualized in the example.

The easiest way to implement this is to do these steps:

1. Send the tree to the main thread. (Needs to be build)
2. Transform the tree into JSON. (I think @hansjovis had already build a utility for this)
3. Put this JSON into `react-json-view`: https://www.npmjs.com/package/react-json-view. **Note:** If a node has a `parent` property this will cause a circular reference and this WILL crash your browser when using `react-json-view`.

The other way is to, which might result in a better visualization but is a ton more work:

1. Send the tree to the main thread. (Needs to be build)
2. Build React components for each node in the tree.

If you find a better visualization library that that can obviously also be used.

Parent #1903

igorschoester:
## Summary

* [non-user-facing] Add visualization of the tree.

* Creating an option to build the tree every text change.
* Represent it using react component: `react-json-view`.

## Screenshots
jsontree

## Test instructions

* Go to the path `cd examples/webpack`.
* Install dependencies `yarn`.
* Start the dev server `yarn start`.
* Test the option `Build the tree`. While on it should build the tree every text change. While off it should do nothing. Turning it back on should (most of the time, there can be a timing thing with the worker results) re-build the tree.
* Check the tree representation itself.

Fixes #2122


nataliashitova:
Acceptance 👍 
I did not notice any delay in tree updates.

Issue #2139, opened 2019-02-12
Morphologize internal linking suggestions


nataliashitova:
The current algorithm of relevant words detection in YoastSEO.js does not collapse over different morphological forms of words, which is inefficient, as we have to save forms like `brand`, `brands`, `branding`, `branded` separately in the database. 
We should add functionality that would collapse over different morphological forms before they get saved in the database.

Parent https://github.com/Yoast/wordpress-seo-premium/issues/2235
Requires #2149

PR #2140, opened 2019-02-12
Add paper attributes to internal linking research

### LINKED ISSUE #2114: Internal linking: make use of paper attributes for relevant words extraction
Currently, we extract relevant words from the text of the paper, while ignoring data from all other attributes of the paper: keyphrase, synonyms, title, and meta description. [A study showed](https://docs.google.com/document/d/1S4NwC5C4eAUrjTV0DVrM-HY2AuRO1O9QcETLwU0pF-c/) that simply adding words from paper attributes to the paper's relevant words can dramatically increase the quality of the internal linking tool. 

We need to add words from the paper attributes to the relevant words of the paper (function words being filtered).

Parent https://github.com/Yoast/wordpress-seo-premium/issues/2235

nataliashitova:
## Summary

* Uses words from keyphrase, metadescription, title, and subheadings for internal linking suggestions

* The current mechanism might generate duplicates of terms used for internal linking if words or phrases used in paper attributes (keyphrase, metadescription, title, and subheadings) also pass the density requirements in the text of the paper itself. This problem will be solved by the same functionality that will take care of different morphological forms (see #2139)

## Test instructions

* Use the `example` to check if words from paper attributes are added to the list of relevant words.
* Try adding more tests to check if words from paper attributes end up in the relevant words combinations.

Fixes #2114

### DIFF
--- a/src/stringProcessing/relevantWords.js
+++ b/src/stringProcessing/relevantWords.js
@@ -2,9 +2,6 @@ import getWords from "../stringProcessing/getWords.js";
 import getSentences from "../stringProcessing/getSentences.js";
 import WordCombination from "../values/WordCombination.js";
 import { normalize as normalizeQuotes } from "../stringProcessing/quotes.js";
-import functionWordListsFactory from "../helpers/getFunctionWords.js";
-const functionWordLists = functionWordListsFactory();
-import getLanguage from "../helpers/getLanguage.js";
 
 import { filter } from "lodash-es";
 import { map } from "lodash-es";
@@ -16,6 +13,7 @@ import { take } from "lodash-es";
 import { includes } from "lodash-es";
 import { intersection } from "lodash-es";
 import { isEmpty } from "lodash-es";
+import { uniq } from "lodash-es";
 
 const densityLowerLimit = 0;
 const densityUpperLimit = 0.03;
@@ -253,17 +251,11 @@ function filterCombinations( combinations, functionWords, language ) {
  * Returns the relevant words in a given text.
  *
  * @param {string} text The text to retrieve the relevant words of.
- * @param {string} locale The paper's locale.
+ * @param {string} language The language of the text.
+ * @param {Object} functionWords The function words to filter out.
  * @returns {WordCombination[]} All relevant words sorted and filtered for this text.
  */
-function getRelevantWords( text, locale ) {
-	let language = getLanguage( locale );
-	if ( ! functionWordLists.hasOwnProperty( language ) ) {
-		language = "en";
-	}
-
-	const functionWords = functionWordLists[ language ];
-
+function getRelevantWords( text, language, functionWords ) {
 	const words = getWordCombinations( text, 1, functionWords.all );
 	const wordCount = words.length;
 
@@ -308,9 +300,36 @@ function getRelevantWords( text, locale ) {
 	return take( combinations, relevantWordLimit );
 }
 
+/**
+ * Gets relevant words from keyphrase and synonyms, metadescription, and subheadings.
+ *
+ * @param {string} keyphrase The keyphrase of the paper.
+ * @param {string} synonyms The synonyms of the paper.
+ * @param {string} metadescription The metadescription of the paper.
+ * @param {string[]} subheadings The subheadings of the paper.
+ * @param {string[]} functionWords The function words to use.
+ *
+ * @returns {WordCombination[]} Relevant word combinations from the paper attributes.
+ */
+function getRelevantWordsFromTopic( keyphrase, synonyms, metadescription, subheadings, functionWords ) {
+	const subheadingsJoined = subheadings.join( " " );
+	const attributes = keyphrase.concat( " ", synonyms, " ", metadescription, " ", subheadingsJoined );
+	const attributesWords = uniq( getWords( attributes.toLocaleLowerCase() ) );
+
+	const attributesWordsFiltered = filter( attributesWords, function( word ) {
+		return ( ! includes( functionWords, word.trim() ) );
+	} );
+
+	return map( attributesWordsFiltered, function( word ) {
+		return new WordCombination( [ word ], 5, functionWords );
+	} );
+}
+
+
 export {
 	getWordCombinations,
 	getRelevantWords,
+	getRelevantWordsFromTopic,
 	calculateOccurrences,
 	getRelevantCombinations,
 	sortCombinations,
@@ -326,6 +345,7 @@ export {
 export default {
 	getWordCombinations: getWordCombinations,
 	getRelevantWords: getRelevantWords,
+	getRelevantWordsFromTopic: getRelevantWordsFromTopic,
 	calculateOccurrences: calculateOccurrences,
 	getRelevantCombinations: getRelevantCombinations,
 	sortCombinations: sortCombinations,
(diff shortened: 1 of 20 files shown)


src/stringProcessing/relevantWords.js
```diff
@@ -308,9 +309,43 @@ function getRelevantWords( text, locale ) {
...
+ *
+ * @param {string} keyphrase The keyphrase of the paper.
+ * @param {string} synonyms The synonyms of the paper.
+ * @param {string} metadescription The metadescription of the paper.
+ * @param {string[]} subheadings The subheadings of the paper.
+ * @param {string} locale The locale of the paper.
+ *
+ * @returns {WordCombination[]} Relevant word combinations from the paper attributes.
+ */
+function getRelevantWordsFromTopic( keyphrase, synonyms, metadescription, subheadings, locale ) {
```
hansjovis: I think `getRelevantWordsFromTopic` is not a good description of this method.
Perhaps `getRelevantWordsFromPaperAttributes`?
(Although subheadings are not necessarily attributes / metadata).

hansjovis: Since we could add new metadata / paper attributes in the future, it could be worthwhile to put the attribute arguments in this function in a single object argument:
```js
/**
 * Gets relevant words from the paper's attributes.
 * 
 * @param {Object} attributes                  The attributes to process
 * @param {string} attributes.keyphrase        The keyphrase of the paper.
 * @param {string} attributes.synonyms         The synonyms of the paper.
 * @param {string} attributes.metadescription  The metadescription of the paper.
 * @param {string[]} attributes.subheadings    The subheadings of the paper.
 * @param {string} locale                      The locale of the paper.
 *
 * @returns {WordCombination[]} Relevant word combinations from the paper attributes.
 */
function getRelevantWordsFromPaperAttributes( attributes, locale ) {
      const { keyphrase, synonyms, metadescription, subheadings } = attributes;
```
(And the parameter descriptions can be aligned as well 😛(going full Igor here) ).

src/stringProcessing/relevantWords.js
```diff
@@ -308,9 +309,43 @@ function getRelevantWords( text, locale ) {
...
+		language = "en";
+	}
+
+	const functionWords = functionWordLists[ language ].all;
+
+	const subheadingsJoined = subheadings.join( " " );
+	const attributes = keyphrase.concat( " ", synonyms, " ", metadescription, " ", subheadingsJoined );
+	const attributesWords = uniq( getWords( attributes.toLocaleLowerCase() ) );
+
+	const attributesWordsFiltered = filter( attributesWords, function( word ) {
```
hansjovis: This is a bit more concise and does not use lodash functions:
```js
const attributesWordsFiltered = attributesWords.filter( word => ! functionWords.includes( word.trim() ) );
```

src/stringProcessing/relevantWords.js
```diff
@@ -308,9 +309,43 @@ function getRelevantWords( text, locale ) {
...
+		return ( ! includes( functionWords, word.trim() ) );
+	} );
+
+	return map( attributesWordsFiltered, function( word ) {
```
hansjovis: Same here:
```js
return attributesWordsFiltered.map( word => WordCombination( [ word ], 5, functionWords );
```

src/stringProcessing/relevantWords.js
hansjovis: I always use this source to find the array methods I can use, instead of lodash functions:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array
(Although lodash has some nice functions as well that plain old javascript arrays do not have).

src/stringProcessing/relevantWords.js
igorschoester: Never go full Igor 😜

igorschoester:
Acceptance 🚧 

I have some trouble knowing what specifically needs to get tested here.

1. When I have a text of 20 lines with: `hello world`, I get 3 relevant word entries: `hello`, `world` and `hello world`. Just like before.
However, when I set the keyphrase (this is the same with any other paper attributes -- *except title, see below*) to `hello world` I would've expected to count towards all of them. But it adds two more entries for only `hello` and `world`. You told me this will be added in the next PR. Still, I would expect it to count towards the whole keyphrase `hello world` too.
Is this on purpose?

1. A mismatch between changelog and testing: The SEO title is not used to determine relevant words.

1. Adding `<h2>hello world</h2>` to the content adds a count to all the non-special entries. And includes the 2 special entries for 5 occurences. Therefor, it is counted twice. Should this be?

1. With the above situation the `density` seems influenced by the `<h2></h2>` part, since it is lower than just adding `hello world` as keyphrase for example. `0.119` instead of the expected `0.125`. Thinking of this, the actual expected is `0.5` for the keyphrase and others, right?

1. Adding `hello world` to all the attributes does not alter the amount of occurrences. Is this on purpose?

nataliashitova:
@igorschoester Thanks for your great questions!

> 1. When I have a text of 20 lines with: `hello world`, I get 3 relevant word entries: `hello`, `world` and `hello world`. Just like before.
   However, when I set the keyphrase (this is the same with any other paper attributes -- _except title, see below_) to `hello world` I would've expected to count towards all of them. But it adds two more entries for only `hello` and `world`. You told me this will be added in the next PR. Still, I would expect it to count towards the whole keyphrase `hello world` too.
   Is this on purpose?

Yes. It did not add any more-than-one-word-long word combinations from the paper attributes, because they will soon be deprecated anyway.

> 2. A mismatch between changelog and testing: The SEO title is not used to determine relevant words.

My bad 🤦‍♀️ 

> 3. Adding `<h2>hello world</h2>` to the content adds a count to all the non-special entries. And includes the 2 special entries for 5 occurences. Therefor, it is counted twice. Should this be?

Yes, for now. I will take care of this inconsistency in one of the later PRs that will merge same-word occurrences in the list of relevant words.

> 4. With the above situation the `density` seems influenced by the `<h2></h2>` part, since it is lower than just adding `hello world` as keyphrase for example. `0.119` instead of the expected `0.125`. Thinking of this, the actual expected is `0.5` for the keyphrase and others, right?

Just as with the one above, I will solve this in the next PR. In general, density is of no importance per se, but it serves a proxy of importance. Therefore, as long as we make sure the `right words` are being saved in the database, the theoretical impeccability of our ways of calculating the density of relevant words is not important.


> 5. Adding `hello world` to all the attributes does not alter the amount of occurrences. Is this on purpose?

Right now I am saving all words from attributes with the number of occurrences = 5, which was chosen pretty much at random. The next PR with fine-tun the mechanism counting occurrences in text and attributes, as well as selecting good words to be saved in the database.

nataliashitova:
@hansjovis 
I agree that we should not allow duplicates in the list of relevant words. It makes a lot of sense to take care of such potential duplicates together with collapsing relevant words based on the stemming mechanism (which was implemented in #2150). What I want to do is:

1. Take the list made of relevant words coming from the text as well as from the attributes (i.e., with potential duplicates)
2. Assign a stem to every word
3. Collapse the relevant words if they have the same stem while keeping number_of_occurrences (both in text and in attributes) valid. 

Obviously, duplicates will have the same stem and would be collapsed by this stemming merging mechanism _for free_. 

To reconcile both parties I created a feature branch for internal linking: This will allow to keep development in small chunks and not decrease the quality of the current suggestions. What do you think?

igorschoester:
> 2. A mismatch between changelog and testing: The SEO title is not used to determine relevant words.

Acceptance on this 👍 

Unfortunately, the devtool now compiles with the following warning:
```
./src/components/RelevantWords.js
  Line 75:  Function 'getRelevantWords' has a complexity of 7  complexity

Search for the keywords to learn more about each warning.
To ignore, add // eslint-disable-next-line to the line before.
```

After a small discussion, it seems best to disable this rule for the example.

igorschoester:
Merging on the premise that the above issues will be taken into consideration in the feature branch.

PR #2142, opened 2019-02-12
Implement TreeAssessor

### LINKED ISSUE #2123: Implement TreeAssessor
A `TreeAssessor` should be able to analyze a `Paper` in combination with the `Tree` that has been generated from the text inside the `Paper`.

Unlike the current `Assessor` we will not have the `extends` mechanism for all the different types of assessors. I've identifier two differences between the different assessors:

* They have different assessments: We can solve this by providing the assessment in the constructor to the `TreeAssessor`.
* They have a different algorithm to calculate the overall score: We can solve this by using the [strategy pattern](https://en.wikipedia.org/wiki/Strategy_pattern)

So the `TreeAssessor` will have the following methods (`async` means the function returns a `Promise`):

* `constructor( options = { i18n, scoreAggregator, researcher, assessments = [] } )`, i18n, overallScoreStrategy and researcher are required.
* `getAvailableAssessments`, returns all the assessments that are available.
* async `isApplicable( assessment )`, the same as the current `isApplicable`, except that it should be async.
* async `assess( paper, node )`, same as the current `assess`, except that is returns the results instead of saving them to the object. It should also return the overallScore. It also doesn't have to check if there is a researcher, because that is a required constructor argument in the `TreeAssessor`.
* async `doAssessment( assessment, paper, node )`, same as current `executeAssessment`. Should execute single assessment, should be used in `assess` like it is used now. We don't need any logic for marks, however, because we are going to solve that in a different way.
* async `calculateOverallScore( results )`, this method calculates the overall result based on the given `results`. It uses the given `scoreAggregator` to do this.
* `addAssessment( name, assessment )`, same as the current `addAssessment`.
* `removeAssessment( name )`, same as the current `removeAssessment`.
* `getAssessment`, same as the current `getAssessment`.
* `getApplicableAssessments` can be written (copied), to make writing `assess` easier/DRY.

I will create issues for creating an `SEOScoreAggregator` and a `ReadabilityScoreAggregator`. So this isn't in scope for this issue, but you are going to need an aggregator to write unit tests (create a mock aggregator). So the API of the aggregator looks like this:

* async `aggregate( results )`, returns the overallScore based on the given results.`

Parent #1903

hansjovis:
## Summary

* _[not-user-facing]_ Adds the `TreeAssessor`, which is going to be used for assessments in the (near) future.

* Adds a test assessment and test aggregator (in `specs/specHelpers`) to help with testing the `TreeAssessor` functionality.
* Fixes the JSDoc module structure for `tree/research` to include the `TreeResearcher` and `Research` classes.

## Test instructions

* Run the tests using `yarn test`. Check that the coverage of all files in the `src/tree/assess` folder is 100% (instead of `index.js`).
* Review the code. Check if you do not see any (edge) cases not covered by the implementation of the `TreeAssessor` and related functionality.

Fixes #2123

### DIFF
--- a/spec/specHelpers/tree/TestAssessment.js
+++ b/spec/specHelpers/tree/TestAssessment.js
@@ -36,14 +36,13 @@ class TestAssessment extends Assessment {
 	 *
 	 * @param {Paper} paper                                    The paper to check.
 	 * @param {module:tree/structure.Node} node                The root node of the tree to check.
-	 * @param {module:tree/research.TreeResearcher} researcher The researcher to use to retrieve researches from.
 	 *
 	 * @returns {Promise<*>} The result of this assessment (wrapped in a promise).
 	 *
 	 * @abstract
 	 */
-	async apply( paper, node, researcher ) {
-		const nrOfTokens = await researcher.doResearch( "word count", node );
+	async apply( paper, node ) {
+		const nrOfTokens = await this.researcher.doResearch( "word count", node );
 		if ( nrOfTokens < this.boundary ) {
 			return Promise.resolve(
 				this.generateResult( "Not enough words, try again.", 3 )
--- a/spec/specHelpers/tree/TestAssessment.js
+++ b/spec/specHelpers/tree/TestAssessment.js
@@ -1,0 +1,75 @@
+import { Assessment } from "../../../src/tree/assess";
+import AssessmentResult from "../../../src/values/AssessmentResult";
+
+/**
+ * A test assessment.
+ */
+class TestAssessment extends Assessment {
+	/**
+	 * Makes a new test assessment.
+	 *
+	 * @param {boolean} applicable Whether to return true if checked whether this assessment is applicable.
+	 * @param {number} boundary    The boundary to use when deciding to return a bad or good assessment result.
+	 * @param {string} [name]      An optional name.
+	 */
+	constructor( applicable, boundary, name ) {
+		super( name );
+		this.applicable = applicable;
+		this.boundary = boundary;
+	}
+
+	/**
+	 * Checks whether this assessment is applicable to the given paper and tree combination.
+	 *
+	 * @param {Paper} paper                     The paper to check.
+	 * @param {module:tree/structure.Node} node The tree to check.
+	 *
+	 * @returns {Promise<boolean>} Whether this assessment is applicable to the given paper and tree combination (wrapped in a promise).
+	 */
+	async isApplicable( paper, node ) {  // eslint-disable-line no-unused-vars
+		// Set a minor time out, to make sure that the promises work.
+		return Promise.resolve( this.applicable );
+	}
+
+	/**
+	 * Applies this assessment to the given combination of paper and tree.
+	 *
+	 * @param {Paper} paper                                    The paper to check.
+	 * @param {module:tree/structure.Node} node                The root node of the tree to check.
+	 * @param {module:tree/research.TreeResearcher} researcher The researcher to use to retrieve researches from.
+	 *
+	 * @returns {Promise<*>} The result of this assessment (wrapped in a promise).
+	 *
+	 * @abstract
+	 */
+	async apply( paper, node, researcher ) {
+		const nrOfTokens = await researcher.doResearch( "word count", node );
+		if ( nrOfTokens < this.boundary ) {
+			return Promise.resolve(
+				this.generateResult( "Not enough words, try again.", 3 )
+			);
+		}
+		return Promise.resolve(
+			this.generateResult( "Enough words, good job!", 9 )
+		);
+	}
+
+	/**
+	 * Generates an assessment result.
+	 *
+	 * @param {string} text  The text to add to the assessment result.
+	 * @param {number} score The score to add to the assessment result.
+	 *
+	 * @returns {AssessmentResult} The generated assessment result.
+	 */
+	generateResult( text, score ) {
+		const assessmentResult = new AssessmentResult();
+
+		assessmentResult.setScore( score );
+		assessmentResult.setText( text );
+
+		return assessmentResult;
+	}
+}
+
+export default TestAssessment;
--- a/spec/specHelpers/tree/TestAssessment.js
+++ b/spec/specHelpers/tree/TestAssessment.js
@@ -43,7 +43,7 @@ class TestAssessment extends Assessment {
 	 * @abstract
 	 */
 	async apply( paper, node ) {
-		const nrOfTokens = await this.researcher.doResearch( "test research", node );
+		const nrOfTokens = await this._researcher.doResearch( "test research", node );
 		if ( nrOfTokens < this.boundary ) {
 			return Promise.resolve(
 				this.generateResult( "Not enough words, try again.", 3 )
--- a/spec/specHelpers/tree/TestAssessment.js
+++ b/spec/specHelpers/tree/TestAssessment.js
@@ -8,15 +8,17 @@ class TestAssessment extends Assessment {
 	/**
 	 * Makes a new test assessment.
 	 *
-	 * @param {boolean}                             applicable Whether to return true if checked whether this assessment is applicable.
-	 * @param {number}                              boundary   The boundary to use when deciding to return a bad or good assessment result.
-	 * @param {string}                              name       The name of this assessment.
-	 * @param {module:tree/research.TreeResearcher} researcher The researcher to use.
+	 * @param {boolean}                             applicable         Whether to return true if checked whether this assessment is applicable.
+	 * @param {number}                              boundary           The boundary to use when deciding to return a bad or good assessment result.
+	 * @param {string}                              name               The name of this assessment.
+	 * @param {module:tree/research.TreeResearcher} researcher         The researcher to use.
+	 * @param {boolean}                             [throwError=false] If this assessment should throw an error when it is run.
 	 */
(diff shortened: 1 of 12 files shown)


spec/specHelpers/tree/TestAssessment.js
```diff
@@ -0,0 +1,75 @@
...
+	 * @param {Paper} paper                                    The paper to check.
+	 * @param {module:tree/structure.Node} node                The root node of the tree to check.
+	 * @param {module:tree/research.TreeResearcher} researcher The researcher to use to retrieve researches from.
+	 *
+	 * @returns {Promise<*>} The result of this assessment (wrapped in a promise).
+	 *
+	 * @abstract
+	 */
+	async apply( paper, node, researcher ) {
+		const nrOfTokens = await researcher.doResearch( "word count", node );
```
igorschoester: This is confusing to me. After asking you already, I can formulate my confusion better.
As this is the `TestAssessment`, it would make more sense to me to also run a research called `TestResearch` or something along those lines. Then resolving in a result that is more `bad` / `neutral` / `good` instead of the specifics.

spec/tree/assess/TreeAssessorSpec.js
```diff
@@ -0,0 +1,227 @@
...
+	describe( "assess", () => {
+		it( "assesses a paper and node", ( done ) => {
+			const researcher = new TreeResearcher();
+			const research = new TestResearch();
+			researcher.addResearch( "word count", research );
+
+			const scoreAggregator = new TestAggregator();
+			const assessments = [
+				// Gives back a 'bad' result when word count < 8.
+				new TestAssessment( true, 8, "word count" ),
```
igorschoester: The same name as the research is slightly confusing here.

spec/tree/assess/TreeAssessorSpec.js
```diff
@@ -0,0 +1,227 @@
...
+			const assessor = new TreeAssessor( {
+				researcher,
+				scoreAggregator,
+				i18n,
+				assessments,
+			} );
+
+			expect( assessor._assessments ).toHaveLength( 2 );
+
+			const removedAssessment = assessor.removeAssessment( "dancing potatoes" );
```
igorschoester: 🥔 🍠 🥔

src/tree/assess/assessments/Assessment.js
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @memberOf module:tree/assess
+ *
+ * @abstract
+ */
+class Assessment {
+	/**
+	 * Creates a new assessment.
+	 *
+	 * @param {string} [name] An optional name.
```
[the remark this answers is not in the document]
hansjovis: Yep, we need to add JSDocs to these properties inside of the constructor, unfortunately 😞 .

src/tree/assess/TreeAssessor.js
```diff
@@ -0,0 +1,142 @@
...
+		 */
+		this.researcher = options.researcher;
+		/**
+		 * Aggregates the scores on the individual assessments into one overall score.
+		 * @type {module:tree/assess.ScoreAggregator}
+		 */
+		this.scoreAggregator = options.scoreAggregator;
+		/**
+		 * The list of assessments to apply.
+		 * @type {module:tree/assess.Assessment[]|Array}
```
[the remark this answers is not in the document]
hansjovis: I think this got automatically added by PHPStorm.
Will take a look!

spec/specHelpers/tree/TestAssessment.js
hansjovis: Going to change this into a more general name that reflects that these are functions used for testing!

src/tree/assess/assessments/Assessment.js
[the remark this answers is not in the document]
hansjovis: I made `name` a required constructor argument.
This makes sense, since you should be able to retrieve an assessment by its name using the `TreeAssessor`, even when you give the assessment as a `TreeAssessor` constructor argument.

src/tree/assess/TreeAssessor.js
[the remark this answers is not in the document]
hansjovis: I decided to remove the `|Array` part, since it does not add anything to the type description.

src/tree/assess/TreeAssessor.js
```diff
@@ -0,0 +1,142 @@
...
+		/**
+		 * A Jed object to use for translations.
+		 * @type {Jed}
+		 */
+		this.i18n = options.i18n;
+		/**
+		 * Supplies the assessments with researches and their (cached) results.
+		 * @type {module:tree/research.TreeResearcher}
+		 */
+		this.researcher = options.researcher;
```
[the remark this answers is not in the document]
hansjovis: Since the researcher is (for now) only used inside of the `assess` function, I would vote for giving it as an extra parameter in the `assess` function. Rather than giving it as a constructor argument.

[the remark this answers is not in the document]
hansjovis: We (@igorschoester and I) discussed this with @atimmer and we decided to give it as a constructor argument in the `TreeAssessor` as well as the `Assessment`.

This is conceptually the right thing to do, since we only have one researcher available at any given time.

src/tree/assess/TreeAssessor.js
```diff
@@ -0,0 +1,142 @@
...
+		 */
+		this._assessments = options.assessments || [];
+	}
+
+	/**
+	 * Returns the list of available assessments.
+	 *
+	 * @returns {module:tree/assess.Assessment[]} The list of all available assessments.
+	 */
+	getAvailableAssessments() {
```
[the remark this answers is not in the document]
hansjovis: I am not renaming this for now, because of possible backward compatibility reasons, like you mentioned.

src/tree/assess/assessments/Assessment.js
```diff
@@ -9,16 +9,42 @@ class Assessment {
...
-		 * This assessment's unique name.
+		 * This assessment's name.
 		 * @type {string}
 		 */
 		this.name = name;
+		/**
+		 * The researcher to do researches with.
+		 * @type {module:tree/research.TreeResearcher}
+		 */
+		this.researcher = researcher;
```
igorschoester: Since this has a getter and a setter, I think it makes more sense to prefix this with an underscore.

src/tree/assess/TreeAssessor.js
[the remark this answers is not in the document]
atimmer: There are no backward compatibility concerns, because we are introducing a new class. Because of the async methods the places where an `Assessor` is currently used this cannot be replaced by the `TreeAssessor`, so this class is completely new.

src/tree/assess/assessments/Assessment.js
hansjovis: Done! 🙂

src/tree/assess/TreeAssessor.js
```diff
@@ -57,19 +59,49 @@ class TreeAssessor {
...
+	 *
+	 * @returns {Promise<AssessmentResult>} The result of the assessment.
+	 */
+	async applyAssessment( assessment, paper, node ) {
+		return assessment.apply( paper, node ).catch(
+			() => {
+				return new AssessmentResult( {
+					text: this.i18n.sprintf(
+						/* Translators: %1$s expands to the name of the assessment. */
+						this.i18n.dgettext( "js-text-analysis", "An error occurred in the '%1$s' assessment" ) ),
```
igorschoester: You should probably also add the assessment's name to the `sprintf` call then 😉

src/tree/assess/TreeAssessor.js
```diff
@@ -57,19 +59,49 @@ class TreeAssessor {
...
+		  and returning the results and final score.
 		 */
 		const results = await Promise.all(
-			applicableAssessments.map( assessment => assessment.apply( paper, node ) )
+			applicableAssessments.map( assessment => this.applyAssessment( assessment, paper, node ) )
+		);
+		// Filter out errored assessments.
+		const validResults = results.filter( result => result.getScore() !== -1 );
+		// Compute overall score.
+		const overallScore = this.scoreAggregator.aggregate( validResults );
```
igorschoester: Why is this called `overallScore` and not `score`? Like the key you return it with.

src/tree/assess/TreeAssessor.js
hansjovis: Good catch 😅 .

manuelaugustin:
CR of the latest commits & acceptance 👍

src/tree/assess/TreeAssessor.js
```diff
@@ -0,0 +1,188 @@
...
+	 * Applies the given assessment to the paper-node combination.
+	 *
+	 * @param {module:tree/assess.Assessment} assessment The assessment to apply.
+	 * @param {Paper}                         paper      The paper to apply the assessment to.
+	 * @param {module:tree/structure.Node}    node       The root node of the tree to apply the assessment to.
+	 *
+	 * @returns {Promise<AssessmentResult>} The result of the assessment.
+	 */
+	async applyAssessment( assessment, paper, node ) {
+		return assessment.apply( paper, node ).catch(
```
atimmer, line 92: Small tidbit: When using `async` & `await` you can use `try`&`catch` if you want instead of `.catch`.

manuelaugustin:
Fixes https://github.com/Yoast/javascript/issues/413

PR #2150, opened 2019-02-15
Add a stemmer for English

### LINKED ISSUE #2149: Add a stemmer function to the English morphology
In order to morphologize the internal linking suggestions (#2139) we need to add a function that would return a stem/base form of an input English word. 

Parent [Yoast/wordpress-seo-premium#2235](https://github.com/Yoast/wordpress-seo-premium/issues/2235)

nataliashitova:
## Summary

* [not-user-facing] Adds a stemmer function to the English morphology.

*

## Test instructions

* Try adding more cases to the functions' specs.

Fixes #2149


src/morphology/english/getStem.js
```diff
@@ -0,0 +1,157 @@
...
+			return buildOneFormFromRegex( baseIfPluralNoun, createRulesFromMorphologyData( regexVerb.ingFormToInfinitive ) );
+		}
+		return baseIfPluralNoun;
+	}
+
+	// Check if the word ends with "ic", "ical" or "ically". Return the "ical" form for consistency.
+	const regexAdjective = morphologyData.adjectives.regexAdjective;
+	const baseIfIcally = buildOneFormFromRegex( word, createRulesFromMorphologyData( regexAdjective.icallyToBase ) );
+	if ( ! isUndefined( baseIfIcally ) ) {
+		return baseIfIcally;
```
igorschoester: This if path is not covered by a test.

src/morphology/english/getStem.js
```diff
@@ -0,0 +1,157 @@
...
+ * Returns the stem of the input word using the morphologyData (language-specific).
+ *
+ * @param   {string} word           The word to get the stem for.
+ * @param   {Object} morphologyData The available morphology data per language (false if unavailable).
+ *
+ * @returns {string} Stemmed (or base) form of the word.
+ */
+export function getStem( word, morphologyData ) {
+	if ( ! morphologyData ) {
+		return word;
```
igorschoester: This if path is not covered by a test.

src/morphology/english/getStem.js
```diff
@@ -0,0 +1,157 @@
...
+	const possibleRegularBases = [];
+
+	// Verbal infinitive.
+	const sFormToInfinitiveRegex = createRulesFromMorphologyData( regexVerb.sFormToInfinitive );
+	const ingFormToInfinitiveRegex = createRulesFromMorphologyData( regexVerb.ingFormToInfinitive );
+	const edFormToInfinitiveRegex = createRulesFromMorphologyData( regexVerb.edFormToInfinitive );
+
+	const baseIfVerb = getInfinitive( word, sFormToInfinitiveRegex, ingFormToInfinitiveRegex, edFormToInfinitiveRegex ).infinitive;
+
+	if ( ! isUndefined( baseIfVerb ) ) {
```
igorschoester: The else path is not covered by a test.

src/morphology/english/getStem.js
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	// Adjectival base.
+	const comparativeToBaseRegex = createRulesFromMorphologyData( regexAdjective.comparativeToBase );
+	const superlativeToBaseRegex = createRulesFromMorphologyData( regexAdjective.superlativeToBase );
+	const adverbToBaseRegex = createRulesFromMorphologyData( regexAdjective.adverbToBase );
+	const lyExceptions = morphologyData.adjectives.stopAdverbs;
+
+	const baseIfAdjective = getBase( word, comparativeToBaseRegex, superlativeToBaseRegex, adverbToBaseRegex, lyExceptions ).base;
+	if ( ! isUndefined( baseIfAdjective ) ) {
```
igorschoester: The else path is not covered by a test.

spec/morphology/english/getStemEnglishSpec.js
```diff
@@ -0,0 +1,146 @@
+import { getStem, getRegularStem, getIrregularStem, getFirstAndShortest } from  "../../../src/morphology/english/getStem.js";
+import morphologyData from "../../../premium-configuration/data/morphologyData.json";
+const morphologyDataEN = morphologyData.en;
```
igorschoester: This can be combined in the import statement like so: `import { en as morphologyDataEN } from "../../../premium-configuration/data/morphologyData.json";
`

src/morphology/english/getStem.js
```diff
@@ -0,0 +1,157 @@
...
+ */
+export function getFirstAndShortest( array ) {
+	array = uniq( flatten( array ) );
+
+	// Sort alphabetically.
+	array.sort();
+
+	// Sort from shortest to longest.
+	array.sort( function( a, b ) {
+		return a.length - b.length;
```
igorschoester: Two sorts can be merged into one, something like this:

```js
	array.sort( function( a, b ) {
		// Sort from shortest to longest.
		const i = a.length - b.length;
		// Otherwise sort alphabetically.
        return i === 0 ? a.localeCompare(b) : i;
	} );
```

Or, because of only one result being the interest, not use a sort and just keep track of the shortest and closest to `a` in a for loop.

```js
export function getFirstAndShortest( array ) {
	const strings = flatten( array );
	let result = strings.pop();

	strings.forEach( str => {
		const lengthDifference = str.length - result.length;
		if ( lengthDifference === 0 ) {
			if ( str.localeCompare( result ) < 0 ) {
				result = str;
			}
		} else if ( lengthDifference < 0 ) {
			result = str;
		}
	} );

	return result;
}
```

src/morphology/english/getStem.js
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * Gets the shortest of the alphabetically ordered strings from an array.
+ *
+ * @param {string[]} array  The array of strings.
+ *
+ * @returns {string} The shortest of the alphabetically ordered strings from the input array.
+ */
+export function getFirstAndShortest( array ) {
+	array = uniq( flatten( array ) );
```
igorschoester: Two comments for this line:

- You are changing the variable passed to this function. This is hidden behavioras it will have consequences for the variable outside of the function with an array or object, like in this case. Generally better to avoid.
- `uniq` is not needed if you sort later and just return the first result. Or are you expecting a lot of duplicates here?

src/morphology/english/getStem.js
```diff
@@ -0,0 +1,157 @@
+import { buildOneFormFromRegex } from "../morphoHelpers/buildFormRule";
+import createRulesFromMorphologyData from  "../morphoHelpers/createRulesFromMorphologyData.js";
+import { getBase } from "./getAdjectiveForms";
+import { getInfinitive, checkIrregulars as getIrregularVerbParadigm, endsWithIng } from "./getVerbForms.js";
+
+import { isUndefined, uniq } from "lodash-es";
+import { flatten } from "lodash-es";
```
igorschoester: Why not merge this in the previous `lodash-es` import?

src/morphology/english/getStem.js
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * Gets the shortest of the alphabetically ordered strings from an array.
+ *
+ * @param {string[]} array  The array of strings.
+ *
+ * @returns {string} The shortest of the alphabetically ordered strings from the input array.
```
igorschoester: What does it return / do when the input is an empty array?

src/morphology/english/getStem.js
```diff
@@ -0,0 +1,157 @@
...
+ * Gets possible stems as a regular noun, adjective and verb.
+ *
+ * @param{string} word              The word for which to determine its base.
+ * @param{Object} morphologyData    The morphology data for the language.
+ *
+ * @returns{string} The shortest and the alphabetically-first of possible noun-like, verb-like and adjective-like bases.
+ */
+export function getRegularStem( word, morphologyData ) {
+	// Try to singularize as a noun.
+	const regexVerb = morphologyData.verbs.regexVerb;
```
igorschoester: Does this path in `morphologyData` always exist in this context?

src/morphology/english/getStem.js
```diff
@@ -0,0 +1,157 @@
...
+import { flatten } from "lodash-es";
+
+/**
+ * Gets the shortest of the alphabetically ordered strings from an array.
+ *
+ * @param {string[]} array  The array of strings.
+ *
+ * @returns {string} The shortest of the alphabetically ordered strings from the input array.
+ */
+export function getFirstAndShortest( array ) {
```
igorschoester: I think this name is a bit unclear in what it does. It does more than just `get`, maybe `determine` or `find`. `First` is confusing for an array. `Alphabetical` would be more precise.
What do you think of `determineAlphabeticalShortest`?

src/morphology/english/getStem.js
```diff
@@ -0,0 +1,157 @@
...
+ * @returns {string} The base form of the irregular word.
+ */
+export function getIrregularStem( word, irregulars, verbMorphology ) {
+	let baseIfIrregular = "";
+
+	irregulars.forEach( function( paradigm ) {
+		if ( baseIfIrregular === "" ) {
+			paradigm.forEach( function( wordInParadigm ) {
+				if ( wordInParadigm === word ) {
+					baseIfIrregular = paradigm[ 0 ];
```
igorschoester: Why is the loop continuing after a `baseIfIrregular` has been found?
You can use an old-fashioned for-loop to break out of.

This code will make `getIrregularStem` return the match, on the first match.
There might be a need for an empty string check. I am not sure if that could be in the `irregulars`.
```js
for( let i = 0; i < irregulars.length; i++ ) {
	for( let j = 0; j < irregulars[ i ].length; j++ ) {
		if ( irregulars[ i ][ j ] === word ) {
			return paradigm[ 0 ];
		}
	}
}
```

src/morphology/english/getStem.js
nataliashitova: Looks great! Thanks 😄

src/morphology/english/getStem.js
nataliashitova: Fair points. I removed `uniq`. I pretty much expect to have maximally two duplicates, so it's not worth it to run `uniq` every time.

src/morphology/english/getStem.js
nataliashitova: Now as `uniq` is removed from the first import statement, both imports are about a single function. I find it more readable with separate imports (and from what I heard from Anton separate imports are no better and no worse than one pooled import).

src/morphology/english/getStem.js
nataliashitova: Fair point. I added a check.

src/morphology/english/getStem.js
nataliashitova: Yes.

src/morphology/english/getStem.js
nataliashitova: After a long discussion changed it to `findShortestAndAlphabeticallyFirst`.

src/morphology/english/getStem.js
nataliashitova: Good point! Thanks 😄 

There are no empty strings in the irregulars.

src/morphology/english/determineStem.js
```diff
@@ -13,17 +13,26 @@ import { flatten } from "lodash-es";
  *
  * @returns {string} The shortest of the alphabetically ordered strings from the input array.
  */
-export function getFirstAndShortest( array ) {
-	array = uniq( flatten( array ) );
-
-	// Sort alphabetically.
-	array.sort();
+export function findShortestAndAlphabeticallyFirst( array ) {
+	if ( array.length === 0 ) {
```
igorschoester: This is not needed with that implementation, here is why:
`[].pop()` returns `undefined`.
`[].forEach()` just doesn't loop.
The return in the end will return the `undefined` result. Which is the same thing this if statement is doing.

src/morphology/english/determineStem.js
```diff
@@ -13,17 +13,26 @@ import { flatten } from "lodash-es";
  *
  * @returns {string} The shortest of the alphabetically ordered strings from the input array.
```
igorschoester: It can return `undefined` on an empty array too.
Maybe you'd want to return an empty string then like in `determineIrregularStem`.
I'm not sure which is better, it depends if you need to be able to check if that happened or not. And if an empty string works fine for code using that as a valid result.
Either way, it is probably nice to be consistent here.

spec/morphology/english/determineStemEnglishSpec.js
```diff
@@ -0,0 +1,172 @@
...
+describe( "findShortestAndAlphabeticallyFirst", function() {
+	it( "returns the shortest and the alphabetically-first word from an array", function() {
+		expect( findShortestAndAlphabeticallyFirst( [ "d", "f", "a", "c", "b" ] ) ).toEqual( "a" );
+		expect( findShortestAndAlphabeticallyFirst( [ "d", "f", "ab", "c", "b" ] ) ).toEqual( "b" );
+		expect( findShortestAndAlphabeticallyFirst( [ "wording", "words", "word", "worded", "worder" ] ) ).toEqual( "word" );
+		expect( findShortestAndAlphabeticallyFirst( [ "worded", "worder", "wording" ] ) ).toEqual( "worded" );
+		expect( findShortestAndAlphabeticallyFirst( [ "word" ] ) ).toEqual( "word" );
+	} );
+
+	it( "returns null if the input array is empty", function() {
```
igorschoester:
```suggestion
	it( "returns undefined if the input array is empty", function() {
```

src/morphology/english/determineStem.js
nataliashitova: Fair point!

src/morphology/english/determineStem.js
nataliashitova: After serious consideration, I decided I want to return `undefined` if any of the sub-functions fails to return a string base. 
I think it's much more valid than to return an empty string, because we definitely don't want empty strings to invade databases. `getInfinitive` and `getBase` functions used in `determineRegularStem` both return the input word as default, which makes them pretty safe and the chances to get an empty array as input to `findShortestAndAlphabeticallyFirst` and to return `underfined` by `defineStem` indefinitely close to 0. And if it ever happens I want to get an error and not to stuff empty strings in the databases.

src/morphology/english/determineStem.js
```diff
@@ -0,0 +1,154 @@
...
+	for ( let i = 0; i < irregulars.length; i++ ) {
+		const paradigm = irregulars[ i ];
+		for ( let j = 0; j < paradigm.length; j++ ) {
+			if ( paradigm[ j ] === word ) {
+				return paradigm[ 0 ];
+			}
+		}
+	}
+
+	if ( verbMorphology ) {
```
hansjovis: I would move the logic within this `if` statement into a separate function.
When looking at the variables that are manipulated, I see that there is little overlap.
(e.g. the code block above uses `word` and `irregulars`, this code block uses `word` and `verbMorphology`).

src/morphology/english/determineStem.js
```diff
@@ -0,0 +1,154 @@
...
+	// If the word is a possessive, it can only be a noun or an ing-noun.
+	if ( ! isUndefined( baseIfPossessive ) ) {
+		bestBase = baseIfPossessive;
+
+		// Only check for nouns.
+		irregulars = nounMorphology.irregularNouns;
+		verbMorphology = false;
+	}
+
+	const baseIfIrregular = determineIrregularStem( bestBase, irregulars, verbMorphology );
```
hansjovis: If you split the `determineIrregularStem` function, I would change the function like this:
( I may got carried away 😅 ).
```js
if ( ! morphology ) {
     return word;
}

const nounIrregulars = morphologyData.nouns.irregularNouns;
const adjectiveIrregulars = morphologyData.adjectives.irregularAdjectives;

const baseIfPossessive = buildOneFormFromRegex( word, createRulesFromMorphologyData( nounMorphology.regexNoun.possessiveToBase ) );

let stem;

// Determine if this is an irregular noun, adjective or verb. 
if ( ! isUndefined( baseIfPossessive ) ) {
    // The word is a possessive, it can only be a noun or an ing-noun; 
    stem = determineIrregularStem( baseIfPossessive, nounIrregulars );
} else {
    // Word can be a noun, adjective or verb.
    stem = determineIrregularStem( word, nounIrregulars ) || 
           determineIrregularStem( word, adjectiveIrregulars ) || 
           determineIrregularVerbStem( word, verbMorphology );
}
if ( stem ) {
    // Found the stem of an irregular word! Return it.
    return stem;
}

// Treat the word as a regular noun, adjective or verb.

if ( ! isUndefined( baseIfPossessive ) ) {
     // Possessive: treat it as a regular noun or ing-noun.
     return determineRegularStem( baseIfPossessive, morphologyData );
} else {
     return determineRegularStem( word, morphologyData ); 
}
```

src/morphology/english/getAdjectiveForms.js
```diff
@@ -80,8 +80,8 @@ const endsWithEst = function( word ) {
  */
 const endsWithLy = function( word ) {
 	const wordLength = word.length;
-	// Consider only words of four letters or more to be adjectives (otherwise, words like "lily" are being treated as adjectives).
-	if ( wordLength > 3 ) {
+	// Consider only words of five letters or more to be adjectives (otherwise, words like "lily" are being treated as adjectives).
+	if ( wordLength > 4 ) {
```
hansjovis: Small suggestion. This would keep it consistent with the comment (`five letters or more`). 
```suggestion
	if ( wordLength >= 5 ) {
```

src/morphology/english/getVerbForms.js
```diff
@@ -164,9 +164,9 @@ const endsWithS = function( word ) {
  * @returns {boolean} True if the word ends with "ing".
  */
 const endsWithIng = function( word ) {
-	const wordLength = word.length;
-	// Consider only words of five letters or more to be ing forms (otherwise, words like "ping" are being treated as verb forms).
-	if ( wordLength > 4 ) {
+	const vowelCount = ( word.match( /([aeiouy])/g ) || [] ).length;
```
hansjovis: If the regex is defined as a constant at the top of the file, it has to be compiled only once.
Right now it is defined and compiled every time this function is called.
This could save some time when calling this function repeatedly. (Or I am just being paranoid).

src/morphology/english/getVerbForms.js
```diff
@@ -180,9 +180,9 @@ const endsWithIng = function( word ) {
  * @returns {boolean} True if the word ends with "ed".
  */
 const endsWithEd = function( word ) {
-	const wordLength = word.length;
-	// Consider only words of four letters or more to be past forms (otherwise, words like "red" are being treated as verb forms).
-	if ( wordLength > 3 ) {
+	const vowelCount = ( word.match( /([aeiouy])/g ) || [] ).length;
```
hansjovis: Same here, you can reuse the definition of this regex at the top of the file.
Another advantage: if you define it only once, you only have to change it in one place (if it ever changes).

src/morphology/english/determineStem.js
nataliashitova: 👍 Fair point!

src/morphology/english/determineStem.js
nataliashitova: Nice! 👍

src/morphology/english/getVerbForms.js
nataliashitova: indeed!

src/morphology/english/getVerbForms.js
nataliashitova: Indeed!

src/morphology/english/getAdjectiveForms.js
nataliashitova: 👍

src/morphology/english/determineStem.js
```diff
@@ -51,12 +50,22 @@ export function determineIrregularStem( word, irregulars, verbMorphology ) {
...
+ *
+ * @param {string}          word            The word for which to determine its base.
+ * @param {Object|false}    verbMorphology  Regexes and irregulars for verb morphology, False if verb rules should not be applied.
+ *
+ * @returns {string|undefined} The base form of the irregular word; undefined if no irregular stem was found.
+ */
+export function determineIrregularVerbStem( word, verbMorphology ) {
+	const paradigmIfIrregularVerb = getIrregularVerbParadigm( word, verbMorphology.irregularVerbs, verbMorphology.regexVerb.verbPrefixes );
+	if ( ! isUndefined( paradigmIfIrregularVerb ) ) {
+		return paradigmIfIrregularVerb[ 0 ];
```
hansjovis, line 69: It is better to explicitly return `null` if something is not found, rather than rely on `undefined` being returned when no `return` statement is encountered.

Going to fix this myself, since it is the only critique I have and we need this PR! 😄

hansjovis:
Acceptance test: ✅ 

Only word I tested that was not stemmed correctly was `dormice`.
It was stemmed to `dormice`, instead of `dormouse` 😉.
Will make an issue for adding it to the list of irregular nouns.

PR #2153, opened 2019-02-19
Add functionality to recognize German verb forms

### LINKED ISSUE #2132: Add word form recognition for German verbs
We should be able to recognize verbs in their different morphological forms in German.

This will require:

- [x] Add regular suffixes to the data
- [x] Add functionality to add regular suffixes in the default pipeline; regular suffixes need to be added dynamically, taking into account the ending of the stem

The following exceptions need to be implemented
- [ ] -el/-le stem change verbs
- [x] Strong verbs
- [ ] Irregular mixed verbs (make sure to check how the stemmer stems these irregular forms)
- [ ] sein/haben - make sure that all forms of these words are in the list of function words; that means that we don't need to take them into account when creating word forms

Past participles
- [ ] If past participle occurs in keyphrase, reconstruct the stem from the participle form
- [x] Construct a participle form from the stem found [partly implemented; some exception cases still require implementation]
- [ ] Separable verbs: construct all single word forms; this means construct the infinitive from the past participle and vice versa; possibly also construct subjunctive clause forms.

In `getForms`:
- [x] Apply regular suffixes

Stretch goals:
- Recognize separable verbs, e.g. "losgehen" -> "geht los". Splitting these words is the easier part, but we need to change the overall morphology pipeline: at the moment, we build forms word-by-word. For this functionality, we need to take into account the whole keyphrase. Also when matching, a match should only occur when both parts of a separable verb are found.

Parent #2095

manuelaugustin:
## Summary

* [not-user-facing] Adds basic functionality to create German verb forms.

* This PR includes basic verb functionality. This includes adding regular suffixes and generating forms for strong verbs. Extra functionality will be added in later PRs.

## Test instructions

* Make sure you have the premium configuration set to the right branch (see https://github.com/Yoast/YoastSEO.js-premium-configuration/pull/10).
* Change the language of your site to German.
* Add a German text and a keyphrase including German verbs.
* Add different grammatical forms of the verbs in your text; make sure these forms are recognized. For this, you use forms of a specific word from a dictionary, e.g. https://de.wiktionary.org/wiki/Flexion:kaufen

Fixes #2132


src/morphology/morphoHelpers/suffixHelpers.js
```diff
@@ -0,0 +1,27 @@
...
+ * Creates word forms by appending all given suffixes on each of the given stems.
+ *
+ * @param {string[]}    stems               The stems on which to apply the suffixes.
+ * @param {string[]}    suffixes            The suffixes to apply.
+ * @param {string}      [appendToStem=""]   Optional material to append between the stem and the suffixes.
+ *
+ * @returns {string[]} The suffixed verb forms.
+ */
+export function applySuffixesToStems( stems, suffixes, appendToStem = "" ) {
+	return flattenDeep( stems.map( stem => applySuffixesToStem( stem, suffixes, appendToStem ) ) );
```
hansjovis: I propose replacing this function with this:
```js
return stems.reduce( ( stem, list ) => {
  const stemWithSuffixes = applySuffixesToStem( stem, suffixes, appendToStem );
  return list.concat( stemsWithSuffixes );
}, [] );
```
This would loop over the list once, without the need of a second `flattenDeep` loop.
(This would only work when `stemsWithSuffixes` is always one level deep, but my guess is it is).

src/morphology/german/generateParticipleForm.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * Adds a prefix and a suffix to a stem in order to create a past participle form
+ *
+ * @param {string}  stemmedWord The stemmed word on which to
+ * @param {Object}  affixes     The suffix and prefix data.
+ *
+ * @returns {string} The participle form.
+ */
+const addParticipleAffixes = function( stemmedWord, affixes ) {
+	return affixes.prefix.concat( stemmedWord ).concat( affixes.suffix );
```
hansjovis: [According to MDN, the `+` operator is several times faster than using `concat` to concatenate strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/concat#Performance). 
It is also more readable:
```js
return affixes.prefix + stemmedWord + affixes.suffix;
```
Can you check if this is done anywhere else in this PR as well?

src/morphology/morphoHelpers/suffixHelpers.js
```diff
@@ -0,0 +1,27 @@
...
+ * Creates word forms from a list of given suffixes and a stem.
+ *
+ * @param {string}      stem                The stem on which to apply the suffixes.
+ * @param {string[]}    suffixes            The suffixes to apply.
+ * @param {string}      [appendToStem=""]   Optional material to append between the stem and the suffixes.
+ *
+ * @returns {string[]} The suffixed verb forms.
+ */
+export function applySuffixesToStem( stem, suffixes, appendToStem = "" ) {
+	return suffixes.map( suffix => stem.concat( appendToStem ).concat( suffix ) );
```
hansjovis: See my other comment about using the `+` operator instead of `concat` to concatenate strings.
`+` is several times faster.
So:
```suggestion
	return suffixes.map( suffix => stem + appendToStem + suffix );
```

src/morphology/german/generateVerbExceptionForms.js
```diff
@@ -0,0 +1,145 @@
...
+ *
+ * @returns {string[]} The created verb forms.
+ */
+const generateFormsPerParadigm = function( morphologyDataVerbs, paradigm, stemmedWordToCheck ) {
+	const verbClass = paradigm.class;
+	const stems = paradigm.stems;
+
+	let stemsFlattened = [];
+
+	forOwn( stems, ( stem ) => stemsFlattened.push( stem ) );
```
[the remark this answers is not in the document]
manuelaugustin: So apparently calling `flatten` on a string (some of the stems are strings, some are arrays) will make an array with the individual letters (e.g., `kling` -> ` [ 'k', 'l', 'i', 'n', 'g' ]`) so I think I'll stick to the original version 😅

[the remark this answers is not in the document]
manuelaugustin: (I did change `flattenDeep` to `flatten` however, since that will also do the job in this case.)

src/morphology/german/generateVerbExceptionForms.js
```diff
@@ -0,0 +1,145 @@
...
+ * @param {string}  verbClass       The verb class of the paradigm.
+ * @param {Object}  stems           The stems of the paradigm.
+ *
+ * @returns {string[]} The created forms.
+ */
+const addSuffixesStrongVerbParadigm = function( dataStrongVerbs, verbClass, stems ) {
+	// All classes have the same present and participle suffixes.
+	const suffixes = {
+		present: dataStrongVerbs.suffixes.presentAllClasses.slice(),
+		pastParticiple: new Array( dataStrongVerbs.suffixes.pastParticiple.slice() ),
```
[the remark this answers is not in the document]
manuelaugustin: Actually the `slice` was incorrectly used here because the value on which it is applied is a string.

src/morphology/german/generateParticipleForm.js
manuelaugustin: Nice find! I changed it everywhere in this PR and I'll make sure to use assignment operators instead of `concat` in the future.

manuelaugustin:
Thanks for the tips! Now the code is quite a bit nicer/more efficient! 🙌

src/morphology/german/generateVerbExceptionForms.js
```diff
@@ -12,41 +12,35 @@ import { flattenDeep, forOwn, uniq as unique } from "lodash-es";
...
 	// Add class-specific suffixes.
-	forOwn( suffixesPerClass, function( stemData, stemClass ) {
-		if ( stemClass === verbClass ) {
-			forOwn( stemData, function( additionalSuffixes, suffixClass ) {
-				suffixes[ suffixClass ] = additionalSuffixes;
-			} );
-		}
-	} );
+	const additionalSuffixes = dataStrongVerbs.suffixes.classDependent[ verbClass ];
+	const allSuffixes = { ...basicSuffixes, ...additionalSuffixes };
```
hansjovis: 👍

nataliashitova:
Acceptance 👍
I've added a bunch of cases that do not work so far to https://github.com/Yoast/YoastSEO.js/issues/2160

PR #2157, opened 2019-02-20
Add language list for morphology

### LINKED ISSUE #2156: Add a list of morphologized languages
Add a list of morphologized languages; similar to the function words.

Export this list so we can use it in Premium (to prevent 404 requests).

Parent https://github.com/Yoast/YoastSEO.js-premium-configuration/issues/11

igorschoester:
## Summary

* [not user-facing] Adds a helper that returns a list of languages that have word forms support

* Export this function in the helpers' index. Which is loaded in the root index

## Test instructions

* Run the tests
* When linked to the plugin you can run the following in the console `yoast.analysis.helpers.getFormsLanguages()` and it should return an array with just `en` inside. It being there tests it got exported. English there matches the currently only available morphology language.

Fixes #2156


manuelaugustin:
I find `getFormsLanguages` a bit ambiguous. Could we maybe change it to something more descriptive, such as `getLanguagesWithWordFormSupport`? (Bonus points if you can think of something more descriptive that's not as long as my suggestion 😅)

manuelaugustin:
CR & acceptance done 👍

PR #2167, opened 2019-02-25
Change sentence length limit for Spanish

### LINKED ISSUE #2166: Increase sentence length limit for Spanish
A user has reported to us that she found it very difficult to write sentences under the recommended sentence length limit of 20 words in Spanish and Catalan. She suggested changing the limit to 25 words. Following this, a re-assessment of the Spanish sentence length limit was made.

The [guidelines](https://github.com/Yoast/research/wiki/Sentence-Length-Spanish) for Spanish seem to lean towards a maximum of 20 words as ideal. 

However, after comparing Italian, Spanish, and English versions of a few texts, the sentence length of the Spanish versions was usually closer to that of the Italian versions (for which we set the sentence length limit to 25) than to English (for which the limit is 20). 

There also seems to be a consensus from different sources that Spanish sentences are longer than English, e.g. [1](http://linguistlist.org/ask-ling/message-details1.cfm?asklingid=200433042), [2](https://english.stackexchange.com/questions/2998/do-most-languages-need-more-space-than-english), [3](https://www.transfluent.com/en/2015/07/why-spanish-uses-more-words-than-english-an-analysis-of-expansion-and-contraction/). 

Given these insights, I think it would be reasonable to change the sentence length limit to 25 for Spanish.

agnieszkaszuba:
## Summary

* Increases the recommended sentence length limit for Spanish, props to [Sílvia Fustegueres/Ampersand Translations](https://www.ampersand.net/en/).

## Test instructions

* In your local testing environment, change the locale to es_ES. 
* Paste a Spanish text.
* Make sure that the correct feedback is returned (e.g. the recommended limit is specified as 25; different results are returned than when locale is set to a language with another limit, such as English).

Fixes #2166

### DIFF
--- a/src/config/content/es.js
+++ b/src/config/content/es.js
@@ -1,0 +1,5 @@
+export default {
+	sentenceLength: {
+		recommendedWordCount: 25,
+	},
+};
--- a/src/config/content/es.js
+++ b/src/config/content/es.js
@@ -1,3 +1,9 @@
+/**
+ * Spanish sentences tend to be longer than English ones (for which the recommended limit is 20 words).
+ * This is why the sentence length limit was increased to 25 for Spanish. For more detailed explanation, see:
+ * https://github.com/Yoast/research/wiki/Sentence-Length-Spanish
+ */
+
 export default {
 	sentenceLength: {
 		recommendedWordCount: 25,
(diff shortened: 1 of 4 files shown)


src/config/content/es.js
```diff
@@ -0,0 +1,5 @@
+export default {
+	sentenceLength: {
+		recommendedWordCount: 25,
```
igorschoester, line 9: It would be really nice if you would document the reasoning behind this change here (basically, the text of the issue). So that we have less magical numbers in the code.

igorschoester (changes requested):
CR 🏗 

It would be really nice if you would document the reasoning behind this change here (basically, the text of the issue). So that we have less magical numbers in the code.

igorschoester (approved):

igorschoester:
Acceptance 👍

PR #2169, opened 2019-02-26
Increase sentence length limit for Catalan

### LINKED ISSUE #2168: Increase sentence length limit for Catalan
We should increase the sentence length limit for Catalan to 25 words. See research: https://github.com/Yoast/research/wiki/Sentence-Length-Catalan

agnieszkaszuba:
## Summary

* Increases the recommended sentence length limit for Catalan, props to [Sílvia Fustegueres](https://www.ampersand.net/en/).

## Test instructions

* In your local testing environment, change the locale to ca_ES. 
* Paste a Catalan text.
* Make sure that the correct feedback is returned (e.g. the recommended limit is specified as 25, and the feedback is different than when locale is set to a language with another limit, such as English).

Fixes #2168

### DIFF
--- a/src/config/content/ca.js
+++ b/src/config/content/ca.js
@@ -1,0 +1,10 @@
+/**
+ * The default limit of 20 words is too strict for Catalan. For more detailed explanation, see:
+ * https://github.com/Yoast/research/wiki/Sentence-Length-Catalan
+ */
+
+export default {
+	sentenceLength: {
+		recommendedWordCount: 25,
+	},
+};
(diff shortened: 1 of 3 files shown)


src/config/content/ca.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * The default limit of 20 words is too strict for Catalan. For more detailed explanation, see:
```
hansjovis: Comments that focus on the _why_ rather than the _what_ are often better.
You can tell by inspecting the code that the default word count is 20 and the new count is 25, but you cannot tell why this is the case 😉.
Perhaps this is a better description:
```suggestion
 * Sentences in Catalan have a similar length distribution as in Spanish, so use the same 25 word limit. For a more detailed explanation, see:
```

hansjovis (changes requested):

hansjovis (approved):
👍

manuelaugustin:
Acceptance 👍

PR #2173, opened 2019-02-27
Add SEOScoreAggregator

### LINKED ISSUE #2124: Create SEOScoreAggregator
This aggregator should mimic the behavior that is currently present in the `Assessor`. The aggregator only has one method `async aggregate( results )` that returns the overall score for the given list of results. Obviously the class can have more method if necessary to facilitate this method, but that method is required.

Requires #2123 

Parent #1903

igorschoester:
## Summary

* [non-user-facing] Adds a SEOScoreAggregator to be used in the TreeAssessor.

* Uses the same calculation as `src/assessor.js`.
* Rearranges the files under `tree/assess` and creates index files for each folder.

## Test instructions

It is not actually used yet. You could compare scores of the current assessor with the scores of this implementation.

But short of that:
* Look at the code.
* Run the tests.

Fixes #2124


src/tree/assess/scoreAggregators/index.js
```diff
@@ -0,0 +1,2 @@
+export { default as ScoreAggregator } from "./ScoreAggregator";
```
hansjovis, line 2: Oh, nice syntax. I should remember this.

src/tree/assess/scoreAggregators/SEOScoreAggregator.js
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * The factor to multiply the amount of results with.
+ *
+ * Individual scores are from 1 to 9.
+ * The make the total score work in the 100 scale, the amount of results needs to get multiplied by this factor.
+ *
+ * @type {number}
+ * @const
+ *
+ * @memberOf module:tree/assess
```
hansjovis, line 25: I am not sure if this is needed, since it is a local variable that is not exported.
Let's discuss this when you have the time!

src/tree/assess/index.js
```diff
@@ -1,6 +1,6 @@
 import TreeAssessor from "./TreeAssessor";
-import ScoreAggregator from "./ScoreAggregator";
-import Assessment from "./assessments/Assessment";
+import * as ScoreAggregators from "./scoreAggregators";
+import * as Assessments from "./assessments";
```
hansjovis, line 3: Hmmm, need to up my import game 😛.

hansjovis (changes requested):

src/tree/assess/scoreAggregators/SEOScoreAggregator.js
igorschoester: As discussed with you. We document private things too. And even though this is not on the class directly, it is still a part of it.

igorschoester:
Still want to change the tests here, see https://github.com/Yoast/YoastSEO.js/pull/2176/files#r261361029

hansjovis (approved):
CR: 👍

hansjovis:
Acceptance test: ✅

PR #2175, opened 2019-02-27
Use language-specific morphology data in specs and dev-tool

manuelaugustin:
## Summary

* [non-user-facing] Makes it possible to load language-specific morphology data in the specs and in the dev-tool.

* To keep it simple for now, `getMorphologyData` currently imports all language-specific morphology files and then sets the right one based on the language. In the future, we should import only the specific language file that's needed.

## Test instructions

* Run `yarn test` and make sure that all unit tests pass.
* In the dev-tool, set your locale to `en_US`.
* Add an English keyphrase, e.g. `story`.
* Add a text with  >150 words, including the word `stories`.
* Make sure `stories` gets recognized.
* Set your locale to `de_DE`.
* Add a German keyphrase, e.g. `Geschichte`.
* Add a text with  >150 words, including the word `Geschichten`.
* Make sure `Geschichten` gets recognized. 

Fixes https://github.com/Yoast/YoastSEO.js-premium-configuration/issues/14


igorschoester (approved):
CR 👍 

Though of course, the nice solution would be to not import all the morphology data.
Let's update that after that solution is made for the word forms.

hansjovis:
Acceptance test: ✅

PR #2176, opened 2019-02-27
Create readability score aggregator

### LINKED ISSUE #2124: Create SEOScoreAggregator

### LINKED ISSUE #2125: Create ReadabilityScoreAggregator
This aggregator should mimic the behavior that is currently present in the `ContentAssessor`. The aggregator only has one method `async aggregate( results )` that returns the overall score for the given list of results. Obviously the class can have more method if necessary to facilitate this method, but that method is required.

Requires #2123

Parent #1903

hansjovis:
## Summary

* _[non-user-facing]_ Adds a `ReadabilityScoreAggregator` that should serve as a replacement for the score aggregation part of the current `contentAssessor` in the (near) future.

* Uses the same logic as in `contentAssessor`.
* May get some merge conflicts when #2173 is merged first, since it restructures the folder structure somewhat. These changes may not coincide with the structure in this branch.

## Test instructions

* Look at the code.
    * Specifically, check if the logic used is the same as the score aggregation functionality in `contentAssessor` (The `aggregate` method in the `ReadabilityScoreAggregator` method should implement the same logic as the `calculateOverallScore` method).
 * Run the tests.
    * Check that the coverage is 100% for the `ReadabilityScoreAggregator`.
    * Check that all tests pass.

Fixes #2125


src/tree/assess/scoreAggregators/ReadabilityScoreAggregator.js
```diff
@@ -0,0 +1,167 @@
...
+
+	/**
+	 * Calculates the total amount of penalty points based on the given assessment results.
+	 *
+	 * @param {AssessmentResult[]} results The valid results from which to calculate the total penalty points.
+	 *
+	 * @returns {number} The total penalty points for the results.
+	 */
+	calculatePenalty( results ) {
+		const penaltyPoints = results.map( result => {
```
igorschoester: The usage of the `reduce` function here would combine the `map` and `sum` 😉

src/tree/assess/scoreAggregators/ReadabilityScoreAggregator.js
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	/**
+	 * Calculates the total amount of penalty points based on the given assessment results.
+	 *
+	 * @param {AssessmentResult[]} results The valid results from which to calculate the total penalty points.
+	 *
+	 * @returns {number} The total penalty points for the results.
+	 */
+	calculatePenalty( results ) {
```
igorschoester, line 146: Is it `penalty` or `penaltyPoints`? I actually like your decision to move away from the `points` part. However, you do not use that consistently. Below they are `penaltyPoints` again, while in `calculateScore` you pass `penalty`.

src/tree/assess/scoreAggregators/ReadabilityScoreAggregator.js
```diff
@@ -0,0 +1,167 @@
...
+		const penaltyPoints = results.map( result => {
+			// Compute the rating ("error", "feedback", "bad", "ok" or "good").
+			const rating = scoreToRating( result.getScore() );
+
+			if ( this.allAssessmentsSupported( results ) ) {
```
igorschoester: What do you think of this snippet?
```js
const penalty = this.allAssessmentsSupported( results )
	? penaltyMappingFullSupport[ rating ]
	: penaltyMappingPartialSupport[ rating ];
return penalty || 0;
```
I think a ternary operator here makes it more readable. Plus the `|| 0` fallback is only needed once.

src/tree/assess/scoreAggregators/ReadabilityScoreAggregator.js
```diff
@@ -0,0 +1,167 @@
...
+/* Internal dependencies */
+import { ScoreAggregator } from "../index";
+
+/**
+ * Total number of available readability assessments.
+ *
+ * @type {number}
+ * @const
+ */
+const totalNrOfAssessments = 8;
```
igorschoester: Do we have to keep this number updated? Maybe an improvement on this would be nicer. I'm thinking a key count of the readability index? Similar to how we do this for language functionality (`getFunctionWordsLanguages`).

src/tree/assess/scoreAggregators/ReadabilityScoreAggregator.js
```diff
@@ -0,0 +1,167 @@
...
+class ReadabilityScoreAggregator extends ScoreAggregator {
+	/**
+	 * Determines whether a language is fully supported. If a language supports 8 content assessments
+	 * it is fully supported.
+	 *
+	 * @param {AssessmentResult[]} results The list of results.
+	 *
+	 * @returns {boolean} True if fully supported.
+	 */
+	allAssessmentsSupported( results ) {
```
igorschoester: I hope this name can be improved on. Maybe `hasFullSupport` or `isFullySupported`, to keep in the same line as the `penaltyMapping` constants.

src/tree/assess/scoreAggregators/ReadabilityScoreAggregator.js
```diff
@@ -0,0 +1,167 @@
...
+	 * @param {AssessmentResult[]} results The assessment results.
+	 *
+	 * @returns {number} The aggregated score.
+	 */
+	aggregate( results ) {
+		const validResults = this.getValidResults( results );
+
+		/*
+		 * If you have no content, you have a red indicator.
+		 * (Assume that one result always means the 'no content' assessment result, (see `contentAssessor`) ).
```
igorschoester: Let's not refer to soon-to-be-deprecated code in the comments?

src/tree/assess/scoreAggregators/ReadabilityScoreAggregator.js
```diff
@@ -0,0 +1,167 @@
...
+		 */
+		if ( validResults.length <= 1 ) {
+			return READABILITY_SCORES.NEEDS_IMPROVEMENT;
+		}
+
+		const penalty = this.calculatePenalty( results );
```
igorschoester: Shouldn't this be calculated from the valid results?

src/tree/assess/scoreAggregators/ReadabilityScoreAggregator.js
```diff
@@ -0,0 +1,167 @@
...
+		return this.calculateScore( this.allAssessmentsSupported( results ), penalty );
```
igorschoester: Since you saved the penalty on a variable, it seems nice to do the same with the supported variable.
Also, doesn't `results` need to be valid results?

spec/tree/assess/scoreAggregators/ReadabilityScoreAggregatorSpec.js
```diff
@@ -0,0 +1,156 @@
...
+import ReadabilityScoreAggregator from "../../../../src/tree/assess/scoreAggregators/ReadabilityScoreAggregator";
+import { READABILITY_SCORES } from "../../../../src/tree/assess/scoreAggregators/ReadabilityScoreAggregator";
+import AssessmentResult from "../../../../src/values/AssessmentResult";
+
+describe( "ReadabilityScoreAggregator", () => {
+	describe( "calculatePenalty", function() {
+		let aggregator;
+
+		beforeEach( function() {
+			aggregator = new ReadabilityScoreAggregator();
```
igorschoester, line 11: I should do it like this in the `SEOScoreAggregator` specs too.

igorschoester (changes requested):
CR 🏗 

Some questions and improvement suggestions.

src/tree/assess/scoreAggregators/ReadabilityScoreAggregator.js
hansjovis: Ah, nice one! 
I reduced the `map` and `sum` calls into one `reduce` call.

src/tree/assess/scoreAggregators/ReadabilityScoreAggregator.js
hansjovis: The code and comments now always refer to `penalty` 🙂instead of `penaltyPoints`.

src/tree/assess/scoreAggregators/ReadabilityScoreAggregator.js
hansjovis: Changed the code to use the ternary operator.

src/tree/assess/scoreAggregators/ReadabilityScoreAggregator.js
hansjovis: Yes. I think the most straight forward way is the way that you describe.
So: a file with all the readability assessments that gets
1. Imported (or defined) and used in the to-be-implemented `constructReadabilityAssessor()` function to make a new `TreeAssessor` for the readability analysis.
2. Gets imported here, with this line changing to (although it can be wrapped into a helper function, if needed):
  ```js
const totalNrOfAssessments = readabilityAssessments.length;
  ```

src/tree/assess/scoreAggregators/ReadabilityScoreAggregator.js
hansjovis: I like `isFullySupported`. Changed the method name!

src/tree/assess/scoreAggregators/ReadabilityScoreAggregator.js
hansjovis: You're right. I updated the comment to not refer to the soon-to-be-deprecated class.

src/tree/assess/scoreAggregators/ReadabilityScoreAggregator.js
hansjovis: Good catch. Changed it to `validResults`.

src/tree/assess/scoreAggregators/ReadabilityScoreAggregator.js
hansjovis: Saved the result of `isFullySupported` to a variable, which I now give as an argument to the function.

The `contentAssessor` checks if the number of applicable assessments is the same as the total number of available assessments. It does not check for validity, since it does not check the assessment results. Either way, we need to change the way language support is checked, since it does not work the way it says it does now.
An assessment can be applicable because of a lot of different reasons, not only because the language is supported.

We could use this function:
```js
const isLanguageAvailable = getLanguageAvailability( paper.getLocale(), availableLanguages );
```
Although we then need a way to get the locale of the paper within the `Aggregator`.

src/tree/assess/scoreAggregators/ReadabilityScoreAggregator.js
```diff
@@ -0,0 +1,169 @@
+import { scoreToRating } from "../../../interpreters";
+
+/* Internal dependencies */
+import { ScoreAggregator } from "../index";
+
+/**
+ * Total number of available readability assessments.
+ *
+ * To do: Compute this, rather than set this and keep this updated (see https://github.com/Yoast/YoastSEO.js/pull/2176#discussion_r261547836)
```
igorschoester: Please create an issue for this instead of a comment in the code!

As discussed, a second look at how we want to score readability would be nice. The logic in this is now a copy of the old one. It is pretty complex and I have the feeling it could be improved. However, it would be better to do that in an iteration.

This could be done in cooperation with someone that has the knowledge on why this is the way it is.

src/tree/assess/scoreAggregators/ReadabilityScoreAggregator.js
```diff
@@ -0,0 +1,169 @@
...
+		 * of applicable assessments either way. So this check suffices.
+		 */
+		return results.length === totalNrOfAssessments;
+	}
+
+	/**
+	 * Calculates the total score (GOOD, OKAY or NEEDS IMPROVEMENT)
+	 * based on the total penalty score.
+	 *
+	 * @param {boolean} allAssessmentsSupported Whether all assessments are supported for this language.
```
igorschoester: I think you forgot to rename this one to `isFullySupported` 😉

src/tree/assess/scoreAggregators/ReadabilityScoreAggregator.js
```diff
@@ -0,0 +1,169 @@
...
+	 * @param {number}  penalty                 The total penalty.
+	 *
+	 * @returns {number} The total score.
+	 */
+	calculateScore( allAssessmentsSupported, penalty ) {
+		if ( allAssessmentsSupported ) {
+			if ( penalty > 6 ) {
```
igorschoester, line 105: Magic numbers: `6`, `4` and `4`, `2`. Let me try to understand.

`isFullySupported`: When the amount of results is the same as the expected amount.
`penalty`: The sum of penalty points for each result.

When the penalty is higher than `6`, the readability needs improvement.
When the penalty is higher than `4`, the readability is ok.
When the penalty is `4` or lower, the readability is good.

To make up for the lesser amount of results, the penalty thresholds are lower for when it is not fully supported.
Meaning that a lower amount of gathered penalties will result in worse scores sooner.

This is on top of the already higher penalty point being given to the `bad` score of the partially supported. Which might just be done for finer control, maybe a better break between the `bad` and `ok` scores. I'm just guessing here. Maybe @IreneStr can enlighten us a bit here?

igorschoester: In general, I would hope some more comments here to explain the magic numbers and proces. But I understand you are also just guessing 😅

igorschoester (changes requested):
CR 🏗 

Please, no `todo` in the code!

As for the discussion(s), I think we could move this to a refactoring issue.
In cooperation with people (Irene!) that know how and why this works the way it does.

I might even accept the magic numbers for now 😉

src/tree/assess/scoreAggregators/ReadabilityScoreAggregator.js
hansjovis: It is actually quite simple once you look at the penalty scores given to each result type (`good`, `ok` or `bad`). E.g. two `bad` results give a total score of 6 (2 x 3) when fully supported, and a total score of 8 when partially supported (2 x 4).

I added this information as comments to the code, as suggested.

src/tree/assess/scoreAggregators/ReadabilityScoreAggregator.js
hansjovis: I agree. It can be improved and we definitely need to ask someone who knows the ins- and outs of the current scoring mechanism.

igorschoester (approved):

igorschoester:
CR & Acceptance 👍

PR #2184, opened 2019-03-04
Remove question about recalibration

manuelaugustin:
## Summary

*  n/a

* The recalibration beta has ended, so we don't need this item in the issue template anymore.


hansjovis (approved):
CR: 👍

PR #2189, opened 2019-03-05
Add specs for German morphology exception nouns

agnieszkaszuba:
## Summary

* [non-user-facing] Adds specs that check whether words from the German noun exception list that end in a consonant + s get stemmed correctly

## Test instructions

* Run the spec and make sure there are no errors


igorschoester:
CR & Acceptance 👍 

Forms found here: https://de.wiktionary.org/wiki/Krebs

PR #2191, opened 2019-03-06
Improve german participle functionality

### LINKED ISSUE #2190: Add functionality for German participles
German word stemming/word form creation should be able to deal with participles. For this, we need to do the following:
* We should add functionality to detect and stem regular participles. These stemmed participles should then get the correct forms (verb + adjective forms).
* We should add functionality to create regular prefixed participle forms from stemmed words.

Parent #2160

manuelaugustin:
## Summary

* Makes it possible to create stems from German participles, e.g. `gekauft` -> `kauf`. From these stems, the correct verb and adjectives forms are then created.
* Adds functionality to create participle forms with prefixes, e.g. `eingekauft`.

* The check for whether a word is a participle takes place after checking exceptions (since these also include participle stems) but before creating regular forms.
* If a participle is detected, we only create verb forms (assuming that the participle was used verbally, e.g. "er hat sich die Haare gefärbt" - "he dyed his hair") and adjective forms (assuming that the participle was used adjectivally, e.g. "die Haare sind gefärbt" - "the hair is dyed"). The adjective forms are based on the stem that has only the suffixes removed, not the prefixes. This is because we want forms such as "die gefärbten Haare" and not (incorrectly) "*die färbten Haare".

## Test instructions

* To check whether participles get stemmed correctly:
  * In the example, make sure your locale is set to `de_DE` and you have enabled morphology.
  * Add a text with at least 150 words.
  * Set a keyphrase containing a participle, e.g. `gekauft`.
  * In the text, add different forms of the verb `kaufen`, e.g. `kaufe`, `kaufte` etc.
  * Make sure these forms get recognized as forms of the keyphrase.
* To check whether participle forms get created correctly:
 * In the example, make sure your locale is set to `de_DE` and you have enabled morphology.
  * Add a text with at least 150 words.
  * Set a keyphrase containing a verb, e.g. `kaufen`.
  * In the text, add the participle `gekauft`.
  * Make sure the participle forms gets recognized as a form of the keyphrase.
* For different verb/verb forms for testing, consult a dictionary, e.g. https://de.wiktionary.org/wiki/Wiktionary:Hauptseite.

Fixes #2190

### DIFF
--- a/src/morphology/german/detectAndStemRegularParticiple.js
+++ b/src/morphology/german/detectAndStemRegularParticiple.js
@@ -15,7 +15,7 @@ const detectAndStemParticiplesWithoutPrefixes = function( morphologyDataVerbs, w
 
 	/*
 	 * Check if it's a ge + stem ending in d/t + et participle.
-	 * As this is the more specific regex, which needs to be checked before the ge + stem + t regex.
+	 * As this is the more specific regex, it needs to be checked before the ge + stem + t regex.
 	 */
 	if ( participleRegex2.test( word ) ) {
 		// Remove the two-letter prefix and the two-letter suffix.
@@ -39,7 +39,7 @@ const detectAndStemParticiplesWithoutPrefixes = function( morphologyDataVerbs, w
  * @param {string[]}    prefixes    The prefixes of a certain prefix class.
  * @param {string}      regexPart   The regex part for a given class (completed to a full regex within the function).
  * @param {number}      startStem   Where to start cutting off the de-prefixed word.
- * @param {number}       endStem     Where to end cutting off the de-prefixed word (from the end index).
+ * @param {number}      endStem     Where to end cutting off the de-prefixed word (from the end index).
  *
  * @returns {string} The stem.
  */
--- a/src/morphology/german/detectAndStemRegularParticiple.js
+++ b/src/morphology/german/detectAndStemRegularParticiple.js
@@ -1,0 +1,86 @@
+import exceptionsParticiplesActive from "../../researches/german/passiveVoice/exceptionsParticiplesActive";
+import { exceptions } from "../../researches/german/passiveVoice/regex";
+
+export function detectAndStemRegularParticiple( morphologyDataVerbs, word ) {
+	const participleRegex1 = new RegExp( "^" + morphologyDataVerbs.participleRegexes.participleRegexGeStemTEnd );
+	const participleRegex2 = new RegExp( "^" + morphologyDataVerbs.participleRegexes.participleRegexGeStemTdEt );
+
+	if ( exceptions( word ).length > 0 || exceptionsParticiplesActive().includes( word ) ) {
+		return "";
+	}
+
+	/*
+	 * Check if it's a ge + stem ending in d/t + et participle.
+	 * As this is the more specific regex, which needs to be checked before the ge + stem + t regex.
+	 */
+	if ( participleRegex2.test( word ) ) {
+		// Remove the two-letter prefix and the two-letter suffix.
+		return ( word.slice( 2, word.length - 2 ) );
+	}
+
+	// Check if it's a ge + stem + t participle.
+	if ( participleRegex1.test( word ) ) {
+		// Remove the two-letter prefix and the one-letter suffix.
+		return ( word.slice( 2, word.length - 1 ) );
+	}
+
+	/*
+	 * Check if it's a separable prefix + ge + stem ending in d/t + et participle.
+	 * As this is the more specific regex, which needs to be checked before the ge + stem + t regex.
+	 */
+	for ( let i = 0; i < morphologyDataVerbs.verbPrefixesSeparable.length; i++ ) {
+		const currentPrefix = morphologyDataVerbs.verbPrefixesSeparable[ i ];
+		const participleRegex = new RegExp( "^" + currentPrefix + morphologyDataVerbs.participleRegexes.participleRegexGeStemTdEt );
+
+		if ( participleRegex.test( word ) ) {
+			const wordWithoutPrefix = word.slice( currentPrefix.length - word.length );
+			const wordWithoutParticipleAffixes = wordWithoutPrefix.slice( 2, wordWithoutPrefix.length - 2 );
+
+			return ( currentPrefix + wordWithoutParticipleAffixes );
+		}
+	}
+
+	// Check if it's a separable prefix + ge + stem + t participle.
+	for ( let i = 0; i < morphologyDataVerbs.verbPrefixesSeparable.length; i++ ) {
+		const currentPrefix = morphologyDataVerbs.verbPrefixesSeparable[ i ];
+		const participleRegex = new RegExp( "^" + currentPrefix + morphologyDataVerbs.participleRegexes.participleRegexGeStemTEnd );
+
+		if ( participleRegex.test( word ) ) {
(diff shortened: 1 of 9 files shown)


src/morphology/german/detectAndStemRegularParticiple.js
```diff
@@ -0,0 +1,196 @@
...
+ * @param {string}      word        The word (not stemmed) to check.
+ * @param {string[]}    prefixes    The prefixes of a certain prefix class.
+ * @param {string}      regexPart   The regex part for a given class (completed to a full regex within the function).
+ * @param {number}      startStem   Where to start cutting off the de-prefixed word.
+ * @param {number}      endStem     Where to end cutting off the de-prefixed word (from the end index).
+ *
+ * @returns {string} The stem.
+ */
+const detectAndStemParticiplePerPrefixClass = function( word, prefixes, regexPart, startStem, endStem ) {
+	for ( let i = 0; i < prefixes.length; i++ ) {
```
hansjovis, line 47: This can be replaced by a `for ... of` loop:
```js
for ( const currentPrefix of prefixes ) {
```

src/morphology/german/detectAndStemRegularParticiple.js
```diff
@@ -0,0 +1,196 @@
...
+/**
+ * Detects whether a word is a regular participle with a prefix and if so, returns the stem.
+ *
+ * @param {Object}  morphologyDataVerbs The German morphology data for verbs.
+ * @param {string}  word                The word (not stemmed) to check.
+ *
+ * @returns {string} The stem.
+ */
+const detectAndStemParticiplesWithPrefixes = function( morphologyDataVerbs, word ) {
+	const prefixExceptionChecks = [
```
hansjovis, line 71: Now I am looking into this list of factory functions, I do think we can make this better! 😄

1. `startStem` and `endStem` is configuration that can be in the `morphology` file.
2. If I read this code correctly, we try to detect and stem each exception class twice. Once for `seperable` or `inseperable` (depending on the class) and once for `seperableOrInseperable`.
I think this can live as configuration in the `morphology` file as well.

So, the morphology file would start to look like this:
  ```json
  "participleGeStemTEnd": {
     "regex": "ge(.*)t$",
     "startStem": 2,
     "endStem": 1,
      "separable": true
  }
  ```
Now the function can look like this:
```js
const prefixesSeparableOrInseparable = morphologyDataVerbs.verbPrefixesSeparableOrInseparable
for ( const key in morphologyDataVerbs.participleRegexes ) {
     const participleClass = morphologyDataVerbs.participleRegexes[ key ];

     const startStem = participleClass.startStem;
     const endStem = participleClass.endStem;
     const separable = participleClass.separable;

     const prefixes = separable ? morphologyDataVerbs.verbPrefixesSeparable : morphologyDataVerbs.verbPrefixesInseparable;

     let stem = detectAndStemParticiplePerPrefixClass( word, prefixes, startStem, endStem );
     if ( stem ) {
         return stem;
     }
     stem = detectAndStemParticiplePerPrefixClass( word, prefixesSeparableOrInseparable, startStem, endStem );
     if ( stem ) {
          return stem;
      }
}
```

src/morphology/german/generateParticipleForm.js
```diff
@@ -18,10 +19,74 @@ const addParticipleAffixes = function( stemmedWord, affixes ) {
...
+ * Generates participle forms with separable or separable/inseparable prefixes.
+ *
+ * @param {Object}      morphologyDataVerbs The German morphology data for verbs.
+ * @param {string}      stemmedWord         The stem to check.
+ * @param {string[]}    prefixes            The prefixes to check.
+ *
+ * @returns {string} The created participle form.
+ */
+const generateParticipleFormWithSeparablePrefix = function( morphologyDataVerbs, stemmedWord, prefixes ) {
+	for ( let i = 0; i < prefixes.length; i++ ) {
```
hansjovis, line 40: can be a `for ... of` loop as well:
```js
for ( const currentPrefix of prefixes ) {
```

src/morphology/german/detectAndStemRegularParticiple.js
```diff
@@ -0,0 +1,196 @@
...
+	];
+
+	for ( let i = 0; i < prefixExceptionChecks.length; i++ ) {
+		const stem = prefixExceptionChecks[ i ]();
+		if ( stem.length > 0 ) {
+			return stem;
+		}
+	}
+
+	return "";
```
hansjovis, line 167: I would return `null` instead of the empty string. 
This communicates more clearly that no stem has been found. It could be, although unlikely (?), that the stemming algorithm stems a word to the empty string.

hansjovis: The same for the other function where the empty string is returned as a "not found" value.

Before I forget: the JSDoc `@returns` description should be updated as well to reflect that the function returns `null` if no stem has been found 😄.

src/morphology/german/detectAndStemRegularParticiple.js
```diff
@@ -0,0 +1,196 @@
...
+ * @returns {string} The stem.
+ */
+export function detectAndStemRegularParticiple( morphologyDataVerbs, word ) {
+	if ( exceptions( word ).length > 0 || exceptionsParticiplesActive().includes( word ) ) {
+		return "";
+	}
+
+	let stem = detectAndStemParticiplesWithoutPrefixes( morphologyDataVerbs, word );
+
+	if ( stem.length > 0 ) {
```
hansjovis, line 185: the empty string evaluates to `false` in JavaScript (don't ask):
```js
console.log( "" == false ); // logs 'true'
```
So this can be replaced by:
```js
if ( stem ) {
```
Bonus: this would also trigger on a `null` value 🙂.

src/morphology/german/detectAndStemRegularParticiple.js
```diff
@@ -0,0 +1,196 @@
...
+/**
+ * Detects whether a word is a regular participle without a prefix and if so, returns the stem.
+ *
+ * @param {Object}  morphologyDataVerbs The German morphology data for verbs.
+ * @param {string}  word                The word (not stemmed) to check.
+ *
+ * @returns {string}    The stem.
+ */
+const detectAndStemParticiplesWithoutPrefixes = function( morphologyDataVerbs, word ) {
+	const participleRegex1 = new RegExp( "^" + morphologyDataVerbs.participleRegexes.participleRegexGeStemTEnd );
```
hansjovis, line 13: These regexes could live outside of this function, on the file level.
Could help with regex compilation, since it would only happen once 🙂.

hansjovis (changes requested):
Looking good!
Do have some comments 😅.

manuelaugustin:
Closed in favor of new PR on the `javascript` mono-repo: https://github.com/Yoast/javascript/pull/12

src/morphology/german/detectAndStemRegularParticiple.js
manuelaugustin: As discussed, I'm not not implementing this change since the regexes make use of the morphology data, which is only available within the scope of the function.

src/morphology/german/detectAndStemRegularParticiple.js
manuelaugustin: I'm super happy that you found a way out of the factory function madness that I had started 😅

Implemented as suggested, just replaced `for in` with `for of` loop (and the object in the morphology data file with an array) to preserve the order in which the participle classes are checked.

